PR #5, opened 2024-02-26
✨(backend) support Authorization Code flow

lebaudantoine:
## Purpose

Preparing project for staging deployment by adding backend support for Authorization Code Flow in OIDC.

## Proposal

- [X] Configure Keycloak server
- [X] Drop JWT authentication
- [X] Add new dependency `mozzila-django-oidc`
- [X] Fix production image


src/backend/core/authentication.py
```diff
@@ -1,59 +1,101 @@
-"""Authentication for the publish core app."""
-from django.conf import settings
-from django.utils.functional import SimpleLazyObject
-from django.utils.module_loading import import_string
+"""Authentication for the People core app."""
```
sampaccoud, line 1:
```suggestion
"""Authentication for the core app."""
```

src/backend/core/authentication.py
```diff
@@ -1,59 +1,101 @@
...
+                    sub=sub,
+                    email=user_info.get("email"),
+                    password="!",  # noqa: S106
+                )
+            else:
+                user = None
+
+        return user
+
+    def create_user(self, claims):
```
sampaccoud, line 89: but you are not calling "create_user" in the above method so it is useless?

sampaccoud (approved):

PR #56, opened 2024-05-13
✨(models/api) allow inviting external users to a document by their email

sampaccoud:
## Purpose

We want to be able to share a document with a person even if this person does not have an account in impress yet.

## Proposal

This code is ported from https://github.com/numerique-gouv/people and adapted to the differences in `Impress`:
- accesses work for users as well as for teams
- no `Identity` object, the identity is carried by the `User`


AntoLC:
I can see the tests failed probably because `mail/html/invitation.html` is not generated. 

I have a question about that, why the mail templates are not git versioned ? If we versioned our templates, we will not have to build them constantly in the CI or in the `make bootstrap`.

----

Otherwise they can be generated like that in the CI:
See: https://github.com/numerique-gouv/people/commit/f6d5f737f444ab2556c040c10fbd3e431083265f

AntoLC (commented):
I can see the tests failed probably because `mail/html/invitation.html` is not generated. 

I have a question about that, why the mail templates are not git versioned ? If we versioned our templates, we will not have to build them constantly in the CI or in the `make bootstrap`.

----

Otherwise they can be generated like that in the CI:
See: https://github.com/numerique-gouv/people/commit/f6d5f737f444ab2556c040c10fbd3e431083265f

src/mail/mjml/invitation.mjml
```diff
@@ -0,0 +1,54 @@
+<mjml>
+  <mj-include path="./partial/header.mjml" />
+
+  <mj-body mj-class="bg--blue-100">
+    <mj-wrapper css-class="wrapper" padding="0 40px 40px 40px">
+      <mj-section background-url="{% base64_static 'images/mail-header-background.png' %}" background-size="cover" background-repeat="no-repeat" background-position="0 -30px">
```
AntoLC, line 6: The images are missing: 
image

You can get them here:
https://github.com/numerique-gouv/people/tree/main/src/backend/core/static/images

src/mail/mjml/invitation.mjml
```diff
@@ -0,0 +1,54 @@
...
+          <mj-text>{% trans "We are delighted to welcome you to our community on Impress, your new companion to collaborate on documents efficiently, intuitively, and securely." %}</mj-text>
+          <mj-text>{% trans "Our application is designed to help you organize, collaborate, and manage permissions." %}</mj-text>
+          <mj-text>
+            {% trans "With Impress, you will be able to:" %}
+            <ul>
+              <li>{% trans "Create documents."%}</li>
+              <li>{% trans "Invite members of your document or community in just a few clicks."%}</li>
+            </ul>
+          </mj-text>
+          <mj-button href="https://impress-staging.beta.numerique.gouv.fr/" background-color="#000091" color="white" padding-bottom="30px">
```
AntoLC, line 35: `https://impress-staging.beta.numerique.gouv.fr/` should be a variable I suppose, it will work only with staging  as it is now.
Maybe the referer domain (frontend) that call the CREATE invitation should be used as the url inside the mail ?

sampaccoud:
@AntoLC I think email templates should not be persisted for reason of "unicity of information". As long as we are able to generate the content, it is technically not part of our source code in my opinion.

src/mail/mjml/invitation.mjml
sampaccoud: I used the Site framework in Django as is usually done.

sampaccoud:
@AntoLC I pushed a fixup. The PR is ready for further review!

src/mail/mjml/invitation.mjml
```diff
@@ -0,0 +1,54 @@
...
+            <p>{% trans "Invitation to join a team" %}</p>
+          </mj-text>
+
+          <!-- Welcome Message -->
+          <mj-text>
+            <h1>{% blocktrans %}Welcome to <strong>Impress</strong>{% endblocktrans %}</h1>
+          </mj-text>
+          <mj-divider border-width="1px" border-style="solid" border-color="#DDDDDD" width="30%" align="left"/>
+
+          <mj-image src="{% base64_static 'images/logo.svg' %}" width="157px" align="left" alt="{%trans 'Logo' %}" />
```
AntoLC, line 23: The svg format does not seem to work.
We can see that in another PR though.

image

AntoLC (approved):
Nice!

src/mail/mjml/invitation.mjml
sampaccoud: fixed.

PR #73, opened 2024-05-24
💚(CI) Add git hook

rouja:
Add git hook to avoid secret leaks


AntoLC:
I think the lint git does not like capital on title, try like that: `💚(CI) add git hook`

PR #77, opened 2024-05-28
♻️(models) rename document/template access rights

sampaccoud:
## Purpose

The "member" access right does not make sense for documents and templates. 

## Proposal

What we really need are "editor" and "reader" access rights.


src/backend/core/api/viewsets.py
```diff
@@ -368,15 +367,15 @@ class DocumentAccessViewSet(
 
     POST /api/v1.0/documents/<resource_id>/accesses/ with expected data:
         - user: str
-        - role: str [owner|admin|member]
+        - role: str [admin|editor|reader]
```
AntoLC, line 370: Seems to be `administrator`.
```suggestion
        - role: str [administrator|editor|reader]
```

src/backend/core/migrations/0001_initial.py
AntoLC, line 1: I think this file should not be modified anymore, because the database is already created in the production environment (staging as well) so we should only add new migration, or I miss something ?

src/backend/core/api/viewsets.py
sampaccoud: done!

src/backend/core/migrations/0001_initial.py
sampaccoud: I would like to break it one more time. We can drop the db in our environments and then we ship v1.0.0?

PR #79, opened 2024-05-29
✨(backend) search users

AntoLC:
## Purpose

We need to be able to research users to invite them.
This PR add the route GET `/api/v1.0/users/`.
2 query params are available:
- `q` to search on `email` field
- `document_id` to exclude user that have access to a `document_id`

## Proposal

- [x] add the GET `/api/v1.0/users/`
- [x] change user serializer to get the `email` field
- [x] fix CI backend linter


src/backend/core/migrations/0002_create_pg_trgm_extension.py
AntoLC, line 1: I Should put it in the `0001_initial`  maybe?

PR #90, opened 2024-06-10
🐛(CI) improve submodule

rouja:
- remove deplucate declaration
- simplify helmfile
- use symlink

### DIFF
--- a/src/helm/env.d/production/secrets.enc.yaml
+++ b/src/helm/env.d/production/secrets.enc.yaml
@@ -1,0 +1,1 @@
+../../../../secrets/numerique-gouv/impress/env/production/secrets.enc.yaml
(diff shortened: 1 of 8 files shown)


src/helm/env.d/production/secrets.enc.yaml
```diff
@@ -0,0 +1 @@
+../../../../secrets/numerique-gouv/impress/env/produciton/secrets.enc.yaml
```
AntoLC, line 1:
```suggestion
../../../../secrets/numerique-gouv/impress/env/production/secrets.enc.yaml
```

rouja: The error is on the secrets repository. I will fix the typo on both !

AntoLC (approved):

PR #117, opened 2024-07-01
♻️(documents) store document file in object storage in a folder

sampaccoud:
## Purpose

We will need to store more than a file for a document in object storage: multiple languages, images, etc.
For this reason, the document ID should be a folder and the content a file should be stored in this folder.

## Proposal

Store document content in object storage under a folder named by the document ID. I propose to call the content file:
 `{document.id}/file`.

Thus we will be able to store the documents' images under an `images` subfolder for example:
 `{document.id}/images/{image.id}`

### DIFF
--- a/src/backend/core/models.py
+++ b/src/backend/core/models.py
@@ -317,11 +317,18 @@ class Document(BaseModel):
         return self.title
 
     @property
+    def key_base(self):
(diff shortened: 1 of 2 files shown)


AntoLC (approved):
Seems good.

PR #178, opened 2024-08-19
Add image upload and download from media url

sampaccoud:
## Purpose

We want to be able to attach images to a document. Strict access control should be applied on all images of a document.

## Proposal

- upload images to a folder next to each document's content file : `{document.id}/attachments/{image_filename}
- allow downloading attachment images from object storage via the nginx ingress. Access control is done thanks to nginx subrequests.
- Update the helm chart to add the new media url with access control delegated to Django


AntoLC (approved):
I think we can merge it.
Let's open issues about: 
- images not working from a pdf and docx.
- add ngnix conf in docker-compose dev mode

WDYT ?

PR #232, opened 2024-09-05
🐛 (backend) gitlab oicd userinfo endpoint

AntoLC:
## Purpose

Gitlab oidc was failing.

It is failing because the userinfo endpoint can return 2 content types:
- application/json
- application/jwt

https://github.com/numerique-gouv/impress/blob/53267c783a0b47b15025dc0d211d2a77a345de8a/src/backend/core/authentication/backends.py#L45-L51

Gitlab oidc returns a json object, while Agent Connect oidc returns a jwt token.
We are adapting the authentication to handle both cases.

----

⚠️ The logout does not seem to work with the Gitlab oidc. I suggest we open an issue, maybe you will want to dig more on that @securitykernel ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+## Fixed
+
+- 🐛 (backend) gitlab oicd userinfo endpoint #232
+
 
 ## [1.4.0] - 2024-09-17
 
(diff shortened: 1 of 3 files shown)


sampaccoud (approved):
I proposed a small simplification using the "responses" library that we have in dev dependencies. Otherwise GTM!

PR #234, opened 2024-09-06
🛂(backend) stop to list public doc to everyone

AntoLC:
## Purpose

Everybody could see the full list of public docs.
We don't want public documents to appear in the list view unless it is explicitly shared with the logged-in user.

We want link access to be more powerfull:
- allow link access to grant editor rights
- possibility to limit link access to all authenticated users or to anonymous users

## Proposal

This is a major refactoring of the link access to a document.

Link access can now be configured with 2 parameters on the document object:
- link reach can be private (users need to obtain specific access by document's administrators), restricted (any authenticated user) or public (anybody including anonymous users)
- link role can be reader or editor.
    
    It is thus now possible to give editor access to an anonymous user or
    any authenticated user.

We add a special route for link access configuration : `/api/v1.0/documents/{document_id}/link-configuration/`
to secure and simplify limiting update on these fields to only administrators/owners of the document.

While working on this PR we bumped on some issues that we fixed:
- Remove possibility to force document id on creation
- Fix randomly failing test on document list ordering via API
- Remove Docker Compose version to remove warning (it is now automatically guessed)


src/backend/core/api/viewsets.py
```diff
@@ -331,30 +321,68 @@ class DocumentViewSet(
...
+        serializer = self.get_serializer(instance)
+
+        if self.request.user.is_authenticated:
+            try:
+                # Add a trace that the user visited the document (this is needed to include
+                # the document in the user's list view)
+                models.LinkTrace.objects.create(
+                    document=instance,
+                    user=self.request.user,
+                )
```
AntoLC, line 364: It creates an error if you access the document twice and more, because of the unicity document/user.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to
 
 - 💄(frontend) code background darkened on editor #214
 - 🔥(frontend) hide markdown button if not text #213
+- 🛂(backend) stop to list public doc to everyone #234
```
AntoLC, line 23: This line will have to move after the `main` rebase.

src/backend/core/api/viewsets.py
sampaccoud: hmmm. The try/except is here for that? I didn't manage to reproduce the error. Can you introspect what are the values of instance and self.request.user when this occurs and what is the error type generated?

AntoLC: I get a 400, "Document/user link trace with this User and Document already exists.".

image

If a change the method to `update_or_create`, it is working. 
```py
models.LinkTrace.objects.update_or_create
```

AntoLC: Ok, it is because of the type of error, it is a `ValidationError` and not a `IntegrityError`
https://github.com/numerique-gouv/impress/blob/cec2531352184163a0fc1988deb829c0222c2aaa/src/backend/core/api/viewsets.py#L365

PR #240, opened 2024-09-09
✨Activate versions feature

AntoLC:
## Purpose

We fixed the 404 error when trying to access the version feature, so we reactivated the version feature in the doc header.


AntoLC:
Duplicate of: https://github.com/numerique-gouv/impress/pull/277

PR #243, opened 2024-09-10
♻️(frontend) adapt doc visibility to new api

AntoLC:
## Purpose

We just adapt to the new way of setting our doc public or not. 
We didn't implement the public editable yet.

We updated the way we handle the visibility of a doc in the backend. 
Now we use a new api to update the visibility (`documents/{id}/link-configuration/`) of a doc. 
We adapted the frontend to use this new api.
We changed the types to reflect the new api and to keep the same logic.


sampaccoud:
Commit was cherry-picked to https://github.com/numerique-gouv/impress/pull/234 and merged.

PR #255, opened 2024-09-13
🌐 (frontend) Add German translation 

securitykernel:
Adds a German translation to the frontend.


AntoLC (approved):

AntoLC (commented):
There is still some translations to do on Crowdin, you can just do the frontend part (frontend-impress.json), the backend part seems to be made here #315 .
When the translation are made, on the project root, if you do:
`make crowdin-download && make frontend-i18n-compile`, it will pull the translations from crowdin and add them automatically in the json file.

PR #258, opened 2024-09-16
♻️(api) refactor getting versions to expose pagination

sampaccoud:
## Purpose

Getting versions was not working properly. Some versions returned were not accessible by the user requesting the list of available versions.

## Proposal

We refactor the code to make it simpler and let the frontend handle pagination (load more style).

### DIFF
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -343,10 +343,10 @@ class InvitationSerializer(serializers.ModelSerializer):
         return attrs
 
 
-class DocumentVersionSerializer(serializers.Serializer):
-    """Serialize Versions."""
+class VersionFilterSerializer(serializers.Serializer):
+    """Validate version filters applied to the list endpoint."""
 
-    etag = serializers.CharField()
-    is_latest = serializers.BooleanField()
-    last_modified = serializers.DateTimeField()
-    version_id = serializers.CharField()
+    version_id = serializers.CharField(required=False, allow_blank=True)
+    page_size = serializers.IntegerField(
+        required=False, min_value=1, max_value=50, default=20
+    )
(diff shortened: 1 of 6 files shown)


src/backend/core/api/serializers.py
```diff
@@ -343,10 +343,10 @@ def validate(self, attrs):
...
+    """Validate version filters applied to the list endpoint."""
 
-    etag = serializers.CharField()
-    is_latest = serializers.BooleanField()
-    last_modified = serializers.DateTimeField()
-    version_id = serializers.CharField()
+    version_id = serializers.CharField(required=False, allow_blank=True)
+    page_size = serializers.IntegerField(
+        required=False, min_value=1, max_value=50, default=20
+    )
```
AntoLC, line 352: `DOCUMENT_VERSIONS_PAGE_SIZE` is at 50, better to put it at 20 maybe, wdyt?

```suggestion
    page_size = serializers.IntegerField(
        required=False, min_value=1, max_value=50, default=settings.DOCUMENT_VERSIONS_PAGE_SIZE
    )
```

AntoLC (commented):
Works great!

I am just wondering if it is necessary to provide the latest version (the current one so).
Is it easy to add a filter on it on the backend side?
Otherwise, on the frontend side I have to apply a filter to display only the version where `is_latest` is at false
image

src/backend/core/api/serializers.py
sampaccoud: I have more than 20 pages on my welcome screen and it works great for the moment. Let's decide later and in another PR.

sampaccoud:
> I am just wondering if it is necessary to provide the latest version (the current one so). Is it easy to add a filter on it on the backend side? Otherwise, on the frontend side I have to apply a filter to display only the version where `is_latest` is at false !

It was not so trivial but I pushed a fixup for your review! :sweat_smile:

AntoLC (approved):
Bravo !

PR #268, opened 2024-09-17
:globe_with_meridians:(frontend) add localization to editor

arnaud-robin:
## Purpose

Right now most of the elements in the frontend are localized but the editor. 
The localization should also be incorporated in the editor itself.

## Proposal

I suggest we use the default translations already offered by blocknotejs and reload the editor whenever a new language is selected.


PanchoutNathan:
Close as duplicate from #368

PR #273, opened 2024-09-20
✨(frontend) add crisp chatbot

arnaud-robin:
## Purpose

We want to add crisp chatbot to allow for easy user interaction with support team.

## Proposal

We add the chatbot component using crisp-web-sdk. We initialize the chatbot with the user email to allow for follow-up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to
 - ✨(backend) config endpoint #425
 - ✨(frontend) config endpoint #424
 - ✨(frontend) add sentry #424
+- ✨(frontend) add crisp chatbot #273
 
 ## Changed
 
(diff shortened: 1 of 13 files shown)


AntoLC (approved):
I added a few tests, I let you rebase.

We added the key in the secret as well, to be set at build time.

src/frontend/apps/impress/src/hook/useSupport.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+const isCrispConfigured = (): boolean => {
+  return typeof window !== 'undefined' && !!window.$crisp;
+};
+
+export const initializeSupportSession = (user: User) => {
+  if (!isCrispConfigured()) {
+    return;
+  }
+  Crisp.setTokenId(user.id);
```
lebaudantoine, line 14: I am not sure anymore of this part … let's discuss it before merging, I am reviewing their code

PR #279, opened 2024-09-23
✅(backend) fix flaky tests

AntoLC:
## Purpose

- ✅(backend) fix flaky test on tmp file: 
It seems to have a race condition, sometimes the tmp file is not deleted before the test assertion.  We let the test sleep for 0.5 second before the assertion.

- ✅(CI) fix flaky test on Minio initialized:
Minio server need to be initialized before running the job to configure Minio.
We add a delay to wait for Minio server to be ready.

### DIFF
--- a/.github/workflows/impress.yml
+++ b/.github/workflows/impress.yml
@@ -168,7 +168,7 @@ jobs:
           path: "src/backend/core/templates/mail"
           key: mail-templates-${{ hashFiles('src/mail/mjml') }}
 
-      - name: Start Minio
+      - name: Start MinIO
         run: |
           docker pull minio/minio
           docker run -d --name minio \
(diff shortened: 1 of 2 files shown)


.github/workflows/impress.yml
```diff
@@ -178,6 +178,9 @@ jobs:
             -v /data/media:/data \
             minio/minio server --console-address :9001 /data
 
+      - name: Add delay to wait for Minio to start correctly
+        run: sleep 5
```
sampaccoud, line 182: maybe use something like this to avoid wasting time?
https://github.com/openfun/joanie/blob/main/.circleci/config.yml#L227

sampaccoud (approved):

PR #301, opened 2024-10-02
✨(backend) add full_name and short_name to user model and API

sampaccoud:
## Purpose

We need to add a name for our users instead of showing their email addresses.

## Proposal

Add a `short_name` and `full_name` fields as per the best international practice.

The full_name and short_name field are synchronized with the OIDC token upon each login. The fields from the token that are synchronized with the the `short_name` and `full_name` fields are configurable (in our case `firstname` + `lastname` goes to `full_name`, while `first_name` goes to `short_name`)


AntoLC (approved):
Top!

Can we add the names here as well, with names can be sometime empty?
https://github.com/numerique-gouv/impress/blob/e6425066757c3afac2325096d70aad41846b9b1f/src/backend/demo/management/commands/create_demo.py#L115-L125

PR #303, opened 2024-10-03
🛂(backend) match email if no existing user matches the sub

sampaccoud:
## Purpose

Some OIDC identity providers may provide a random value in the "sub" field instead of an identifying ID. In this case, it may be a good idea to fallback to matching the user on its email field.

## Proposal

- Add a setting `OIDC_FALLBACK_TO_EMAIL_FOR_IDENTIFICATION` which defaults to `True`
- When logging-in, if the sub does not match any existing user, try matching on the `email` field before creating a new user for the unknown sub.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to
 
 ## Fixed
 
+- 🛂(frontend) match email if no existing user matches the sub
 - 🐛(backend) gitlab oicd userinfo endpoint #232
 - 🛂(frontend) redirect to the OIDC when private doc and unauthentified #292
 - ♻️(backend) getting list of document versions available for a user #258
--- a/src/backend/core/authentication/backends.py
+++ b/src/backend/core/authentication/backends.py
@@ -60,57 +60,61 @@ class OIDCAuthenticationBackend(MozillaOIDCAuthenticationBackend):
         return userinfo
 
     def get_or_create_user(self, access_token, id_token, payload):
-        """Return a User based on userinfo. Get or create a new user if no user matches the Sub.
-
-        Parameters:
-        - access_token (str): The access token.
-        - id_token (str): The ID token.
-        - payload (dict): The user payload.
-
-        Returns:
-        - User: An existing or newly created User instance.
-
-        Raises:
-        - Exception: Raised when user creation is not allowed and no existing user is found.
-        """
+        """Return a User based on userinfo. Create a new user if no match is found."""
 
         user_info = self.get_userinfo(access_token, id_token, payload)
+        email = user_info.get("email")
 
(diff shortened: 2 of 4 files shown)


qbey:
If the sub is random, how random is it? Is there not a risk of getting another user trying to use it?
Is there a way to know from the user info, the sub is random (so we can just ignore it and use the email instead)?

src/backend/impress/settings.py
```diff
@@ -384,6 +384,12 @@ class Base(Configuration):
     OIDC_STORE_ID_TOKEN = values.BooleanValue(
         default=True, environ_name="OIDC_STORE_ID_TOKEN", environ_prefix=None
     )
+    OIDC_FALLBACK_TO_EMAIL_FOR_IDENTIFICATION = values.BooleanValue(
+        default=True,
+        environ_name="OIDC_FALLBACK_TO_EMAIL_FOR_IDENTIFICATION",
+        environ_suffix=None,
```
qbey, line 390: Did you mean `environ_prefix` (like other settings)?

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to
 
 ## Fixed
 
+- 🛂(frontend) match email if no existing user matches the sub
```
mjeammet, line 27: Can you move this line to "unreleased", please ?

src/backend/core/authentication/backends.py
```diff
@@ -60,57 +60,61 @@ def get_userinfo(self, access_token, id_token, payload):
...
-            )
-            if has_changed:
-                updated_claims = {key: value for key, value in claims.items() if value}
-                self.UserModel.objects.filter(sub=sub).update(**updated_claims)
+        user = self.get_existing_user(sub, email)
+
+        if user:
+            self.update_user_if_needed(user, claims)
+        elif self.get_settings("OIDC_CREATE_USER", True):
+            user = User.objects.create(sub=sub, password="!", **claims)  # noqa: S106
```
qbey, line 89: Do you really want to create a user with a random sub?

src/backend/impress/settings.py
sampaccoud: Oups. Thanks!

src/backend/core/authentication/backends.py
sampaccoud: As discussed IRL. It's not random, it's a uuid so unique but the identity provider generates a new one each time the user logs in... With this PR, we will create an account with a sub that we will never see again but at least we will then continue using this account thanks to the email address that we can recognize...

src/backend/impress/settings.py
sampaccoud: @qbey fixed it!

qbey (approved):

PR #309, opened 2024-10-07
♻️(backend) allow uploading more types of attachments

sampaccoud:
## Purpose

We want to allow users to upload files to a document, not just images. 

## Proposal

We try to enforce coherence between the file extension and the real mime type of its content using magic.

If a file is deemed unsafe, it is still accepted during upload and the information is stored as metadata on the object for display to readers in the frontend.

### DIFF
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -6,6 +6,7 @@ from django.conf import settings
 from django.db.models import Q
 from django.utils.translation import gettext_lazy as _
 
+import magic
 from rest_framework import exceptions, serializers
 
 from core import models
@@ -219,15 +220,41 @@ class FileUploadSerializer(serializers.Serializer):
                 f"File size exceeds the maximum limit of {max_size:d} MB."
             )
 
-        # Validate file type
-        mime_type, _ = mimetypes.guess_type(file.name)
-        if mime_type not in settings.DOCUMENT_IMAGE_ALLOWED_MIME_TYPES:
-            mime_types = ", ".join(settings.DOCUMENT_IMAGE_ALLOWED_MIME_TYPES)
-            raise serializers.ValidationError(
-                f"File type '{mime_type:s}' is not allowed. Allowed types are: {mime_types:s}"
-            )
+        extension = file.name.rpartition(".")[-1] if "." in file.name else None
+
+        # Read the first few bytes to determine the MIME type accurately
+        mime = magic.Magic(mime=True)
+        magic_mime_type = mime.from_buffer(file.read(1024))
+        file.seek(0)  # Reset file pointer to the beginning after reading
+
+        self.context["is_unsafe"] = (
+            magic_mime_type in settings.DOCUMENT_UNSAFE_MIME_TYPES
+        )
+
+        extension_mime_type, _ = mimetypes.guess_type(file.name)
+
+        # Try guessing a coherent extension from the mimetype
+        if extension_mime_type != magic_mime_type:
+            self.context["is_unsafe"] = True
+
+        guessed_ext = mimetypes.guess_extension(magic_mime_type)
+        # Missing extensions or extensions longer than 5 characters (it's as long as an extension
+        # can be) are replaced by the extension we eventually guessed from mimetype.
+        if (extension is None or len(extension) > 5) and guessed_ext:
+            extension = guessed_ext[1:]
+
+        if extension is None:
+            raise serializers.ValidationError("Could not determine file extension.")
+
+        self.context["expected_extension"] = extension
 
         return file
+
+    def validate(self, attrs):
+        """Override validate to add the computed extension to validated_data."""
+        attrs["expected_extension"] = self.context["expected_extension"]
+        attrs["is_unsafe"] = self.context["is_unsafe"]
+        return attrs
 
 
 class TemplateSerializer(BaseResourceSerializer):
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -1,6 +1,5 @@
 """API endpoints"""
 
-import os
 import re
 import uuid
 from urllib.parse import urlparse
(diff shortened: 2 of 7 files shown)


src/backend/core/api/serializers.py
```diff
@@ -218,16 +219,36 @@ def validate_file(self, file):
...
+        if extension is None:
+            raise serializers.ValidationError("Could not determine file extension.")
+
+        self.context["expected_extension"] = extension
 
         return file
 
+    def validate(self, attrs):
+        """Override validate to add the computed extension to validated_data."""
+        attrs["expected_extension"] = self.context.get("expected_extension")
```
qbey, line 248: Nit: If `expected_extension` is not in the context, it means the validation failed right? So we can make explicit that the key is in the context don't you think?.

```suggestion
        attrs["expected_extension"] = self.context["expected_extension"]
```

src/backend/core/api/serializers.py
```diff
@@ -218,16 +219,36 @@ def validate_file(self, file):
...
-            )
+        extension = file.name.rpartition(".")[-1] if "." in file.name else None
+
+        # Read the first few bytes to determine the MIME type accurately
+        mime = magic.Magic(mime=True)
+        magic_mime_type = mime.from_buffer(file.read(1024))
+        file.seek(0)  # Reset file pointer to the beginning after reading
+
+        if magic_mime_type in settings.DOCUMENT_UNSAFE_MIME_TYPES:
+            self.context["is_unsafe"] = True
```
qbey, line 230: Why don't you want to fill this information always? Like 
```suggestion
        self.context["is_unsafe"] = magic_mime_type in settings.DOCUMENT_UNSAFE_MIME_TYPES
```

As a slight optimization, the `settings.DOCUMENT_UNSAFE_MIME_TYPES` could be a `set` instead of a list.

src/backend/core/api/viewsets.py
```diff
@@ -476,15 +475,22 @@ def attachment_upload(self, request, *args, **kwargs):
...
 
         # Generate a generic yet unique filename to store the image in object storage
         file_id = uuid.uuid4()
-        key = f"{document.key_base}/{ATTACHMENTS_FOLDER:s}/{file_id!s}{extension:s}"
+        extension = serializer.validated_data["expected_extension"]
+        key = f"{document.key_base}/{ATTACHMENTS_FOLDER:s}/{file_id!s}.{extension:s}"
+
+        # Prepare metadata for storage
+        metadata = {"Metadata": {"owner": str(request.user.id)}}
+        if serializer.validated_data.get("is_unsafe", False):
```
qbey, line 486: I think this is misleading, because `serializer.validated_data` always has an `is_unsafe` (that may sometimes be `None`).

I would rather use:
```suggestion
        if serializer.validated_data["is_unsafe]:
```

src/backend/core/api/serializers.py
```diff
@@ -218,16 +219,36 @@ def validate_file(self, file):
...
+            raise serializers.ValidationError("Could not determine file extension.")
+
+        self.context["expected_extension"] = extension
 
         return file
 
+    def validate(self, attrs):
+        """Override validate to add the computed extension to validated_data."""
+        attrs["expected_extension"] = self.context.get("expected_extension")
+        attrs["is_unsafe"] = self.context.get("is_unsafe")
```
qbey, line 249: As said above, I would fill the `is_unsafe` context always, so I would remove the `.get(` here. If you want to keep with the actual implementation, I would at least default to `False` instead of `None` here.

qbey (commented):
Nice :) I've made few nit picking comments

src/backend/core/api/serializers.py
sampaccoud: True that :+1:

src/backend/core/api/serializers.py
sampaccoud: I suppose because semantically it's overstated to say they are "safe" just because we are not sure that they are "unsafe" ^^ But yes good idea.

qbey (approved):

sampaccoud:
@qbey I had to change things a little because I realized the mimetype guessed by magic depends on the OS when it failed to pass CI...
With the new fixup, I stopped trying to correct incoherent extensions and will just mark the file à unsecure... not sure what we will do about it but it seems hard to do more without introducing a lot of issues for peaceful users.
If the extension is missing, I can still add.
Curious to know your thoughts on this refacto!

src/backend/core/api/serializers.py
```diff
@@ -234,8 +234,11 @@ def validate_file(self, file):
 
         # Try guessing a coherent extension from the mimetype
         if extension_mime_type != magic_mime_type:
-            if guessed_ext := mimetypes.guess_extension(magic_mime_type):
-                extension = guessed_ext[1:]
+            self.context["is_unsafe"] = True
+
+        guessed_ext = mimetypes.guess_extension(magic_mime_type)
+        if (extension is None or len(extension) > 5) and guessed_ext:
```
qbey, line 240: Why you need to check `len(extension) > 5`? Is it because no extension should be longer? Maybe a quick comment can help understand here.

qbey (approved):
I was not aware the extension guess depends on the OS, moreover for simple file types... This fix looks good to me then, I don't have a better approach in mind.

AntoLC (approved):
My review is from a frontside perspective.
Works well! Nice to have the video tag automatically added in the doc when we add a video ext.

Bad point but on the frontend side, I can see we have to implement the toolbar to be able to download the files.

[scrnli_4VrNcOng6R55mX.webm](https://github.com/user-attachments/assets/9ea21cfd-4d33-434e-99f5-7e89ae88da4d)

src/backend/core/api/serializers.py
sampaccoud: yep. It seems that 5 characters is as long as it gets. Will add comment.

PR #314, opened 2024-10-09
♻️ Add username in email

AntoLC:
## Purpose

For the moment we are using the sender's email in the email, but we have sometimes the full name of the sender.

## Proposal

Use the fullname instead of the email if it is available.

## Demo english

image

## Demo French

image


src/mail/mjml/invitation.mjml
```diff
@@ -31,7 +31,7 @@
           <!-- Main Message -->
           <mj-text>
             {% blocktrans %}
-              {{username}} invited you as an {{role}} on the following document : 
+              {{username}} {{email}} invited you as an {{role}} on the following document :
```
virgile-dev, line 34: {{email}} invited you with the role "{{role}}" on the following document :

src/backend/core/api/viewsets.py
```diff
@@ -576,8 +576,15 @@ def perform_create(self, serializer):
...
+        
+        username_sender = self.request.user.full_name
+        email_sender = self.request.user.email
+        if(not username_sender):
+            username_sender = self.request.user.email
+            email_sender = ""
+
         access.document.email_invitation(
-            language, access.user.email, access.role, self.request.user.email
+            language, access.user.email, access.role, username_sender, email_sender
```
sampaccoud, line 587: Semantically, shouldn't it be: 
```suggestion
            language, access.user.email, access.role, sender_name, sender_email
```

The term "username" is usually used with another meaning.

src/backend/core/models.py
```diff
@@ -520,7 +520,7 @@ def get_abilities(self, user):
             "versions_retrieve": can_get_versions,
         }
 
-    def email_invitation(self, language, email, role, username_sender):
+    def email_invitation(self, language, email, role, username_sender, email_sender):
```
sampaccoud, line 523:
```suggestion
    def email_invitation(self, language, email, role, username_sender, sender_email):
```

src/mail/mjml/invitation.mjml
sampaccoud: why do we want to add the email here?

src/backend/core/api/viewsets.py
```diff
@@ -778,6 +785,13 @@ def perform_create(self, serializer):
         invitation = serializer.save()
 
         language = self.request.headers.get("Content-Language", "en-us")
+
+        username_sender = self.request.user.full_name
+        email_sender = self.request.user.email
```
sampaccoud, line 790:
```suggestion
        sender_name = self.request.user.full_name
        sender_email = self.request.user.email
```

src/backend/core/api/viewsets.py
```diff
@@ -576,8 +576,15 @@ def perform_create(self, serializer):
         """Add a new access to the document and send an email to the new added user."""
         access = serializer.save()
         language = self.request.headers.get("Content-Language", "en-us")
+        
+        username_sender = self.request.user.full_name
+        email_sender = self.request.user.email
```
sampaccoud, line 581:
```suggestion
        sender_name = self.request.user.full_name
        sender_email = self.request.user.email
```

src/mail/mjml/invitation.mjml
virgile-dev: It's handy in order to know from which organization the person is coming from.
It's also a unique identifier. 
It's a good practice I catched from gdocs.

sampaccoud (approved):

PR #329, opened 2024-10-11
Fix dysfunctional permissions on document create

sampaccoud:
## Purpose

When creating a document access, users are benefiting on the targeted document from the highest access right they have among all documents.

## Proposal

- [x] Add missing filtering on the document ID when retrieving the role of the logged-in user
- [x]  Improve all tests to secure this issue


AntoLC (approved):

qbey (approved):
I find it a bit confusing that the permission is checked inside the serializer, otherwise it looks good to me :)

sampaccoud:
> I find it a bit confusing that the permission is checked inside the serializer, otherwise it looks good to me :)

Good point. I will open an issue to consider refactoring.

PR #341, opened 2024-10-15
♻️(backend) Compress images after upload

lindenb1:
## Purpose

To ensure that uploaded images are appropriately validated and compressed before being stored. This will prevent oversized images from exceeding storage limits and improve loading times by reducing image file sizes.

## Proposal

1. Integrate file validation for both type and size in the `FileUploadSerializer` to ensure only allowed images are accepted.
2. Apply compression to images that exceed size constraints, aiming to optimize storage usage.

- [x] Validate file type and size in `FileUploadSerializer`.
- [x] Implement image compression within the validate_file method if file type is allowed.
- [x] Add checks to ensure the compressed image also meets the size constraints before storing.
- [x] Document the maximum file size and allowed types for the upload endpoint.


src/backend/core/api/serializers.py
```diff
@@ -226,6 +229,23 @@ def validate_file(self, file):
...
+            original_extension = os.path.splitext(file.name)[1]
+            image = Image.open(file)
+            image_io = BytesIO()
+            image.save(image_io, format=image.format, quality=85, optimize=True)
+            image_io.seek(0)
+            image_io.name = f"compressed{original_extension}"
+            file = image_io  # Replace the file with the compressed version
+
+            # Validate compressed file size
+            if file.getbuffer().nbytes > settings.DOCUMENT_IMAGE_MAX_SIZE:
```
sampaccoud, line 243: I don't think we want to reject an image just for its size as long as it is compressed.
What is missing is probably to resize the image in number of pixels and keep only what we deem necessary for the use cas of docs? We should probably use a library to do this?

lindenb1: @sampaccoud In that case my Idea would be to simply add another step to limit the resolution before the compression, as "from PIL import Image" can already do this for us, there should be no need for another lib in my opinion.

What do you think of this instead?

```python3
class FileUploadSerializer(serializers.Serializer):
    """Receive file upload requests with validation, resizing, and compression."""

    file = serializers.FileField()

    def validate_file(self, file):
        """Validate, resize, and compress the file if it’s an image."""
        # Validate initial file size
        if file.size > settings.DOCUMENT_IMAGE_MAX_SIZE:
            max_size = settings.DOCUMENT_IMAGE_MAX_SIZE // (1024 * 1024)
            raise serializers.ValidationError(
                f"File size exceeds the maximum limit of {max_size:d} MB."
            )

        # Validate file type
        mime_type, _ = mimetypes.guess_type(file.name)
        if mime_type not in settings.DOCUMENT_IMAGE_ALLOWED_MIME_TYPES:
            mime_types = ", ".join(settings.DOCUMENT_IMAGE_ALLOWED_MIME_TYPES)
            raise serializers.ValidationError(
                f"File type '{mime_type:s}' is not allowed. Allowed types are: {mime_types:s}"
            )

        # Resize and compress the image if it’s a valid image type
        if mime_type and mime_type.startswith("image/"):
            original_extension = os.path.splitext(file.name)[1]
            image = Image.open(file)

            # Resize the image
            max_width, max_height = settings.DOCUMENT_IMAGE_MAX_DIMENSIONS
            image.thumbnail((max_width, max_height))

            # Compress the resized image
            image_io = BytesIO()
            image.save(image_io, format=image.format, quality=85, optimize=True)
            image_io.seek(0)
            image_io.name = f"compressed{original_extension}"
            file = image_io  # Replace the file with the resized and compressed version

            # Validate resized and compressed file size
            if file.getbuffer().nbytes > settings.DOCUMENT_IMAGE_MAX_SIZE:
                max_size = settings.DOCUMENT_IMAGE_MAX_SIZE // (1024 * 1024)
                raise serializers.ValidationError(
                    f"Compressed file size exceeds the maximum limit of {max_size:d} MB."
                )

        return file
```

I set DOCUMENT_IMAGE_MAX_DIMENSIONS to (1920, 1080). After testing, I found that the saved image size on disk didn’t exceed ~700 KB, even when the uploaded source image had a resolution of around 6000x4000 pixels. With the proposed solution, we would save the 6000x4000 source image at a resolution of 1620×1080 plus the additional compression, also keeping the dimension intact.

We could also implement a loop to compress the image until it fits the limits. But in my opinion, with the solution proposed above, we should never even come close to 10MB ...

Please let me know what you think.

src/backend/core/api/serializers.py
```diff
@@ -226,6 +229,23 @@ def validate_file(self, file):
                 f"File type '{mime_type:s}' is not allowed. Allowed types are: {mime_types:s}"
             )
 
+        # Compress the image if it’s a valid image type
+        if mime_type and mime_type.startswith("image/"):
```
qbey, line 233: I'm not sure this test is enough, because I guess it will fail for `image/svg+xml`. `image = Image.open(file)` will probably raise an error + I'm not sure we can "compress" svg files.

Maybe we should use a 
``` python
try:
    image = Image.open(file)
except IOError:
    # not an image
else:
    # compress
```

src/backend/core/api/serializers.py
```diff
@@ -226,6 +229,23 @@ def validate_file(self, file):
...
+            original_extension = os.path.splitext(file.name)[1]
+            image = Image.open(file)
+            image_io = BytesIO()
+            image.save(image_io, format=image.format, quality=85, optimize=True)
```
qbey, line 237: This will only optimize few image formats. Should we check the original image size then use a `.resize(` to reduce the size? 

Edit: ok, this is mentioned in the following comment ^^

lindenb1: @qbey I'm not sure what's the preferred way to go as there are multiple solutions to choose from, we could directly passthrough svg images, or we could convert them to png using e.g. cairosvg

```
        # Handle SVG files using CairoSVG
        if mime_type == "image/svg+xml":
            svg_io = BytesIO()
            file.seek(0)
            cairosvg.svg2png(file_obj=file, write_to=svg_io)
            svg_io.seek(0)
            file = InMemoryUploadedFile(
                svg_io, None, f"{os.path.splitext(file.name)[0]}.png", "image/png", svg_io.getbuffer().nbytes, None
            )
```

We would then also apply the max image size and compression onto the final png created from the source svg. But I'm not sure if this is really the best way to go, as svg files have features png files don't have ... And I'm not sure if you wanted to support things like animated svg for example. As in that case, I would simply suggest to passthrough the svg file directly, just limiting the user by the file size he wants to upload (currently 10 MB).

qbey: Hi @lindenb1 ,

As I see it, I would as you said, for SVG:
> I would simply suggest to passthrough the svg file directly, just limiting the user by the file size he wants to upload (currently 10 MB).

I don't really like the idea of converting SVG to PNG. But, this actually raises the question about security here, because SVG should be sanitized also... So maybe we should accept SVG as it is, block if above 10MB, and simply add an issue to address the security problem.

Anyway, the SVG format should be addressed separately IMHO. 

What do you think @sampaccoud ?

lindenb1: @qbey I just updated my last commit to include the passthrough with svg files.

src/backend/core/api/serializers.py
sampaccoud: I don't think we want to reject an image because of its size. We can always resize it. Have you looked how we could use [easythumbnails](https://pypi.org/project/easy-thumbnails/) or equivalent? The original image would be deleted after a while and the thumbnailed image would remain?

src/backend/core/api/serializers.py
sampaccoud: As commented on the other discussion, I would try using a library like [easythumbnails](https://pypi.org/project/easy-thumbnails/) because resizing images probably has even more hidden complexities than what we raised here?

src/backend/core/api/serializers.py
qbey: My two cents: yes it would be nice to have something on the shelf, but also, we are not doing something that complex. easy-thumbnails is useful for ... thumbnails ^^ and I'm not sure we want something that complex (because it comes with a lot of thumbnail specific enhancements (see the default processors https://github.com/SmileyChris/easy-thumbnails/blob/master/easy_thumbnails/conf.py#L234-L240) and also with a Django model field to spare time. 

PIL is that thing on the shelf ;)

Under the hood, for the resize, easy thumbnails only call the PIL `resize` method (https://github.com/SmileyChris/easy-thumbnails/blob/master/easy_thumbnails/processors.py#L252).

src/backend/core/api/serializers.py
```diff
@@ -226,9 +230,36 @@ def validate_file(self, file):
...
+        if mime_type == "image/svg+xml":
+            return file  # Simply pass SVG files through without modification
+
+        # Resize and compress other image formats
+        elif mime_type and mime_type.startswith("image/"):
+            original_extension = os.path.splitext(file.name)[1]
+            image = Image.open(file)
+
+            max_width, max_height = settings.DOCUMENT_IMAGE_MAX_DIMENSIONS
+            image.thumbnail((max_width, max_height))
```
qbey, line 243: Why using the "thumbnail" method? Is it really a thumbnail we want? I'm sorry because I don't have the initial context, but from the first review I was guessing we only want to "reduce" big images, but not making thumbnails, am I right?

Also, I think, I would not try to guess mime type nor extension: I would directly make:

```python
try:
    image = Image.open(file)
except PIL.UnidentifiedImageError: 
    # not an image or a supported format
else:
    # ...
    image.resize(....
```

for sure you may try to call PIL only `if mime_type and mime_type.startswith("image/")` but anyway, I would let PIL say if it can manage the format or not (and actually PIL does not manage SVG so you would not have to make a specific exception for them).

src/backend/core/api/serializers.py
sampaccoud: so let's forget this comment on mine :ok_hand:

src/backend/core/api/serializers.py
lindenb1: @qbey 

according to your thumbnail question:

1. Aspect Ratio Preservation: Unlike some resizing methods that might distort the image to fit the specified dimensions, .thumbnail() ensures that the aspect ratio of the image is preserved.

2. Efficiency: .thumbnail() is optimized for performance. It resizes the image in place, meaning that it modifies the original image object directly without creating a new one, which can be faster and consume less memory than other approaches. The outcome is the same, it has nothing to do with a small image size or so, it still will be saved at a maximum resolution of 1920x1080 as currently set with DOCUMENT_IMAGE_MAX_DIMENSIONS.

Using PIL.UnidentifiedImageError, is a nice idea. I will implement this right away and update the PR accordingly

lindenb1: Please don't merge yet, this appears to have conflicts with recent changes from main

lindenb1:
@sampaccoud I just implemented storing the image in webp format, the gain we got by this is kinda minimal, but still present, at best we save like 30-100KB on average per image compared to the previous method that was in use. But there are still some leftover we need to address, e.g. "# Validate compressed file size", this is currently commented out.
Besides the is_unsafe flag now also brings some new questions. Currently, if there’s a mismatch between the guessed MIME type and the actual MIME type, the current condition of:

```
extension_mime_type, _ = mimetypes.guess_type(file.name)

        # Try guessing a coherent extension from the mimetype
            self.context["is_unsafe"] = True
```
 Doesn't make sense anymore, since the MIME type discrepancy would be irrelevant once the image is converted to WebP.
 
 Should we simply remove this, or do you have any other plans on that?

virgile-dev:
@lindenb1 All tests aren't passing. I can't ask for a review from our team until it's the case.

securitykernel:
> @lindenb1 All tests aren't passing. I can't ask for a review from our team until it's the case.

@virgile-dev I don't think that is an issue with the changes, but the GH actions setup has a problem to run multiple jobs, failing with the same error message:

> Run actions/create-github-app-token@v1
/home/runner/work/_actions/actions/create-github-app-token/v1/dist/main.cjs:39845
  throw new Error("Input required and not supplied: app-id");
  ^

securitykernel:
@lindenb1 Can you rebase?

lindenb1:
> @lindenb1 Can you rebase?

Has been rebased again

PR #344, opened 2024-10-16
⚗️(backend) Text to yjs base64

AntoLC:
## Purpose

We want to be able to convert text to yjs base64 to be able to save it to a document.
This format will be readable by the Blocknote editor.

## Proposal

Create the function `text_to_yjs_base64`.

## Demo

By doing `make demo FLUSH_ARGS='--no-input'` you can see the function in action.

### DIFF
--- a/src/backend/core/tests/test_utils.py
+++ b/src/backend/core/tests/test_utils.py
@@ -10,7 +10,7 @@ from django.core import mail
 
 import pytest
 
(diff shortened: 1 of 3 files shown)


AntoLC:
The initial purpose was for this feature: https://github.com/suitenumerique/docs/pull/488
We made it another way.

PR #355, opened 2024-10-18
🐛(frontend) add default toolbar buttons

AntoLC:
## Purpose

We are overriding the default toolbar to add the
markdown and ai buttons. By doing that we were
missing some default buttons that are useful depend on the block type. 

## Proposal

- [x] 🐛(frontend) add default toolbar buttons


src/frontend/apps/e2e/__tests__/app-impress/doc-editor.spec.ts
```diff
@@ -9,6 +9,35 @@ test.beforeEach(async ({ page }) => {
...
+    ).toBeVisible();
+    await expect(
+      page.locator('button[data-test="alignTextCenter"]'),
+    ).toBeVisible();
+    await expect(
+      page.locator('button[data-test="alignTextRight"]'),
+    ).toBeVisible();
+    await expect(page.locator('button[data-test="colors"]')).toBeVisible();
+    await expect(page.locator('button[data-test="unnestBlock"]')).toBeVisible();
+    await expect(page.locator('button[data-test="createLink"]')).toBeVisible();
```
PanchoutNathan, line 38: Just the add the BlockTypeSelect inside tests and good for me !

PanchoutNathan (approved):
Just one minor thing and let's go :)

PR #357, opened 2024-10-18
✨Public and restricted doc editable

AntoLC:
## Purpose

We increased the visibility options for docs:
- public
- restricted
- authenticated

We also have 2 editability options:
- readonly
- editable

The editability options are only available for public and authenticated docs.

## Proposal

- [x] ♻️(frontend) minor components update
- [x] ✨(frontend) increase doc visibility options
- [x] tests

## Demo

[scrnli_VZHQd70DS9qxw9.webm](https://github.com/user-attachments/assets/48fb17ca-933c-4ccc-8b23-186bf8240db4)

### DIFF
--- a/src/frontend/apps/e2e/__tests__/app-impress/doc-member-list.spec.ts
+++ b/src/frontend/apps/e2e/__tests__/app-impress/doc-member-list.spec.ts
@@ -164,14 +164,22 @@ test.describe('Document list members', () => {
     const shareModal = page.getByLabel('Share modal');
 
     // Admin still have the right to share
-    await expect(shareModal.getByLabel('Doc private')).toBeEnabled();
+    await expect(
+      shareModal.getByRole('combobox', {
+        name: 'Visibility',
+      }),
+    ).not.toHaveAttribute('disabled');
 
     await SelectRoleCurrentUser.click();
     await page.getByRole('option', { name: 'Reader' }).click();
     await expect(page.getByText('The role has been updated')).toBeVisible();
 
     // Reader does not have the right to share
-    await expect(shareModal.getByLabel('Doc private')).toBeDisabled();
+    await expect(
+      shareModal.getByRole('combobox', {
+        name: 'Visibility',
+      }),
+    ).toHaveAttribute('disabled');
   });
 
   test('it checks the delete members', async ({ page, browserName }) => {
--- a/src/frontend/apps/impress/src/features/docs/doc-management/components/DocVisibility.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-management/components/DocVisibility.tsx
@@ -1,16 +1,17 @@
 import {
-  Button,
-  Switch,
+  Radio,
+  RadioGroup,
+  Select,
   VariantType,
   useToastProvider,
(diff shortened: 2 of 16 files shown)


src/frontend/apps/impress/src/features/docs/doc-management/types.tsx
```diff
@@ -27,14 +27,19 @@ export enum LinkReach {
...
 export interface Doc {
   id: string;
   title: string;
   content: Base64;
   link_reach: LinkReach;
-  link_role: 'reader' | 'editor';
+  link_role: LinkRole;
   accesses: Access[];
   created_at: string;
   updated_at: string;
```
PanchoutNathan, line 45: I think this type is not only related to this feature, but can be used everywhere. I moved this file to the root of the feature/docs folder. But maybe in another PR? and maybe. And maybe you don't agree with that

src/frontend/apps/impress/src/features/docs/doc-management/components/DocVisibility.tsx
```diff
@@ -44,55 +58,83 @@ export const DocVisibility = ({ doc }: DocVisibilityProps) => {
...
+              {[
+                {
+                  label: t('Read only'),
+                  value: LinkRole.READER,
+                },
+                {
+                  label: t('Can read and edit'),
+                  value: LinkRole.EDITOR,
+                },
+              ].map((radio) => (
```
PanchoutNathan, line 120: Maybe create a constant above?

src/frontend/apps/impress/src/features/docs/doc-management/components/DocVisibility.tsx
```diff
@@ -44,55 +58,83 @@ export const DocVisibility = ({ doc }: DocVisibilityProps) => {
...
-                });
-              });
-          }}
-          color="primary"
-          icon={<span className="material-icons">copy</span>}
-        >
-          {t('Copy link')}
-        </Button>
+        {(doc.link_reach === LinkReach.AUTHENTICATED ||
+          doc.link_reach === LinkReach.PUBLIC) && (
```
PanchoutNathan, line 96: what do you think of creating a constant with explicit naming? like for example : 

`const showRestrictedOptions = doc.link_reach === LinkReach.AUTHENTICATED || doc.link_reach === LinkReach.PUBLIC`

or 

`const showRestrictedOptions = [LinkReach.AUTHENTICATED, LinkReach.PUBLIC].includes(doc.link_reach);`

src/frontend/apps/e2e/__tests__/app-impress/doc-member-list.spec.ts
```diff
@@ -164,14 +164,22 @@ test.describe('Document list members', () => {
     const shareModal = page.getByLabel('Share modal');
```
PanchoutNathan, line 164: Add tests to know if restrictions should be shown or not based on visibility

AntoLC (commented):
> ✅ Nice work ! This is a cool feature 

Yes this feature is waited since the beginning of the project. ^^

PanchoutNathan (approved):

PR #365, opened 2024-10-21
🐛(i18n) same frontend and backend language using shared cookies

### LINKED ISSUE #333: Unify frontend and backend language preference using shared cookie
## Bug Report

**Problematic behavior**
Here you can see "Doc private" appears in English but the user interface is in French. 
https://github.com/user-attachments/assets/0ae0e457-15f7-423c-b4bc-c1c5fc9d48bc

Same when I'm loading a doc I don't have access to. The message only is available in French even if I sent the interface in English and reload.
Capture d’écran du 2024-10-14 17-26-11

rvveber:
We switch to cookie based language selection, so that the django backend can pick up our selected language. In turn, error messages from the django/backend that are embedded in the frontend, will be in the correct language (as long as django has translations for it).

Also we replace utils and its test with i18next-browser-languagedetector to reliably detect the current language of any browser and enforce setting the closest of the allowed languages, initially.
The dependency has no subdependencies and is a solid replacement, it also avoids having to deal with splitting locales.

Closes #333


src/frontend/apps/impress/src/i18n/initI18n.ts
```diff
@@ -1,30 +1,32 @@
...
+      lookupCookie: LANGUAGE_COOKIE_NAME,
+      cookieMinutes: 525600, // Expires after one year
+    },
     interpolation: {
       escapeValue: false,
     },
     preload: Object.keys(LANGUAGES_ALLOWED),
-    nsSeparator: '||',
+    nsSeparator: false,
+    keySeparator: false,
```
AntoLC, line 26: These changes don't seem related to this issue, the `nsSeparator` is necessary.

src/frontend/apps/e2e/__tests__/app-impress/language.spec.ts
```diff
@@ -25,4 +25,25 @@ test.describe('Language', () => {
...
+    await header.getByRole('combobox').getByText('English').click();
+    await header.getByRole('option', { name: 'Français' }).click();
+
+    // check if cookie is updated
+    const cookies = await context.cookies();
+    const language_cookie = cookies.find(
+      (cookie) => cookie.name == 'impress_language',
+    );
+    await expect(language_cookie?.value).toBe('fr');
+  });
```
AntoLC, line 48: I think we should improve this test, it could be nice to assert that the backend is switching from 1 language to another.
If you go to this url (http://localhost:3000/docs/something/) you will generate a backend 404 error, you can check the error message to assert that the backend is synch with the frontend.

To intercept response with playwright you can check this test as an example:
https://github.com/rvveber/impress/blob/0cfcb84139452ffff9ba243f1e660bc33e1d5020/src/frontend/apps/e2e/__tests__/app-impress/doc-member-create.spec.ts#L11-L15
 
To redirect to a special page, you can do: `await page.goto('/docs/something/');`

AntoLC (commented):
It works well, it's nice.

src/frontend/apps/impress/src/i18n/initI18n.ts
rvveber: You are right, they are an improvement with the intentions of commit bfde526
I will create a separate commit to adress this

src/frontend/apps/e2e/__tests__/app-impress/language.spec.ts
rvveber: Thanks for the instructions!

src/frontend/apps/e2e/__tests__/app-impress/language.spec.ts
```diff
@@ -25,4 +25,57 @@ test.describe('Language', () => {
...
+    await header.getByRole('combobox').getByText('English').click();
+    await header.getByRole('option', { name: 'Français' }).click();
+
+    // check if cookie is updated
+    const cookies = await context.cookies();
+    const language_cookie = cookies.find(
+      (cookie) => cookie.name == 'impress_language',
+    );
+    expect(language_cookie?.value).toBe('fr');
+  });
```
AntoLC, line 48: I think we should remove this test, because it is testing the internal mechanism, and this mechanism could change if we do a refacto. What we want above all, it is that the feature works, `checks the language picker` and `checks that backend uses the language cookie` assert already that.

src/frontend/apps/e2e/__tests__/app-impress/language.spec.ts
```diff
@@ -25,4 +25,57 @@ test.describe('Language', () => {
...
+
+  test('checks that backend uses the language cookie', async ({ page }) => {
```
AntoLC, line 50:
```suggestion
  test('checks that backend uses the same language as the frontend', async ({ page }) => {
```

src/frontend/apps/impress/src/i18n/conf.ts
```diff
@@ -2,5 +2,5 @@ export const LANGUAGES_ALLOWED: { [key: string]: string } = {
   en: 'English',
   fr: 'Français',
 };
-export const LANGUAGE_LOCAL_STORAGE = 'impress-language';
-export const BASE_LANGUAGE = 'fr';
+export const LANGUAGE_COOKIE_NAME = 'impress_language';
```
AntoLC, line 5:
```suggestion
export const LANGUAGE_COOKIE_NAME = 'docs_language';
```

AntoLC (commented):
I can see you add the issue ID in the commit message, it is not optimal because it creates many entries on the issue during rebase, i think you can remove it, a commit is part of a PR, from this PR we can find the issue.

image


----

Capture d'écran 2024-10-22 133231

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to
 
 ## Fixed
 
+- 🐛(i18n) same frontend and backend language using shared cookies #333
```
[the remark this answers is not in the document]
rvveber: Oh okay, but then this also needs to be changed (in a new commit?):
```suggestion
- ✨(frontend) add buttons to copy document to clipboard as HTML/Markdown #300
+ ✨(frontend) add buttons to copy document to clipboard as HTML/Markdown #318
```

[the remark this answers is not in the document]
AntoLC: Same commit, it's fine, it is a very minor change.

AntoLC (approved):
Top! Thank you.

rvveber:
Fixed conflicts

PR #366, opened 2024-10-21
🗑️(ci) clean old deployment and ci

rouja:
We move deployment stuff to a new repository. we don't need this code anymore


AntoLC (approved):
Let's go !

Issue #328, opened 2024-10-11
Placeholder for title becomes editable when the user double clicks.


virgile-dev:
## Bug Report

**Problematic behavior**
Here is a [video](https://www.loom.com/share/22c41836a9254727b0d024d761b2011f) of the problem.

vinhyan:
Hi @virgile-dev! I'd like to give this a try please. :)

Thanks!

vinhyan:
## Expected Behavior

For a document, the title in grey indicates that it is default placeholder text. If the title is in black, it means the title has been edited and updated successfully.  
To edit the title, click once on the placeholder text:

- If the title is not edited (or is blank), the default placeholder text remains.
- If the title is edited, the new title will be updated and displayed in black.

## Actual Behavior

Everything mentioned above works as expected; however, an issue arises if the following actions are performed quickly enough that the placeholder text remains instead of the textbox becoming empty for a new title:

1. Double-clicking to highlight the placeholder text.
2. Clicking and dragging the mouse to highlight the placeholder text, then stopping outside the textbox area.

In these cases, the user is allowed to add to the existing placeholder text (in grey). After clicking elsewhere on the page (or pressing Enter), the title is updated with a success toast notification. However, it remains grey (as if it's still a placeholder), instead of becoming black as expected for an edited title.

vinhyan:
I'm also participating in Hacktoberfest 2024, it would be great if a `hacktoberfest` tag can be added to this issue. Thanks a lot! :)

virgile-dev:
> I'm also participating in Hacktoberfest 2024, it would be great if a `hacktoberfest` tag can be added to this issue. Thanks a lot! :)

Hey @vinhyan make you sure you checkout https://github.com/TypeCellOS/BlockNote/issues
We rely on them for the editing experience and they have a lot of stuff you can contribute on.
I have a couple myself ^^ : https://github.com/TypeCellOS/BlockNote/issues/created_by/virgile-dev

vinhyan:
Hi @virgile-dev! Thank you for the recommendation! I'll check them out for sure :).

Also, would you be able to add another label that is just "hacktoberfest", since the one added was not accepted for some reasons.

Thanks a lot! :)

virgile-dev:
There you go @vinhyan !

PR #370, opened 2024-10-22
🐛(frontend) fix issue where title placeholder does not turn black after editing from default

vinhyan:
## Purpose
This pull request fixes a bug (issue #328) where the document title placeholder remains grey after being edited. The fix ensures the title turns black upon editing, providing clear visual feedback that the placeholder has been replaced.

### Problem
In `DocTitle.tsx`, when the default placeholder text is highlighted and edited (mutated), `handleTitleSubmit` gets executed. Initially, the code block handling the mutated title only updates `updateDoc` with the new title. However, without updating the `titleDisplay` state, the `isUntitled` flag does not get updated properly, leaving it as `true` when it should be `false`. This results in the title being styled in grey (indicating it’s still untitled) when it should be black.

### Solution
Set the `titleDisplay` state to the new title value in the code block that handles title mutation in `handleTitleSubmit`. This ensures the `isUntitled` flag is updated correctly, allowing the title to be styled in black once edited, as expected.

https://github.com/user-attachments/assets/9ea5c30d-387f-437f-872e-5790c4e53c91


vinhyan:
Hi @AntoLC! Could you send me the invite again? Apparently it has been revoked. I'm currently on a field trip in a remote area with very limit access to the internet so won't be able to look at this until I come back home this weekend. Thanks so much!

image

AntoLC:
Thank you for your interest.
Solved here:
https://github.com/numerique-gouv/impress/pull/454/commits/a063863f7f8aa90499746324bc0a3dd2f29a0a94#diff-85a45203f721ed832b1d70ca61103363a678da2fd66754e6dbcf2e01f6bdf019R136

PR #372, opened 2024-10-22
🐛(frontend) avoid documents indexing in search engine

PanchoutNathan:
## Purpose

Some documents are available publicly (without being logged) and may thus end-up being indexed by search engine.

## Proposal

Add a meta tag in the <head> section of the page (See https://developers.google.com/search/docs/crawling-indexing/block-indexing for more information):
```
<meta name="robots" content="noindex">
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to
 - 🐛(i18n) same frontend and backend language using shared cookies #365
 - 🐛(frontend) add default toolbar buttons #355
 - 🐛(frontend) throttle error correctly display #378
+- 🐛(frontend) (frontend) avoid documents indexing in search engine #372
 
 ## Removed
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to
 ## Fixed
 
 - 🐛(frontend) add default toolbar buttons #355
+- 🐛(frontend) (frontend) avoid documents indexing in search engine #330
```
AntoLC, line 19: We usually target the PR.

```suggestion
- 🐛(frontend) (frontend) avoid documents indexing in search engine #372
```

src/frontend/apps/impress/src/pages/docs/[id]/index.tsx
```diff
@@ -22,6 +23,9 @@ export function DocLayout() {
 
   return (
     <MainLayout withoutFooter>
+      <Head>
+        <meta name="robots" content="noindex" />
+      </Head>
```
Comment (AntoLC) [inline src/frontend/apps/impress/src/pages/docs/[id]/index.tsx:28]: Is it possible to put it above MainLayout ? I know it is working but I find that disturbing to have it has a child of MainLayout.

AntoLC (approved):

Issue #373, opened 2024-10-22
🐛fix some error messages


AntoLC:
## Bug Report

To display some error messages, we are doing a condition on the error message that we get from the backend. It is error prone because the backend error message are internationalized.

## What to do

Find another way to display clean error message (code status ?).

## Code location

- [ ] Fix on AddMembers.tsx
https://github.com/numerique-gouv/impress/blob/60120852f502c40da961e07110a7cdafc5c1b3e0/src/frontend/apps/impress/src/features/docs/members/members-add/components/AddMembers.tsx#L88-L104

- [x] Fix on AIButton.tsx
https://github.com/numerique-gouv/impress/blob/60120852f502c40da961e07110a7cdafc5c1b3e0/src/frontend/apps/impress/src/features/docs/doc-editor/components/AIButton.tsx#L365-L374

PR #378, opened 2024-10-23
♻️(frontend) improve handleAIError

AntoLC:
## Purpose

To display the throttle error messages, we are doing a condition on the error message that we get from the backend.
It is error prone because the backend error message are internationalized.

It fix partially this issue: #373

### DIFF
--- a/src/frontend/apps/impress/src/features/docs/doc-editor/components/AIButton.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-editor/components/AIButton.tsx
@@ -9,14 +9,7 @@ import {
   VariantType,
   useToastProvider,
 } from '@openfun/cunningham-react';
-import {
-  PropsWithChildren,
-  ReactNode,
-  useCallback,
-  useEffect,
-  useMemo,
-  useState,
-} from 'react';
+import { PropsWithChildren, ReactNode, useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 
 import { isAPIError } from '@/api';
@@ -70,9 +63,8 @@ export function AIGroupButton() {
   const { t } = useTranslation();
   const { currentDoc } = useDocStore();
   const { data: docOptions } = useDocOptions();
-  const [languages, setLanguages] = useState<LanguageTranslate[]>([]);
 
-  useEffect(() => {
+  const languages = useMemo(() => {
     const languages = docOptions?.actions.POST.language.choices;
 
     if (!languages) {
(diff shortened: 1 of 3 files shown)


src/frontend/apps/impress/src/features/docs/doc-editor/components/AIButton.tsx
```diff
@@ -359,26 +329,21 @@ const useHandleAIError = () => {
...
-      console.error(error);
-    },
-    [toast, t],
-  );
+    toast(t('AI seems busy! Please try again.'), VariantType.ERROR);
+    console.error(error);
+  };
 
   return handleAIError;
 };
```
PanchoutNathan, line 349: Do you authorize `consoles.error`? 
You can simplify this hook :  

```
const useHandleAIError = () => {
  const { toast } = useToastProvider();

  return (error: unknown) => {
    if (isAPIError(error) && error.status === 429) {
      toast(t('Too many requests. Please wait 60 seconds.'), VariantType.ERROR);
      return;
    }

    toast(t('AI seems busy! Please try again.'), VariantType.ERROR);
  };
};
```

src/frontend/apps/impress/src/features/docs/doc-editor/components/AIButton.tsx
```diff
@@ -321,19 +265,45 @@ const AIMenuItemTranslate = ({
...
+    }
+
+    const markdown = await editor.blocksToMarkdownLossy(selectedBlocks);
+
+    try {
+      const responseAI = await requestAI(markdown);
+
+      if (!responseAI) {
+        return;
+      }
```
PanchoutNathan, line 298: I'm not a big fan of this generic way. I would explain more because Depending on the type, it can only be a string and never undefined or null. So I would try a string

```
if (responseAI === '' ) {
    return;
}
```

PanchoutNathan (approved):
No blocking things, LGTM ✅

src/frontend/apps/impress/src/features/docs/doc-editor/components/AIButton.tsx
```diff
@@ -70,9 +63,8 @@ export function AIGroupButton() {
   const { t } = useTranslation();
   const { currentDoc } = useDocStore();
   const { data: docOptions } = useDocOptions();
-  const [languages, setLanguages] = useState<LanguageTranslate[]>([]);
 
-  useEffect(() => {
+  const languages = useMemo(() => {
```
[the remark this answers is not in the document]
AntoLC: It is still doing a sorting on a dict of 100 elements.

PR #400, opened 2024-11-02
🚸(backend) improve users similarity search and sort results

### LINKED ISSUE #399: Improve suggested emails in sharing modal
## Bug Report

**Problematic behavior**
As a user when I go into the sharing panel and start typing the email of a user I want to invite, sometimes the suggestions are not relevant.
One user made the feedback that by typing exactly the email address of an existing user the suggestion would appear last in the dropdown and would lead him to select the wrong person by hitting enter.

**Expected behavior/code**
@PanchoutNathan and @sampaccoud suggested to increase the threshold of the similarity criterion to get better results.
```
class UserViewSet(
    ...
            # Filter users by email similarity
            if query := self.request.GET.get("q", ""):
-                queryset = queryset.filter(email__trigram_word_similar=query)
+                queryset = queryset.annotate(
+                    similarity=TrigramSimilarity('email', query)
+                ).filter(similarity__gt=0.3).order_by('-similarity')  # Adjust threshold as needed

```

sampaccoud:
## Purpose

In some edge cases, the domain part the email addresse is longer than the name part. Users searches by email similarity then return a lot of unsorted results.

## Proposal

- [x] We can improve this by being more demanding on similarity when the query looks like an email. 
- [x] Sorting results by the similarity score is also an obvious improvement.

At the moment, we still think it is good to propose results with a weak similarity on the name part because we want to avoid as much as possible creating duplicate users by inviting one of is many emails, a user who is already in our database.

Fixes #399

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -6,6 +6,7 @@ from urllib.parse import urlparse
 
 from django.conf import settings
 from django.contrib.postgres.aggregates import ArrayAgg
+from django.contrib.postgres.search import TrigramSimilarity
 from django.core.exceptions import ValidationError
 from django.core.files.storage import default_storage
 from django.db.models import (
@@ -156,7 +157,20 @@ class UserViewSet(
 
             # Filter users by email similarity
             if query := self.request.GET.get("q", ""):
+                # For performance reasons we filter first by similarity, which relies on an index,
+                # then only calculate precise similarity scores for sorting purposes
                 queryset = queryset.filter(email__trigram_word_similar=query)
+
+                queryset = queryset.annotate(
+                    similarity=TrigramSimilarity("email", query)
+                )
+                # When the query only is on the name part, we should try to make many proposals
+                # But when the query looks like an email we should only propose serious matches
+                threshold = 0.6 if "@" in query else 0.1
+
+                queryset = queryset.filter(similarity__gt=threshold).order_by(
+                    "-similarity"
+                )
 
         return queryset
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+## Changed
+
+- 🚸(backend) improve users similarity search and sort results #391
+
 ## [1.7.0] - 2024-10-24
 
 ## Added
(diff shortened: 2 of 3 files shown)


src/backend/core/api/viewsets.py
```diff
@@ -156,8 +157,21 @@ def get_queryset(self):
...
+                queryset = queryset.annotate(
+                    similarity=TrigramSimilarity("email", query)
+                )
+                # When the query only is on the name part, we should try to make many proposals
+                # But when the query looks like an email we should only propose serious matches
+                threshold = 0.6 if "@" in query else 0.1
+
+                queryset = queryset.filter(similarity__gt=threshold).order_by(
+                    "-similarity"
+                )
```
qbey, line 173: Note: I think this was the missing part in the previous code 👍 This should really help not displaying totally wrong email addresses. Also, maybe the `0.1` value is a bit low, but we will adapt it later.

qbey (approved):
Looks good to me, we may adapt is again later, because it may be confusing we return other "non related" email address even if the user requested a specific one (eg the `michael.johnson@example.gouv.fr` test).

It's a nice will to allow user to find "similar" emails, but I'm not sure it will be that easy, for instance 

``` python
    response = client.get("/api/v1.0/users/?q=ajohnson@other.gouv.fr")

    assert response.status_code == 200
    user_ids = [user["id"] for user in response.json()["results"]]
    assert user_ids == []  # I would have Alice here ^^' but it's too complex
```

I'm just raising the fact we should consider the tradeoff "user confusion" vs. "don't duplicate users".

AntoLC (approved):
What happen if it match 100% with the request, you get 1 or many results?

Lightly out of scope, but we now have the full name available, could be maybe nice to search on it as well.

sampaccoud:
@qbey @AntoLC You will be happy: I added these two edge cases to my test and they both pass:
```python
    response = client.get(
        "/api/v1.0/users/?q=ajohnson@example.gouv.f",
    )
    assert response.status_code == 200
    user_ids = [user["id"] for user in response.json()["results"]]
    assert user_ids == [str(alice.id)]

    response = client.get(
        "/api/v1.0/users/?q=michael.wilson@example.gouv.f",
    )
    assert response.status_code == 200
    user_ids = [user["id"] for user in response.json()["results"]]
    assert user_ids == [str(michael_wilson.id)]
```

sampaccoud:
> Lightly out of scope, but we now have the full name available, could be maybe nice to search on it as well.

Yeah I thought of it but:
- we must wait a bit more so that most users actually have it set on their user
- we will first need to add the name to what is displayed to users searching in the input field so that we are not confused why a result matches?

PR #401, opened 2024-11-04
🐛(i18n/backend) Fix/email in receiving user language

### LINKED ISSUE #323: Receive email notification in my language
## Bug Report

**Problematic behavior**
Even though my interface is in french I'm receiving notification emails in english.
Right the emails are sent in the language  setting  the sender using on the app. It should be the opposite.

rvveber:
Implements the following:
- E-mails sent for granting access are sent in the users language if it exists - else they will be sent in system default language
- On Language change in the frontend, the language attribute of the user is updated via API.
- Fixes to language related Tests
- Added helper for language related Tests

Closes #323


rvveber:
@sampaccoud 
Is it okay, if i refactor the backend language attribute to just the first two letters for the locale? eg. `fr-fr` -> `fr`
(with refactor and migration of existing users ofcourse)

Currently we have the full locale in the backend, but only translations for main languages in the frontend, and no dialects.
In the frontend we detect just `fr`/`en`/`de` regardless what the full locale was.

When i sync this up with the user via API, it would be nice if i could avoid an additional map that hard maps `en` to `en-us` and needs to be maintained additionally.

src/frontend/apps/e2e/__tests__/app-impress/common.ts
```diff
@@ -238,3 +238,39 @@ export const mockedAccesses = async (page: Page, json?: object) => {
...
+        resp.url().includes('/user') && resp.request().method() === 'PATCH',
+    ),
+    header.getByRole('option', { name: lang.label }).click(),
+  ]);
+
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  const updatedUserResponse = await response.json();
+
+  expect(lang.expectedLocale).toContain(updatedUserResponse.language);
+}
```
AntoLC, line 276: This part seems to be used only in `language.spec.ts`, I would let it inside this test so, at least until we need this part in another test.

src/frontend/apps/impress/src/core/AppProvider.tsx
```diff
@@ -36,6 +38,16 @@ export function AppProvider({ children }: { children: React.ReactNode }) {
...
+  useEffect(() => {
+    if (userData?.language) {
+      Object.keys(LANGUAGES_ALLOWED).forEach((key) => {
+        if (userData.language.includes(key) && i18n.language !== key) {
+          void i18n.changeLanguage(key);
+        }
+      });
+    }
+  }, [userData?.language]);
+
```
AntoLC, line 50: If you rebase from main, you will see, we have now a component 
`ConfigProvider`, better to put it here, I think you will be able to get the `LANGUAGES_ALLOWED` from the conf endpoint, you will see.
https://github.com/numerique-gouv/impress/blob/main/src/frontend/apps/impress/src/core/config/ConfigProvider.tsx

src/frontend/apps/impress/src/features/language/api/useChangeUserLanguage.tsx
```diff
@@ -0,0 +1,50 @@
...
+        value: language,
+        type: 'language',
+      }),
+    );
+  }
+
+  return response.json() as Promise<User>;
+};
+
+export function useSetUserLanguage() {
```
AntoLC, line 40: Let's keep it like the filename.

```suggestion
export function useChangeUserLanguage() {
```

src/frontend/apps/impress/src/features/language/LanguagePicker.tsx
```diff
@@ -33,43 +34,54 @@ const SelectStyled = styled(Select)<{ $isSmall?: boolean }>`
...
-        });
+        void i18n
+          .changeLanguage(e.target.value as string)
+          .then(() => {
+            setUserLanguage({ language: i18n.language }).catch((err) => {
+              console.error('Error changing users language', err);
+            });
+          })
+          .catch((err) => {
+            console.error('Error changing language', err);
```
AntoLC, line 83: Maybe add a error toast ?
```ts
toast(t('Failed to change the language'), VariantType.ERROR, {
  duration: 3000,
});
```

src/backend/core/api/viewsets.py
```diff
@@ -883,7 +882,7 @@ def perform_create(self, serializer):
         """Save invitation to a document then send an email to the invited user."""
         invitation = serializer.save()
 
-        language = self.request.headers.get("Content-Language", "en-us")
+        language = settings.LANGUAGE_CODE
```
AntoLC, line 885: There is big chances that the new user will have the same language as the sender, I will put the sender language in this case.
```suggestion
        language = self.request.user.language
```

AntoLC (commented):
Looks good ! 
A few comments.

@sampaccoud Should we add a migration to make all our current users french by default, because they will all switch to english after that, wdyt ?

src/frontend/apps/e2e/__tests__/app-impress/common.ts
rvveber: Agree, if language is involved, we should instead put the test in `language.spec.ts`

src/frontend/apps/impress/src/features/language/api/useChangeUserLanguage.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+export const setUserLanguage = async ({
+  language,
+}: SetUserLanguageParams): Promise<User> => {
+  const { userData } = useAuthStore.getState();
+
+  if (!userData?.id) {
+    console.warn('Id of user is needed for this request.');
+    return {} as Promise<User>;
+  }
```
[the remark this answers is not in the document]
rvveber: The `id`? 
Like this:
```
interface SetUserLanguageParams {
  language: User['language'];
  id: User['id'];
}

export const setUserLanguage = async ({
  language,
  id
}:... 
```

src/frontend/apps/impress/src/features/language/LanguagePicker.tsx
```diff
@@ -33,43 +39,62 @@ const SelectStyled = styled(Select)<{ $isSmall?: boolean }>`
...
 
   return (
     <SelectStyled
       label={t('Language')}
       showLabelWhenSelected={false}
       clearable={false}
       hideLabel
-      defaultValue={i18n.language}
+      value={
+        Object.keys(LANGUAGES_ALLOWED).find(
```
AntoLC, line 73: Should we use `LANGUAGES` of the conf instead ?
```
const { data: conf } = useConfig();
```

src/frontend/apps/impress/src/i18n/conf.ts
```diff
@@ -1,7 +1,7 @@
...
-  en: 'English',
-  fr: 'Français',
-  de: 'Deutsch',
+  'en-us': 'English',
+  'fr-fr': 'Français',
+  'de-de': 'Deutsch',
 };
 export const LANGUAGE_COOKIE_NAME = 'docs_language';
-export const BASE_LANGUAGE = 'en';
+export const BASE_LANGUAGE = Object.keys(LANGUAGES_ALLOWED)[0];
```
AntoLC, line 7: Are `LANGUAGES_ALLOWED` and `BASE_LANGUAGE` still necessary since we have about the same in the conf endpoint ?

src/backend/core/tests/documents/test_api_document_invitations.py
```diff
@@ -417,94 +417,11 @@ def test_api_document_invitations_create_privileged_members(
         }
 
 
-def test_api_document_invitations_create_email_from_content_language():
```
AntoLC, line 420: I think we should put back this test with the updated behavior (from sender language).

src/frontend/apps/impress/src/i18n/conf.ts
rvveber: I implemented it using the conf endpoint, 
but this had an unintended side effects.

1. All language labels come from the backend and there, they are translated.
- If the page loads in the wrong language (whatever the reason may be) finding the right language is difficult.
2. The conf endpoint has to be refetched once the language is changed to get the conf values in the right language

I really think, we should not have translation in config values.
Should i remove the translation for the languages and re-implement it via conf endpoint?

rvveber: I implemented it, and it works nicely, check it out.

src/backend/core/tests/documents/test_api_document_invitations.py
rvveber: ✅

src/frontend/apps/impress/src/features/language/LanguagePicker.tsx
rvveber: ✅

sampaccoud:
I was expecting the language to be set on the user on first logging here:
https://github.com/numerique-gouv/impress/blob/main/src/backend/core/authentication/backends.py#L91

Otherwise how and when is this field set on the user? :thinking:

src/frontend/apps/impress/src/features/language/LanguagePicker.tsx
```diff
@@ -33,44 +38,80 @@ const SelectStyled = styled(Select)<{ $isSmall?: boolean }>`
...
+    const actions: Promise<unknown>[] = [i18n.changeLanguage(targetLocale)];
+
+    if (userData?.id) {
+      actions.push(
+        changeUserLanguage({
+          userId: userData.id,
+          language: targetLocale,
+        }),
+      );
+    }
```
rvveber, line 95: > I was expecting the language to be set on the user on first logging here: https://github.com/numerique-gouv/impress/blob/main/src/backend/core/authentication/backends.py#L91
> 
> Otherwise how and when is this field set on the user? 🤔

It happens here.
When the user is unsatisfied with the language that is set, he switches it and in turn it updates the attribute at the user.

rvveber: @sampaccoud

src/backend/core/tests/documents/test_api_document_invitations.py
```diff
@@ -417,11 +417,11 @@ def test_api_document_invitations_create_privileged_members(
...
+def test_api_document_invitations_create_email_from_senders_language():
     """
-    The email generated is from the language set in the Content-Language header
+    The email generated is from the language set on the sending user
```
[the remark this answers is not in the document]
rvveber: ✅

src/frontend/apps/impress/src/features/language/LanguagePicker.tsx
rvveber: I thought users are created through the Identity Provider? Imo the IdP should transmit the preferred language for initial Account creation. 

It is not garantueed that the user has visited the frontend before he logs in/creates an account, is it?

sampaccoud: Ok. It could be one possibility but it's not in the case of our idP. If it came from the idP it would have to be added here right as well right?
If it's not set by the idP, you sould set it from the request at the moment of user creation ie on first loggging like I propose?

rvveber: Ok, i will try to implement it on monday

rvveber: Unfortunately, i did not find an elegant way, with the time i had.
I'm in the holidays now and will revist this in january 2025.

sampaccoud: You are in [this authenticate method](https://github.com/mozilla/mozilla-django-oidc/blob/main/mozilla_django_oidc/auth.py#L284) called with `request` as parameter so you can get the preferred language as you were doing elsewhere?

rvveber: @sampaccoud check it out

AntoLC (commented):
A few comments but it seems good to me, all seems working fine.

We try to not mix backend and frontend in the same commit, to be easier for reviewers, some of us are specialized in backend, other in front.

src/frontend/apps/impress/src/features/left-panel/components/LefPanelTargetFilters.tsx
```diff
@@ -23,25 +23,23 @@ export const LeftPanelTargetFilters = () => {
...
+      icon: 'lock',
+      label: t('My docs'),
+      targetQuery: DocDefaultFilter.MY_DOCS,
+    },
+    {
+      icon: 'group',
+      label: t('Shared with me'),
+      targetQuery: DocDefaultFilter.SHARED_WITH_ME,
+    },
+  ];
```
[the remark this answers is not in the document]
rvveber: Function `t` was not changing, when language was updated automatically to whatever is defined on the user. 
Not sure why it happens, seems to be an issue of async nature.

Here is how you can reproduce it:
0. Switch back to the variant with `useMemo`
1. Log in to django backend. 
2. Go to frontend (new tab)
3. Change language to french
image
4. Go to django backend and change superusers language to any other eg. deutsch
image
5. Switch back to frontend tab and reload
image
6. Left panel still has old translation

src/frontend/apps/impress/src/core/AppProvider.tsx
```diff
@@ -41,7 +41,10 @@ export function AppProvider({ children }: { children: React.ReactNode }) {
     <QueryClientProvider client={queryClient}>
       <CunninghamProvider theme={theme}>
         <ConfigProvider>
-          <Auth>{children}</Auth>
+          <Auth>
+            <LanguageInitializer />
```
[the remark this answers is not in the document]
rvveber: It can work if i make LanguageInitializer an async hook as you suggested below.
working on it

src/frontend/apps/impress/src/features/language/LanguageInitializer.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { useEffect, useState } from 'react';
+
+import { useConfig } from '@/core';
+import { useAuthQuery } from '@/features/auth/api/useAuthQuery';
+import { useChangeUserLanguage } from '@/features/language/api/useChangeUserLanguage';
+import { availableFrontendLanguages } from '@/i18n/initI18n';
+
+import { getMatchingLocales } from './utils/locale';
+
+const LanguageInitializer = () => {
```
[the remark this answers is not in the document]
rvveber: Sounds good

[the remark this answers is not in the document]
rvveber: ✅

src/frontend/apps/impress/src/features/language/LanguageInitializer.tsx
```diff
@@ -0,0 +1,66 @@
+import i18n from 'i18next';
```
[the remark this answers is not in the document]
rvveber: ✅

src/frontend/apps/impress/src/core/AppProvider.tsx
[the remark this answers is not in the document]
rvveber: ✅

src/frontend/apps/e2e/__tests__/app-impress/language.spec.ts
```diff
@@ -52,15 +69,83 @@ test.describe('Language', () => {
...
+    await expect(page.getByText('Titres', { exact: true })).toBeVisible();
+    await header.click();
+    await expect(page.getByText('Titres', { exact: true })).toBeHidden();
+  });
 });
+
+test.afterEach(async ({ page }) => {
+  // Switch back to English - important for other tests to run as expected
+  await waitForLanguageSwitch(page, TestLanguage.English);
+});
```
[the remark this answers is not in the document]
rvveber: ✅

src/frontend/apps/e2e/__tests__/app-impress/language.spec.ts
```diff
@@ -52,15 +69,83 @@ test.describe('Language', () => {
...
+  const header = page.locator('header').first();
+  await header.getByRole('button', { name: 'arrow_drop_down' }).click();
+
+  const [response] = await Promise.all([
+    page.waitForResponse(
+      (resp) =>
+        resp.url().includes('/user') && resp.request().method() === 'PATCH',
+    ),
+    page.getByLabel(lang.label).click(),
+  ]);
```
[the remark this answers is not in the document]
rvveber: ✅

virgile-dev:
@sampaccoud could you approve and merge ?

sampaccoud (approved):

sampaccoud:
@rvveber Amazing fix! Thanks for that Robin :muscle: :tada:

PR #404, opened 2024-11-07
🧑‍💻(helm) demo helm config

AntoLC:
## Purpose

We provide a helm config example. 
It is a good base to implement your own helm config.


AntoLC:
Close in favor of: #427

PR #408, opened 2024-11-08
🦺(backend) add comma to sub regex

AntoLC:
## Purpose

Some sub have comma, the regex was a bit too strict and didn't allow it, this PR fixes that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,11 @@ and this project adheres to
 - 🚸(backend) improve users similarity search and sort results #391
 - 🌐(backend) add german translation #259
 
+## Fixed
+
+- 🦺(backend) add comma to sub regex #408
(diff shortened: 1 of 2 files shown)


src/backend/core/migrations/0009_alter_user_language_and_sub.py
```diff
@@ -0,0 +1,24 @@
...
+
+import django.core.validators
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0008_alter_document_link_reach'),
+    ]
```
sampaccoud, line 11: You don't need to add a migration. This is not impacting the db schema, only validation in the Django code.

sampaccoud (approved):

PR #410, opened 2024-11-08
🔊Add sentry

### LINKED ISSUE #103: ✨Error reporting / monitoring
## Feature Request

We want to integrate Sentry with Docs, here an example PR: https://github.com/numerique-gouv/people/pull/378

AntoLC:
## Purpose

We want to be able to monitor the application. 
To do so, we add Sentry configuration, it will be added on the frontend and backend side.

## Proposal

- [x] 🧐(frontend) add sentry on the frontend side
- [x] 🧐(backend) adapt Sentry config
- [x] 🩹(backend) get current release from pyproject.toml


.github/workflows/docker-hub.yml
```diff
@@ -109,7 +110,10 @@ jobs:
           context: .
           file: ./src/frontend/Dockerfile
           target: frontend-production
-          build-args: DOCKER_USER=${{ env.DOCKER_USER }}:-1000
+          build-args: |
+            DOCKER_USER=${{ env.DOCKER_USER }}:-1000
+            SENTRY_DSN=${{ env.SENTRY_DSN }}
+            SENTRY_ENV=${{ github.event_name != 'pull_request' && github.ref == 'refs/heads/main' && 'production' || 'staging' }}
```
sampaccoud, line 116: I don't think this should be set in the CI. Sentry is to monitor our envs.

src/backend/impress/settings.py
```diff
@@ -372,7 +365,8 @@ class Base(Configuration):
     CORS_ALLOWED_ORIGIN_REGEXES = values.ListValue([])
 
     # Sentry
-    SENTRY_DSN = values.Value(None, environ_name="SENTRY_DSN")
+    SENTRY_DSN = values.Value(None, environ_name="SENTRY_DSN", environ_prefix=None)
+    SENTRY_ENV = values.Value(None, environ_name="SENTRY_ENV", environ_prefix=None)
```
sampaccoud, line 369: I would use the env so we have a one-to-one match between our environments and our events in Sentry. Is there a reason to split these 2 notions?

.github/workflows/docker-hub.yml
AntoLC: I don't see another way for the moment to distinguish the environment on the frontend side. I talk about it with Jacques, we wanted maybe to add a `config.json` for the frontend (mount point), and add these vars in it, like that we could know if the logs are from `staging` / `preprod` / `production`.

src/backend/impress/settings.py
AntoLC: It is to set them up from the helm files, to know which env is giving errors.
image

sampaccoud: On the backend this is already taken care of: https://github.com/numerique-gouv/impress/blob/main/src/backend/impress/settings.py#L527

The frontend could read it from the API as well instead of being baked in the build.

AntoLC: Not accurate enough this way, we cannot distinguish the environment if we do like that because `cls.__name__.lower()` gives us the class name, and the class name is `production` when we build the image. We need a way to know which environment raise the error.

sampaccoud: But the setting class being used is set by an environment variable `DJANGO_CONFIGURATION` which should be set so that in staging we use the `Staging` class, in preprod the `PreProduction` class and in production the `Production` class.

AntoLC:
Merged here: https://github.com/numerique-gouv/impress/pull/424

PR #411, opened 2024-11-09
Add favorite documents

sampaccoud:
## Purpose

We want users to be able to mark/unmark documents as favorite so we can keep them at hand in the interface.

## Proposal

This is done via a new action of the document API endpoint: `/api/v1.0/documents/{document_id}/favorite/`
- POST to mark as favorite
- DELETE to unmark

## Collateral

While working on this PR, I realized that most of the database queries made when getting a document list view were to include nested accesses. This detailed information about accesses in fact only necessary for the document detail view.  @PanchoutNathan do you confirm?

I introduced a specific serializer for the document list view with less fields. For a list of 20 documents with 5 accesses, we go down from 3x5x20= 300 queries to just 3 queries.


virgile-dev:
@sampaccoud on the index we need info about the access right as we display a sharing icon with the number of people on the doc. When you click on it we even display modal where you can edit access settings.

PanchoutNathan:
> @sampaccoud on the index we need info about the access right as we display a sharing icon with the number of people on the doc. When you click on it we even display modal where you can edit access settings.

Yes but we just use the ``documents/${docId}/accesses/?page=${page}`` route, so we can delete accesses info on the document list route

qbey (commented):
Two main things to discuss here:
 - The boolean field on the Favorite model, I would like to understand the use case where it's needed
 - Regarding the viewset/serializer, I'm not very happy with the `is_user_favorite` method field, because it allows future dev to make N+1 queries without knowing it (because we don't test the query number everywhere). If I was doing it, I would more probably let the serializer fail if the annotated field is not present.

src/backend/core/api/viewsets.py
```diff
@@ -362,6 +325,34 @@ def get_serializer_class(self):
...
+
+        if not user.is_authenticated:
+            return queryset
+
+        # Annotate the queryset with the logged-in user id only if she liked the document
+        favorite_subquery = models.DocumentFavorite.objects.filter(
+            document_id=OuterRef("pk"), user=user, is_favorite=True
+        ).values("user_id")
+        queryset = queryset.annotate(
+            favorited_by_user_id=Subquery(favorite_subquery[:1])
```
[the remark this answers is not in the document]
qbey: Also, I guess using the 'Exists' will allow you to sort the favorited ones first.

sampaccoud:
> Yes but we just use the `documents/${docId}/accesses/?page=${page}` route, so we can delete accesses info on the document list route

@PanchoutNathan so that means I should delete them also on the detail view right? The current implementation doesn't scale if there are many accesses as we need to paginate them. 
It would be nice if you could fetch accesses only when the user clicks on the "Share" button.

src/backend/core/api/viewsets.py
```diff
@@ -686,15 +737,35 @@ class TemplateViewSet(
...
-    resource_field_name = "template"
     queryset = models.Template.objects.all()
 
+    def get_queryset(self):
+        """Custom queryset to get user related templates."""
+        queryset = super().get_queryset()
+        user = self.request.user
+
+        if not user.is_authenticated:
+            return queryset
```
qbey, line 756: Should you filter on public templates only? Because like this you may be able to get any template anonymously

src/backend/core/api/viewsets.py
```diff
@@ -686,15 +737,35 @@ class TemplateViewSet(
...
         permissions.AccessPermission,
     ]
+    ordering = ["-created_at"]
+    ordering_fields = ["created_at", "updated_at", "title"]
     serializer_class = serializers.TemplateSerializer
-    access_model_class = models.TemplateAccess
-    resource_field_name = "template"
     queryset = models.Template.objects.all()
 
+    def get_queryset(self):
```
qbey, line 750: I guess you need to clean the ˋlist` action because you filter objects here already, no?

src/backend/core/api/viewsets.py
```diff
@@ -494,6 +508,44 @@ def link_configuration(self, request, *args, **kwargs):
...
+        """
+        # Check permissions first
+        document = self.get_object()
+        user = request.user
+
+        if request.method == "POST":
+            # Try to mark as favorite
+            _, created = models.DocumentFavorite.objects.get_or_create(
+                document=document, user=user
+            )
```
qbey, line 524: Nit: For performance, and because you don't need to actually get the data from the database, I would rather do directly a try/except to create the object.

sampaccoud: agreed!

src/backend/core/api/viewsets.py
sampaccoud: I'm not trying to handle permissions here because it's done in the permission class : 
This was done following one of your feedbacks a few weeks ago where you requested to split more clearly what is permission and what is annotations (queryset preparation).

More specifically, this is the code that will block anonymous users from accessing any template: https://github.com/numerique-gouv/impress/blob/main/src/backend/core/models.py#L679

I'm confident that anonymous users can't retrieve any template thanks to the test suite e.g https://github.com/numerique-gouv/impress/blob/main/src/backend/core/tests/templates/test_api_templates_retrieve.py#L38

src/backend/core/api/viewsets.py
sampaccoud: I didn't find any optimization to make. Filtering in get_queryset is what should apply to any actions and the filtering here is what should apply only to the `list` action.

src/backend/core/api/viewsets.py
qbey: True ^^

qbey (approved):

src/backend/core/api/viewsets.py
```diff
@@ -332,6 +333,9 @@ def get_queryset(self):
         queryset = super().get_queryset()
         user = self.request.user
 
+        # Annotate the number of accesses associated with each document
+        queryset = queryset.annotate(nb_accesses=Count("accesses"))
```
AntoLC, line 337: The `nb_accesses` is too big on some documents, adding distinct seems to solve the problem.
```suggestion
        queryset = queryset.annotate(nb_accesses=Count("accesses", distinct=True))
```

AntoLC (commented):
This changes break this part, should we remote it ?
b9738895-e202-4c0d-82e4-ffe2a018232f

sampaccoud:
Closed in favor of https://github.com/numerique-gouv/impress/pull/429 to group related changes.

PR #412, opened 2024-11-09
✨(email) white brand email

AntoLC:
## Purpose

The email was branded "La Suite Numérique", we updated the template to make it generic, 

## Proposal

We will use settings env variables to customize the email for each brand.


src/mail/mjml/invitation.mjml
```diff
@@ -3,19 +3,14 @@
...
+      <mj-section>
         <mj-column>
           <mj-image
             align="center"
-            src="{{domain}}/assets/logo-suite-numerique.png"
+            src="{{logo_img}}"
             width="250px"
             align="left"
-            alt="{%trans 'La Suite Numérique' %}"
+            alt="{%trans 'Logo email' %}"
```
sampaccoud, line 13: I think you should now test the presence of the brand and logo in the email sent.
You should be able to do it in this existing test `test_api_document_invitations_create_privileged_members`

sampaccoud (approved):

PR #422, opened 2024-11-13
⚡️(docker) improve y-provider image

AntoLC:
## Purpose

Improve y-provider image by having the `node_modules` as small as possible.

## Proposal

image


PanchoutNathan (approved):
GG :) I had the problem, and cherry picking was committed, that solved the problem

PR #423, opened 2024-11-14
✨(frontend) add styledCss props to Box component

PanchoutNathan:
## Purpose

In order to facilitate DX and not to use a string in the code for the css. We add the $styledCss props to the Box component. This object comes from Styled component


AntoLC (approved):
As said together, let's reuse $css (`$css?: string | RuleSet<object>;`) and go ^^

PR #424, opened 2024-11-14
🔧(frontend) Runtime config

AntoLC:
## Purpose

We want to be able to pass environment variables at runtime to the frontend. 

To do so, we created a public endpoint on the backend to share some configurations (see: https://github.com/numerique-gouv/impress/pull/425). 
The frontend will get the conf, then adapt itself depend of them.

## Proposal

- [x] ✨(frontend) add config provider
- [x] ✨(frontend) add sentry
- [x] ♻️(frontend) get theme from config endpoint
- [x] ♻️(frontend) get media url from config endpoint
- [x] ♻️(frontend) get collaboration server url from config endpoint
- [x] Tests


src/backend/core/api/viewsets.py
```diff
@@ -886,3 +889,20 @@ def perform_create(self, serializer):
...
+
+    def get(self, request):
+        """
+        GET /api/v1.0/config/
+            Return a dictionary of public settings.
+        """
+        array_settings = ["SENTRY_DSN", "DJANGO_CONFIGURATION"]
+        dict_settings = {}
+        for setting in array_settings:
+            dict_settings[setting] = os.environ.get(setting, None)
```
sampaccoud, line 906: I think it's better not to shortcircuit Django Configuration settings so that we are forced to declare our settings properly and document them. This would be a bit too wild IMO.
Also we want to keep a common pattern with our other projects e.g https://github.com/numerique-gouv/people/blob/main/src/backend/core/api/viewsets.py#L501

src/backend/core/api/viewsets.py
```diff
@@ -886,3 +889,20 @@ def perform_create(self, serializer):
...
+    """API ViewSet for sharing some public settings."""
+
+    permission_classes = [AllowAny]
+
+    def get(self, request):
+        """
+        GET /api/v1.0/config/
+            Return a dictionary of public settings.
+        """
+        array_settings = ["SENTRY_DSN", "DJANGO_CONFIGURATION"]
```
sampaccoud, line 903: A great place to expose settings.LANGUAGE _CODE and settings.LANGUAGES so that the frontend starts with the same choice of languages and default language as the backend!
Any other settings that are duplicated between frontend and backend and that we could harmonize here?

src/backend/core/api/viewsets.py
AntoLC: I separated the backend in this PR: https://github.com/numerique-gouv/impress/pull/425

src/backend/core/api/viewsets.py
AntoLC: I separated the backend in this PR: https://github.com/numerique-gouv/impress/pull/425

PanchoutNathan (approved):
LGTM :)

PR #425, opened 2024-11-15
✨(backend) config endpoint

AntoLC:
## Purpose

We want to be able to pass environment variables at runtime to the frontend.
To do so, we created a public endpoint to share the needed configurations.

## Proposal

- [x] ✨(backend) add public endpoint /api/v1.0/config/
- [x] 🔧(backend) add MEDIA_BASE_URL setting
- [x] 🔧(backend) add COLLABORATION_SERVER_URL setting
- [x] 🔧(backend) add FRONTEND_THEME setting
- [x] 🩹(backend) get current release from pyproject.toml
- [x] ⬆️(CI) bump python version in backend test


src/backend/impress/settings.py
```diff
@@ -104,6 +104,7 @@ class Base(Configuration):
     STATIC_ROOT = os.path.join(DATA_DIR, "static")
     MEDIA_URL = "/media/"
     MEDIA_ROOT = os.path.join(DATA_DIR, "media")
+    MEDIA_BASE_URL = values.Value(None, environ_name="MEDIA_BASE_URL", environ_prefix=None)
```
sampaccoud: Isn't it more generally the `API_BASE_URL` ?

env.d/development/common.dist
```diff
@@ -21,6 +21,7 @@ STORAGES_STATICFILES_BACKEND=django.contrib.staticfiles.storage.StaticFilesStora
 AWS_S3_ENDPOINT_URL=http://minio:9000
 AWS_S3_ACCESS_KEY_ID=impress
 AWS_S3_SECRET_ACCESS_KEY=password
+MEDIA_BASE_URL=http://localhost:8083
```
sampaccoud, line 24: nitpick: technically the previous commit already introduced something about this setting which should only arrive in the commit.

env.d/development/common.dist
```diff
@@ -45,3 +45,6 @@ OIDC_AUTH_REQUEST_EXTRA_PARAMS={"acr_values": "eidas1"}
 AI_BASE_URL=https://openaiendpoint.com
 AI_API_KEY=password
 AI_MODEL=llama
+
+# Collaboration
+COLLAB_SERVER_URL=ws://localhost:4444
```
sampaccoud, line 50: I wouldn't shorten it:
COLLABORATION_SERVER_URL

.github/workflows/impress.yml
```diff
@@ -199,7 +201,7 @@ jobs:
       - name: Install Python
         uses: actions/setup-python@v3
         with:
-          python-version: "3.10"
+          python-version: "3.12.6"
```
sampaccoud, line 204: Should be a separate commit about upgrading to python 3.12? If you do this you also need to upgrade the python version in the Docker File and in the src/backend/pyproject.toml file!

AntoLC: The Docker file is already in `3.12.6`, it is the test not testing on the good version:
Capture d'écran 2024-11-18 122202

env.d/development/common.dist
AntoLC: Done: https://github.com/numerique-gouv/impress/pull/425/commits/f10740acdc673888a3fa18f1572e95ded75200b4

.github/workflows/impress.yml
AntoLC: [⬆️(CI) bump python version in backend test](https://github.com/numerique-gouv/impress/pull/425/commits/77f3b21e75e7a529bfc34afa73e3bc750131c7aa)

AntoLC: And what I done is bumped the required version to 3.11 in `pyproject.toml` since `tomllib` is available ony from the 3.11.
https://github.com/numerique-gouv/impress/pull/425/commits/68e6404b2bcc6ae88d2d581b6dd359b73cbea2f8#diff-3c0d9933efa94182a3acbfeec8f5b1075f72debdee61ca34232ad6421153b8d1R26

Seems good ?

sampaccoud (approved):

PR #429, opened 2024-11-18
Add creator field to document and allow filtering

sampaccoud:
## Purpose

We are currently working on version 2 of Docs' UI. The new UI requires more filters on the documents API endpoint.

In particular, we want to be able to limit the documents displayed on a logged-in user's list view by the documents they created or by the documents that other users created. This is different from having the "owner" role on a document because this can be acquired and even lost. What we want here is to be able to identify documents by the user who created them so **we had to add a new field**.

## Proposal

Add all filters required for the new UI and corresponding tests:
- [x] ✨ annotate number of accesses on documents in list view
- [x] ✨ allow users to mark/unmark documents as favorite
- [x] ✨ allow filtering documents by their titles (basic search while waiting for the fulltext search engine)
- [x] ✨ allow filtering documents by link reach
- [x] ✨ allow filtering by documents marked as favorite
- [x] ✨add creator field on document and allow filtering on it

    
While working on this PR, I identified a flaky test and a db query optimization that I fixed:
- [x] 🐛 fix flaky test by clarifying user ordering
- [x] ⚡️optimize number of queries on document list view


src/backend/core/api/filters.py
```diff
@@ -0,0 +1,61 @@
...
+
+    # pylint: disable=unused-argument
+    def filter_is_creator_me(self, queryset, name, value):
+        """
+        Filter documents based on the `creator` being the current user.
+
+        Example:
+            - /api/v1.0/documents/?is_creator_me=true
+                → Filters documents created by the logged-in user
+            - /api/v1.0/documents/?creator_is_me=false
```
AntoLC, line 37:
```suggestion
            - /api/v1.0/documents/?is_creator_me=false
 ```

File: src/backend/core/api/viewsets.py [hunk 2]
```diff
@@ -308,9 +312,23 @@ class DocumentViewSet(
...
-    """Document ViewSet"""
+    """
+    Document ViewSet for managing documents.
+
+    Provides endpoints for creating, updating, and deleting documents,
+    along with filtering options.
+
+    Filtering:
+        - `creator=me`: Returns documents created by the current user.
+        - `creator=other`: Returns documents created by other users.
```
AntoLC, line 323: It is `is_creator_me`, isn't it ?

src/backend/core/tests/documents/test_api_documents_list.py
```diff
@@ -532,3 +534,40 @@ def test_api_documents_list_filter_link_reach_invalid():
...
+@pytest.mark.parametrize("query,nb_results", [
+    ("Project Alpha", 1),       # Exact match
+    ("project", 2),             # Partial match (case-insensitive)
+    ("Guide", 1),               # Word match within a title
+    ("Special", 0),             # No match (nonexistent keyword)
+    ("2024", 2),                # Match by numeric keyword
+    ("", 5),                    # Empty string
+])
+def test_api_documents_list_filter_link_reach(query, nb_results):
+    """Authenticated users should be able to search documents by their title."""
```
AntoLC, line 548: Will you be able to find any public or authenticated document as well? I guess it should not.

src/backend/core/api/filters.py
```diff
@@ -0,0 +1,61 @@
...
+                → Filters documents created by other users
+        """
+        user = self.request.user
+
+        if not user.is_authenticated:
+            return queryset
+
+        clause = "filter" if value else "exclude"
+        return getattr(queryset, clause)(creator=user)
```
qbey, line 46: nit: I prefer explicit code:

``` python
if value:
    return queryset.filter(creator=user)
else:
    return queryset.exclude(creator=user)
```

src/backend/core/api/filters.py
```diff
@@ -0,0 +1,61 @@
...
+        """
+        user = self.request.user
+
+        if not user.is_authenticated:
+            return queryset
+
+        clause = "filter" if value else "exclude"
+        return getattr(queryset, clause)(
+            favorited_by_users__user=user, favorited_by_users__is_favorite=True
+        )
```
qbey: I guess this will need refacto after previous PR is merged (`is_favorite` is not here anymore i guess)

qbey: Ok, removed in next commit.

src/backend/core/tests/documents/test_api_documents_list.py
qbey: As I understand it, the search is restricted to the current "list" view, so you will not have public/restricted results here. 

Anyway, it could be explicitly tested here, just to be sure ^^

qbey (approved):

src/backend/core/migrations/0011_populate_creator_field_and_make_it_required.py
```diff
@@ -0,0 +1,52 @@
...
+    the `creator` field required.
+    """
+    Document = apps.get_model("core", "Document")
+    DocumentAccess = apps.get_model("core", "DocumentAccess")
+
+    # Update `creator` using the "owner" role
+    owner_subquery = DocumentAccess.objects.filter(
+        document=OuterRef('pk'),
+        user__isnull=False,
+        role='owner',
```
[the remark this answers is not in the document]
sampaccoud: Yes, and we have no better guess...

src/backend/core/tests/documents/test_api_documents_list.py
sampaccoud: you are filtering only among the results you can see if this is the question. Is it the question?

src/backend/core/migrations/0011_populate_creator_field_and_make_it_required.py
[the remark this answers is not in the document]
AntoLC: If you remove the role filter, you will get the first user access, that is the most likely the "creator", but by thinking more about it, we should let it like that because we have a case where the "creator" can be removed from a doc, and so better to take the first owner.

src/backend/core/tests/documents/test_api_documents_list.py
sampaccoud: I added this test to secure that filtering and permissions don't interfere: https://github.com/numerique-gouv/impress/pull/429/commits/1e1fce5fa75e1760f619d76d34794f23e12411af

PR #443, opened 2024-11-22
🔒️(backend) whitelist the pod's IP address

AntoLC:
## Purpose

In a Kubernetes environment, we need to whitelist the pod's IP address to allow health checks to pass. This ensures that Kubernetes liveness and readiness probes can access the application to verify its health.

Based on this commit: https://github.com/numerique-gouv/meet/commit/d965ee706014b6395fdb2d2ce0f3e4f520f64503

### DIFF
--- a/src/backend/impress/settings.py
+++ b/src/backend/impress/settings.py
@@ -12,6 +12,7 @@ https://docs.djangoproject.com/en/3.1/ref/settings/
 
 import os
 import tomllib
+from socket import gethostbyname, gethostname
 
 from django.utils.translation import gettext_lazy as _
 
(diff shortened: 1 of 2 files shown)


src/backend/impress/settings.py
```diff
@@ -649,7 +650,10 @@ class Production(Base):
     """
 
     # Security
-    ALLOWED_HOSTS = values.ListValue(None)
+    ALLOWED_HOSTS = [
+        *values.ListValue([], environ_name="ALLOWED_HOSTS"),
+        gethostbyname(gethostname()),
```
sampaccoud, line 655: add a comment here as well so we know why it's done?

sampaccoud (approved):

PR #466, opened 2024-11-30
Sheets Prototype

sylvinus:
That's a nice Docs you have there. What if we made it better with Sheets?

This is a quick proof-of-concept for a simple spreadsheet integration, using the [IronCalc](https://www.ironcalc.com/) library.

I don't expect this to be merged any time soon, but it might be good food for thought or high-level demos ;)

It looks like this:

Screenshot 2024-11-30 at 15 54 31

I didn't implement document persistence or Y.Doc collaboration yet but this could easily be added to this POC through the same methods as text documents.

There would be a few roadblocks before it could be actually useable:
 - IronCalc is not mature enough yet. It has a main (single?) developer in Berlin. However I believe it's a potentially strong technical foundation given its Rust=>WASM + Canvas architecture, fully client-side.
 - Better build and encapsulation of IronCalc. I vendored it a bit agressively in the main app, not sure it's ideal.
 - UX improvements: how much would we want to differentiate Docs from Sheets, including in URLs?
 - Alternatively, could IronCalc be wrapped as a BlockNote.js block? It would allow including spreadsheets nicely in docs, but would the "main" single-spreadsheet usecase suffer? Could we do both?
 - There is early support in IronCalc for XLSX export and import, but it would have to be heavily battle-tested.
 - PDF export support
 - Explaning how this is different from Grist and why we'd have both in the Suite.

Let me know what you think! ;-)


AntoLC:
That it's quite nice 😄 
For sure we will have a look at it !

virgile-dev:
Great to see you experimenting with Docs :)
Definitely super interesting to think that what we built in Django for user management and permissions could be reused to create other type of documents.Of course, we need to think of all the roadblocks that you mention above.
Another approach for this could be to have a dedicated app for spreasheets using the same framework. Could be the same code base and a feature switch.

PR #472, opened 2024-12-03
Collaboration access improvement

AntoLC:
## Purpose

Access improvement on the collaboration server.

## Proposal

- [x] ✨(backend) add subrequest auth view for collaboration server
- [x] ✨(y-provider) endpoint POST /collaboration/api/reset-connections 
- [x] ✅(y-provider) add tests for y-provider server
- [x] ✨(backend) notify collaboration server
- [x] 🔧(helm) add ingress collaboration api


docker/files/etc/nginx/conf.d/default.conf
```diff
@@ -4,9 +4,58 @@ server {
...
+        proxy_set_header Host $host;
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header X-Original-URL $request_uri;
+        
+        # Prevent the body from being passed
+        proxy_pass_request_body off;
+        proxy_set_header Content-Length "";
+        proxy_set_header X-Original-Method $request_method;
+    }
```
sampaccoud, line 47: I don't see why this is necessary and not use the normal api url

docker/files/etc/nginx/conf.d/default.conf
```diff
@@ -4,9 +4,58 @@ server {
...
+
+    location  /collaboration/api/ {
+        # Collaboration server
+        proxy_pass http://y-provider:4444;
+        proxy_set_header Host $host;
+    }
```
sampaccoud, line 53: then who is doing authentication on this route? :thinking:

src/backend/core/api/viewsets.py
```diff
@@ -750,17 +820,39 @@ def perform_create(self, serializer):
...
+
+    def perform_destroy(self, instance):
+        """Delete an access to the document and notify the collaboration server."""
+        instance.delete()
+
+        # Notify collaboration server about the access removed
+        CollaborationService().reset_connections(
+            str(instance.document.id), str(instance.user.id)
+        )
+
```
sampaccoud, line 844: a signal directly on the DocumentAccess model is better place to do that because the access could be changed from elsewhere.

src/backend/impress/settings.py
```diff
@@ -372,6 +372,12 @@ class Base(Configuration):
     SENTRY_DSN = values.Value(None, environ_name="SENTRY_DSN", environ_prefix=None)
 
     # Collaboration
+    COLLABORATION_API_URL = values.Value(
+        None, environ_name="COLLABORATION_API_URL", environ_prefix=None
+    )
+    COLLABORATION_SERVER_SECRET = values.Value(
+        None, environ_name="COLLABORATION_SERVER_SECRET", environ_prefix=None
+    )
```
sampaccoud, line 380: I would find it easier to understand if these 2 settings shared the same root:
COLLABORATION_SERVER_URL 
COLLABORATION_SERVER_SECRET

src/frontend/servers/y-provider/src/server.ts
```diff
@@ -1,18 +1,142 @@
...
+      requestHeaders['x-user-id'],
+      'canEdit:',
+      canEdit,
+      'room:',
+      requestParameters.get('room'),
+    );
+
+    if (documentName !== roomParam) {
+      console.error(
+        'Invalid room name - Probable hacking attempt:',
```
sampaccoud, line 38: Why do you need the room parameter if you already have the info in documentName? :thinking:

src/frontend/servers/y-provider/src/server.ts
```diff
@@ -1,18 +1,142 @@
...
+
+  type ResetConnectionsRequestQuery = {
+    room?: string;
+  };
+
+  /**
+   * Route to reset connections in a room
+   */
+  app.post(
+    routes.RESET_CONNECTIONS,
```
sampaccoud, line 84: you need authentication on this view or it can be abused?

src/helm/impress/values.yaml
```diff
@@ -48,32 +48,62 @@ ingress:
...
-    nginx.ingress.kubernetes.io/upstream-hash-by: "$request_uri"
+    nginx.ingress.kubernetes.io/proxy-read-timeout: "86400"
+    nginx.ingress.kubernetes.io/proxy-send-timeout: "86400"
+    nginx.ingress.kubernetes.io/upstream-hash-by: $arg_room
+
+## @param ingressCollaborationApi.enabled whether to enable the Ingress or not
+## @param ingressCollaborationApi.className IngressClass to use for the Ingress
+## @param ingressCollaborationApi.host Host for the Ingress
+## @param ingressCollaborationApi.path Path to use for the Ingress
+ingressCollaborationApi:
```
sampaccoud, line 86: I didn't understand the need for another full url range. To me, the collaboration server should only be available through urls that are checked by Django as this is the only way to compute access rights. Let's not create things that we don't need yet and see later when the problem arises?

docker/files/etc/nginx/conf.d/default.conf
AntoLC: We can maybe clear a bit.
We cannot use the normal api url, when we arrive here we are in the ngnix container, so we need to call the back with its service name.

docker/files/etc/nginx/conf.d/default.conf
AntoLC: This route will be used only by the django backend, so we don't want to pass by `collaboration_auth`.

Ngnix is here only to do the "sticky" to find automatically the good pods.
The auth is inside the middlelayer `httpSecurity` on the endpoint:
https://github.com/numerique-gouv/impress/blob/c6c4eec18f5e1b7de32d838dc81ff4a8421f7de3/src/frontend/servers/y-provider/src/server.ts#L83-L88

src/backend/impress/settings.py
AntoLC: I put `COLLABORATION_API_URL` because we have as well `COLLABORATION_WS_URL`, both of them target the server somehow, we need to dissociate the base url as well because of docker compose:
```
COLLABORATION_API_URL=http://nginx:8083/collaboration/api/
COLLABORATION_WS_URL=ws://localhost:8083/collaboration/ws/
 ```
 If you think it is better, I can change, no prob.

File: src/frontend/servers/y-provider/src/server.ts (hunk 5 above)
Comment (AntoLC) [reply src/frontend/servers/y-provider/src/server.ts:38]: `roomParam` is needed for 2 things:
- we extract the document from the request_uri in `collaboration-auth` endpoint -  ([here](https://github.com/numerique-gouv/impress/pull/472/files#diff-cc27a5de6f1d8131250e4be30d0fd4c2f1a816bdc35951a9c86e6f13cd795019R49))
- We need it to stick to the good pod between clients
https://github.com/numerique-gouv/impress/pull/472/files#diff-259585b31c6635e6f972a090c26e576f7ae41f7cc7d3f65be7cf16c08fbd2f2fR80

File: src/frontend/servers/y-provider/src/server.ts (hunk 6 above)
Comment (AntoLC) [reply src/frontend/servers/y-provider/src/server.ts:84]: This route is not for the clients, it is for the backend (signal), so we don't need / want the `collaboration_auth`.
The security is from the midlelayer [httpSecurity](https://github.com/numerique-gouv/impress/pull/472/files#diff-931f2e9c315f6fc95eae032ee4f167e486cab63966647873520cb12d39d56fdfR85), the "authentication" is here:
https://github.com/numerique-gouv/impress/pull/472/files#diff-4fa4cb104ce6aac1c7d731c6a4dbd6faaece1554a0052bfd6e4cf4f7a3f769acR28-R32

The keys have to match. We can increase the secu if you want. 
This endpoint does not gives you access to the document, you could reset a connection, that will trigger a reconnection on the client side.

File: src/helm/impress/values.yaml (hunk 7 above)
Comment (AntoLC) [reply src/helm/impress/values.yaml:86]: I need it,  I really try to not create a new ingress, but with Jacques it was what we found the cleanest.
First of all, the endpoints `/collaboration/api/` is callable only by the backend (django), we don't want clients to use them, the endpoints are secure with a key known only by the 2 servers.
Second things, we don't want the backend to pass by the auth_url `collaboration-auth`, `collaboration-auth` is made to secures clients access, the backend does not have notion of accesses or abilities, it does not make sense for the backend to call again himself, to get a key that it has already.

To avoid the auth_url `collaboration-auth`, we need to have another ingress.

I try first to server to server (django / yjs server), but it is working only aleatory, when the backend connect to the good pod. The backend must connect to the same pod as the client.
Then I try to create a ConfiMaps, like that I could indicate the stickiness, and depend the road avoid the auth_url `collaboration-auth`, but it is not the good way, the ConfigMaps is commun to other service than `Impress`, it can messed up things

If you have others things in mind, curious to hear about.

File: src/frontend/servers/y-provider/src/server.ts (hunk 6 above)
Comment (sampaccoud) [reply src/frontend/servers/y-provider/src/server.ts:84]: Ok understood. Then you can remove this ingress and access the service directly via its svc. Exposing an ingress is only to give access on internet.

File: src/helm/impress/values.yaml (hunk 7 above)
Comment (sampaccoud) [reply src/helm/impress/values.yaml:86]: see above comment. We should remove this ingress.

File: docker/files/etc/nginx/conf.d/default.conf (hunk 1 above)
Comment (sampaccoud) [reply docker/files/etc/nginx/conf.d/default.conf:47]: let's peer code because I don't see very clearly how it could be by just reviewing

File: src/frontend/servers/y-provider/__tests__/server.test.ts [hunk 8]
```diff
@@ -0,0 +1,230 @@
...
+
+describe('Server Tests', () => {
+  beforeAll(async () => {
+    await new Promise<void>((resolve) => {
+      if (server.listening) {
+        resolve();
+      }
+    });
+
+    await new Promise<void>((resolve) => {
```
YousefED, line 34: you can just do `await hocuspocusServer
        .configure({
          port: portWS,
        })
        .listen()`

Personally, I try to avoid `.then` and `.catch` as much as possible and prefer `async` / `await`

src/frontend/servers/y-provider/__tests__/server.test.ts
```diff
@@ -0,0 +1,230 @@
...
+  });
+
+  test('POST /collaboration/api/reset-connections?room=[ROOM_ID] with correct API key should reset connections', async () => {
+    const response = await request(app as any)
+      .post('/collaboration/api/reset-connections?room=test-room')
+      .set('Origin', origin)
+      .set('Authorization', 'test-secret-api-key');
+
+    expect(response.status).toBe(200);
+    expect(response.body.message).toBe('Connections reset');
```
YousefED, line 96: do you want to test whether the connection has actually been reset (i.e.: users "kicked"?

src/frontend/servers/y-provider/src/server.ts
```diff
@@ -1,18 +1,142 @@
...
+        'room:',
+        room,
+      );
+
+      if (!room) {
+        res.status(400).json({ error: 'Room name not provided' });
+        return;
+      }
+
+      const docConnection = await hocuspocusServer.openDirectConnection(room);
```
YousefED, line 107: I think you don't need to call `openDirectConnection`, but it you can just access `hocuspocusServer.documents`

YousefED: Actually, there seems to be a `hocuspocusServer.closeConnections(docName)`

YousefED (commented):
Nice! Left some comments. Note that I haven't looked in depth at the Python / nginx / kubernetes code as this is not my expertise.

One suggestion about the architecture to make it safer would be to periodically reauthenticate (refresh) websocket connections (to make sure the auth is still valid) instead of 100% relying on the "kick" mechanism - but I think it's

src/frontend/servers/y-provider/src/middlelayers.ts
```diff
@@ -0,0 +1,59 @@
+import { NextFunction, Request, Response } from 'express';
```
lebaudantoine, line 1: nit: should it be `middlewares.ts`?

src/frontend/servers/y-provider/src/middlelayers.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  // Secret API Key check
+  const apiKey = req.headers['authorization'];
+  if (apiKey !== COLLABORATION_SERVER_SECRET) {
+    res.status(403).json({ error: 'Forbidden: Invalid API Key' });
+    return;
+  }
+
+  next();
+};
```
lebaudantoine, line 35: Wouldn't it be better to reuse an existing solution? Express might already provide tools for Bearer token authentication, which would save us from maintaining this ourselves.  

On a side note, should we clarify the exact type of authorization header we expect, such as Bearer?  

This can be refactored later, as the priority is to implement the feature quickly.

lebaudantoine:
Discussed IRL: The current code organization is becoming confusing. Placing server code in the frontend folder is misleading, especially now that it functions as a backend microservice.

src/backend/core/api/viewsets.py
AntoLC: I think I will do it in another PR if it is ok. It asks to modify lot of tests in the modals.

src/frontend/servers/y-provider/__tests__/server.test.ts
AntoLC: I added a check on `hocuspocusServer.closeConnections`, I don't succeed to do a full check test, I think we are in half mock env with Jest, so users don't really connect totally to the websocket, or they are kicked out because of a ping pong thing between the HocusPocus server and the HocusPocus Provider?

I have as well a e2e test on that assert the deconnection and the reconnection. I will try to came back on this tests when I will get more times.

src/frontend/servers/y-provider/__tests__/server.test.ts
AntoLC: https://github.com/numerique-gouv/impress/pull/472/commits/e7565f06795b35036580aa9ace4e3d0d81c3795b#diff-28f2602f409539020f08398a838eea589448bd58ccf86f9670f9c06f1fbeeef4R28-R30

I added this helper, to reduce promise hell:
https://github.com/numerique-gouv/impress/pull/472/commits/e7565f06795b35036580aa9ace4e3d0d81c3795b#diff-7b4e113632ac6a5668894882898d4c59b8260390aa9c61ebe6962395e4b5c94e

src/frontend/servers/y-provider/src/server.ts
AntoLC: Improved:
https://github.com/numerique-gouv/impress/pull/472/commits/096837a123543a7cda9396168e3e0ae37bdf54bf#diff-aabe2f42c0dee383e91e6ad77a11be21ddb2246bee1007ee1678dcc09c55557e

src/frontend/servers/y-provider/src/middlelayers.ts
AntoLC: Yes, on a next PR.

src/frontend/servers/y-provider/src/routes.ts
```diff
@@ -0,0 +1,4 @@
+export const routes = {
+  WS: '/collaboration/ws/',
+  RESET_CONNECTIONS: '/collaboration/api/reset-connections/',
+};
```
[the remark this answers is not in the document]
AntoLC: Not sure if it was that you suggested:
https://github.com/numerique-gouv/impress/pull/472/commits/0297052c1f2e24cbc734f72af60a78edb88aaa0b#diff-2b19f0c32ea898d6dabf592155ef17def8850d530cb8923c64cfd94b2108eaf0R2-R3

src/frontend/servers/y-provider/src/server.ts
```diff
@@ -1,18 +1,142 @@
...
+
+        const connectionUserId = connection.request.headers['x-user-id'];
+        if (connectionUserId === userId) {
+          connection.close();
+        }
+      });
+
+      await docConnection.disconnect();
+
+      res.status(200).json({ message: 'Connections reset' });
```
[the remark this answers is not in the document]
AntoLC: https://github.com/numerique-gouv/impress/pull/472/commits/0297052c1f2e24cbc734f72af60a78edb88aaa0b#diff-931f2e9c315f6fc95eae032ee4f167e486cab63966647873520cb12d39d56fdfR77-R81

src/frontend/servers/y-provider/src/routes.ts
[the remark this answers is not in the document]
lebaudantoine: no, but that's fine

src/frontend/servers/y-provider/src/middlelayers.ts
lebaudantoine: okay I am aligned !

Issue #455, opened 2024-11-26
Prometheus Metrics


lindenb1:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
I wanted to ask if there are already plans or discussions in place for bringing Prometheus metrics support upstream for the Django backend. If not, it might be worth exploring how we can expose relevant metrics to improve monitoring and observability. We could use https://github.com/korfuri/django-prometheus to implement this functionality. The referenced library is actively maintained and has worked well for me in recent projects.

**Describe the solution you'd like**
Adding Prometheus metrics to the Django backend to track key performance indicators such as request counts, latencies, error rates, and custom application metrics. If this is already planned, it would be great to know the scope and timeline. If not, perhaps we can discuss any drawbacks or prerequisites for implementing it.

**Describe alternatives you've considered**
Alternative solutions might include using third-party plugins for metrics or implementing metrics logging with other tools, but Prometheus seems like the most standardized and widely supported approach, especially when running the app on top of Kubernetes.

**Discovery, Documentation, Adoption, Migration Strategy**
For adoption, we could start with basic metrics like HTTP request stats and then expand to custom application-level metrics. This would require documentation updates to explain how to enable and query these metrics in Prometheus-compatible systems.

**Do you want to work on it through a Pull Request?**
I am happy to contribute to the effort or collaborate with others to make this happen. If anyone is already working on it, I'd appreciate any updates or pointers to relevant discussions or branches.

lindenb1:
PR available here: https://github.com/numerique-gouv/impress/pull/476

PR #476, opened 2024-12-04
✨(backend) Adding /prometheus metrics endpoint #455

lindenb1:
implements various metrics for users, documents and the actual django application

## Purpose

The purpose of this implementation is to provide detailed insights into the usage patterns and performance of the Django application. By collecting metrics for users, documents, and the application itself, developers and administrators can monitor activity, identify bottlenecks, and optimize performance. These metrics are also essential for integrating with monitoring systems like Prometheus, enabling observability and proactive issue resolution.

## Proposal

Introduce a custom Prometheus metrics collector that gathers and exposes detailed information about user activity, document usage, and the application's operational status.

## Key Features

    - Monitor user activity over different time periods (e.g., today, last 7 days, last 30 days).
    - Track document-related statistics, such as total documents, active documents, and shared documents.
    - Expose application-level metrics to ensure seamless integration with Prometheus and related monitoring tools.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to
 
 ## [Unreleased]
 
+## Added
+
+✨(backend) Adding /prometheus metrics endpoint #455
 
 ## [1.8.2] - 2024-11-28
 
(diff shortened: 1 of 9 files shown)


lindenb1:
Will reopen in new PR

PR #477, opened 2024-12-04
:wrench:(helm) add option to disable default tls setting

dominikkaminski:
## Purpose

This merge requests sets an option for those who uses impress with a different secretName in ingress.

This defaults does not change the current deployments.

### DIFF
--- a/src/helm/impress/templates/ingress.yaml
+++ b/src/helm/impress/templates/ingress.yaml
@@ -28,7 +28,7 @@ spec:
   {{- end }}
   {{- if .Values.ingress.tls.enabled }}
   tls:
-    {{- if .Values.ingress.host }}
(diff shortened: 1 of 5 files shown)


rouja:
Hi,

I'm not sure to understand. If I set useDefaultSecretName to False, I no longer have a secretName for the tls secret. Is it what you want ?

dominikkaminski:
@rouja Yes, that was what I wanted - to provide my own tls secret name...

But after thinking about your comment - we would lile to provide an other name for secret, and so I updated my changes.

Could you please check it again?

Tested it already in our setup.

rouja (approved):
LGTM

AntoLC:
Merged from this pull request: https://github.com/numerique-gouv/impress/pull/519

PR #479, opened 2024-12-05
🎨Unlossy AI translation

AntoLC:
## Purpose

To translate with the AI we were using lossy functions (`blocksToMarkdownLossy` / `tryParseMarkdownToBlocks`), so we were loosing indentations, colors, background, lot of formatting. Some errors were popping up as well about tables loosing ids after the translation.

## Proposal

- We are now keeping the Blocknote json structure. 
- We associate an id to every node with the type text ([see](https://github.com/numerique-gouv/impress/pull/479/files#diff-e0d57a4975c27ca89a6756f94ae486adb23e7a844fa553ce45a6c1d619c961f0R24-R26))
- We extract the text with the id in a json
- We send only this json to the AI:
```json
{
    "tid-102": "Hello",
    "tid-103": "world",
    "tid-104": "Nice",
}
```
- The AI send back the same json with the translation:
```json
{
    "tid-102": "Bonjour",
    "tid-103": "le monde",
    "tid-104": "Super",
}
```
- We replace in the json Blocknote the text with the ID associated
- We replace in the editor the previous blocks with the new ones, block by block in checking if the block is still existing in the editor.

## Demo

[scrnli_ulUKPeGj2U7Hbw.webm](https://github.com/user-attachments/assets/b5fbb254-a227-4358-b764-378e70660821)

### DIFF
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -426,11 +426,12 @@ class AITranslateSerializer(serializers.Serializer):
     language = serializers.ChoiceField(
         choices=tuple(enums.ALL_LANGUAGES.items()), required=True
     )
-    text = serializers.CharField(required=True)
+    text = serializers.JSONField(required=True)
 
     def validate_text(self, value):
         """Ensure the text field is not empty."""
 
-        if len(value.strip()) == 0:
-            raise serializers.ValidationError("Text field cannot be empty.")
+        if not isinstance(value, dict):
+            raise serializers.ValidationError("Text field must be a json object.")
+
         return value
--- a/src/frontend/apps/impress/src/features/docs/doc-editor/components/AIButton.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-editor/components/AIButton.tsx
@@ -1,3 +1,4 @@
+import { Block } from '@blocknote/core';
 import {
   ComponentProps,
   useBlockNoteEditor,
@@ -21,6 +22,12 @@ import {
   useDocAITransform,
   useDocAITranslate,
 } from '../api/';
+import {
+  Node,
+  addIdToTextNodes,
+  extractTextWithId,
+  updateTextsWithId,
+} from '../utilsAI';
 
 type LanguageTranslate = {
   value: string;
@@ -212,14 +219,23 @@ const AIMenuItemTransform = ({
   icon,
 }: PropsWithChildren<AIMenuItemTransform>) => {
   const { mutateAsync: requestAI, isPending } = useDocAITransform();
+  const editor = useBlockNoteEditor();
 
-  const requestAIAction = async (markdown: string) => {
+  const requestAIAction = async (selectedBlocks: Block[]) => {
+    const text = await editor.blocksToMarkdownLossy(selectedBlocks);
+
     const responseAI = await requestAI({
-      text: markdown,
+      text,
       action,
       docId,
     });
-    return responseAI.answer;
+
+    if (!responseAI || !responseAI.answer) {
+      throw new Error('No response from AI');
+    }
+
+    const markdown = await editor.tryParseMarkdownToBlocks(responseAI.answer);
+    editor.replaceBlocks(selectedBlocks, markdown);
   };
 
   return (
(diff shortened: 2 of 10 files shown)


src/frontend/apps/impress/src/features/docs/doc-editor/utilsAI.ts
```diff
@@ -0,0 +1,145 @@
+import { Block as CoreBlock } from '@blocknote/core';
+
+type Block = Omit<CoreBlock, 'type'> & {
```
YousefED, line 3: why do we need this? I think you could use `Block` from BlockNote directly?

src/frontend/apps/impress/src/features/docs/doc-editor/components/AIButton.tsx
```diff
@@ -212,14 +219,23 @@ const AIMenuItemTransform = ({
...
       action,
       docId,
     });
-    return responseAI.answer;
+
+    if (!responseAI || !responseAI.answer) {
+      throw new Error('No response from AI');
+    }
+
+    const markdown = await editor.tryParseMarkdownToBlocks(responseAI.answer);
```
YousefED, line 237: Of course, non-translate AI actions like these will still be "lossy". I'll think of a blocknote-based solution for this (as part of BlockNote AI features)

YousefED (commented):
Nice solution! When we discussed on Matrix I missed that we were purely talking about translations (not other AI operations). So, basically you extract all text strings from the document and translate those via an LLM. That's quite nice and I think your approach should work!

src/frontend/apps/impress/src/features/docs/doc-editor/components/AIButton.tsx
AntoLC: Yes, this part is still lossy, I tried but I didn't have good AI results when it is with actions, the AI has difficulty to do the request when the text is too split:
https://github.com/numerique-gouv/impress/blob/f45e724b1797693841b9304c4d7dcc6939309bad/src/frontend/apps/impress/src/features/docs/doc-editor/api/useDocAITransform.tsx#L5-L9

I think it is less likely that users will select tables or complicated structures for these actions (to monitor).

src/frontend/apps/impress/src/features/docs/doc-editor/components/AIButton.tsx
```diff
@@ -280,31 +310,17 @@ const AIMenuItem = ({
   const editor = useBlockNoteEditor();
   const handleAIError = useHandleAIError();
 
-  const handleAIAction = async () => {
+  const handleAIAction = () => {
```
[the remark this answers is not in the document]
AntoLC: I removed all the `await` so `async` was not needed anymore.

src/frontend/apps/impress/src/features/docs/doc-editor/utilsAI.ts
AntoLC: I get some overlapping warning because of the unions, changing the type to string help to pass over. 
Not sure it is the best though.

image

src/frontend/apps/impress/src/features/docs/doc-editor/components/AIButton.tsx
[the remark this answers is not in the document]
YousefED: Got it, was wondering as there's a `then` below there's not much of a change.

src/frontend/apps/impress/src/features/docs/doc-editor/utilsAI.ts
YousefED: got it. It's possible for sure but would require a bit more work on the typings. If it's a priority we can pair on this

src/backend/core/api/serializers.py
```diff
@@ -426,11 +426,12 @@ class AITranslateSerializer(serializers.Serializer):
     language = serializers.ChoiceField(
         choices=tuple(enums.ALL_LANGUAGES.items()), required=True
     )
-    text = serializers.CharField(required=True)
+    text = serializers.JSONField(required=True)
```
arnaud-robin, line 429: Nit: wouldn't call this text in this context

arnaud-robin:
Hello Antho, nice solution ! However I think there might be issues when the order of the words differ in the two languages. 

For example I tried translating "Je **te** donne mon livre", which should translate to "I give **you** my book".
The payload is 
`{
    "tid-52": "Je ",
    "tid-53": "te",
    "tid-54": " donne mon livre."
}`
However the answer from gpt-4o-mini is:
`{
    "tid-52": "I ",
    "tid-53": "give ",
    "tid-54": "you my book."
}`
Here the bold word has been modified to give instead of you.

However, If i ask llama3.2 the answer becomes:
`{ 
"tid-52": "I ",
 "tid-53": "to", 
"tid-54": " give me my book." 
}
`
where the sentence doesn't mean anything anymore.

YousefED:
@arnaud-robin you requested my review here, but I think you're right about:

> there might be issues when the order of the words differ in the two languages

Good spot, this seems like a potential blocker and I missed that one earlier. I'm doing an experiment to keep markdown formatting across LLM responses, which might be a viable alternative and also makes it possible for operations other than "translate" (i.e.: rewrite, simplify, etc.). It's very much wip atm though

AntoLC:
Yes, I can see it is not the best so, even with gpt-4o-mini...

I am still wondering if markdown is the good way to solve this problem.
It should be easier to keep all the formatting with HTML, no ?

Curious to see how you do it with markdown @YousefED

PR #484, opened 2024-12-08
🐛(frontend) fix initial content with collaboration

AntoLC:
## Purpose

The initial document creation could create bug with collaboration. As soon as a char was enter and saved, this bug stopped to appear (basically after 1 mn).

The initial creation involving collaboration seems more complicated that what we were doing:
https://github.com/TypeCellOS/BlockNote/blob/71792c2fbdc99785cd8ce5b540d1a468e6c66131/packages/core/src/editor/BlockNoteTipTapEditor.ts#L73-L131

## Proposal

We will let Blocknote managing the initial creation, and we will update the Blocknote initial creation with our initial creation. 
Collaborating issue are gone, the document sync correctly between collaborators.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to
 
 ## Fixed
 
+- 🐛(frontend) fix initial content with collaboration #484
 - 🐛(frontend) Fix hidden menu on Firefox #468
 - 🐛(backend) fix sanitize problem IA #490
 
--- a/src/frontend/apps/impress/src/features/docs/doc-editor/components/BlockNoteEditor.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-editor/components/BlockNoteEditor.tsx
@@ -128,6 +128,23 @@ export const BlockNoteEditor = ({ doc, provider }: BlockNoteEditorProps) => {
   );
   useHeadings(editor);
 
+  /**
+   * With the collaboration it gets complicated to create the initial block
+   * better to let Blocknote manage, then we update the block with the content.
+   */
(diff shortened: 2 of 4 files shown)


YousefED (commented):
This should work, though I think it's a bit cleaner to do this on the server, so there's 1 place responsible for creating a new document (both the metadata and the document data), and `initialDoc` would always be set. 

Having said that, I do think this approach should work, there might be a small bug with two headings appearing if 2 users open a new document at the same time, but this is very unlikely

AntoLC:
> This should work, though I think it's a bit cleaner to do this on the server, so there's 1 place responsible for creating a new document (both the metadata and the document data), and `initialDoc` would always be set.
> 
> Having said that, I do think this approach should work, there might be a small bug with two headings appearing if 2 users open a new document at the same time, but this is very unlikely

Yes but the server is in python, and we re a SPA, so we cannot use `@blocknote/server-util` for the moment.

PanchoutNathan (approved):

PanchoutNathan:
@YousefED Hi youseff, Do you think it is possible to add an "editor is ready" event to avoid timeouts and to perform actions on the editor when we are sure that it is possible?

YousefED:
We'll make it so that you can immediately call updateBlock (without a
timeout or event). I have this on a dev branch already, I'll let you know
once it's merged (might be a few weeks, so until that time I think a
timeout is ok)

AntoLC:
> I have this on a dev branch already, I'll let you know once it's merged

Nice.

We will have the possibility soon to create it from the server as well. 
I will create an issue to keep it in mind.

PR #488, opened 2024-12-10
Mardown converter endpoint

lebaudantoine:
Allow backend service to convert markdown content in yjs one using a nodejs microservice bootstrapped by @AntoLC for collaboration purpose.

Blocked by #472 

Please check my commits;


AntoLC (approved):
It seems good, I didn't try it though.
We will split `server.test.ts` and `server.ts` to make it more readable and maintainable, but let's do it in another PR.

PR #490, opened 2024-12-11
🐛(backend) fix sanitize problem IA

AntoLC:
## Purpose

Albert send us back a malformed IA json, the sanitize function was not able to handle it correctly. 

## Proposal 

Better to remove the sanitize function for now, until we find a way to get the IA to send us a correct json. 
Downside, it will probably make the IA answer flakier.


src/backend/core/services/ai_services.py
```diff
@@ -66,11 +65,7 @@ def call_ai_api(self, system_content, text):
             ],
         )
 
-        content = response.choices[0].message.content
-        sanitized_content = re.sub(r"(?<!\\)\n", "\\\\n", content)
-        sanitized_content = re.sub(r"(?<!\\)\t", "\\\\t", sanitized_content)
-
-        json_response = json.loads(sanitized_content)
+        json_response = json.loads(response.choices[0].message.content)
```
lebaudantoine, line 68: Your reasoning is solid, and if the fix is functional and solves the problem effectively, it seems reasonable to move forward without additional complexity. 

However, we could enhance the implementation further with a try except, ignore if the sanitization fails and fallback on the `json.loads`

CHANGELOG.md
```diff
@@ -17,11 +17,15 @@ and this project adheres to
...
 
 - 🔨(frontend) encapsulated title to its own component #474
-- 🐛(frontend) Fix hidden menu on Firefox #468
 - ⚡️(backend) optimize number of queries on document list view #411
 - ♻️(frontend) stop to use provider with version #480
 - 🚚(collaboration) change the websocket key name #480
 
+## Fixed
+
+- 🐛(frontend) Fix hidden menu on Firefox #468
```
lebaudantoine, line 26: this line is wrongly committed

src/backend/core/services/ai_services.py
AntoLC: Yes better to do that.

CHANGELOG.md
AntoLC: I think its ok, I don't want to make a commit for that.

src/backend/core/tests/test_services_ai_services.py
```diff
@@ -85,20 +85,3 @@ def test_api_ai__success(mock_create):
...
-    """The AI response should be sanitized"""
-
-    answer = '{"answer": "Salut\\n \tle \nmonde"}'
-    mock_create.return_value = MagicMock(
-        choices=[MagicMock(message=MagicMock(content=answer))]
-    )
-
-    response = AIService().transform("hello", "prompt")
-
-    assert response == {"answer": "Salut\n \tle \nmonde"}
```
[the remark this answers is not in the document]
AntoLC: I added a new test asserting this behavior:
https://github.com/numerique-gouv/impress/pull/490/commits/31dce82a3813ae046dafa6e61da58ca5e79e51bb#diff-a57da39da9ad2fc788d1e3b887ea591ca8916b273c5749c40538d62acd9751f7

arnaud-robin (approved):

PR #503, opened 2024-12-13
Fix server-to-server doc creation

lebaudantoine:
Misc fixes

### DIFF
--- a/src/helm/env.d/dev/values.impress.yaml.gotmpl
+++ b/src/helm/env.d/dev/values.impress.yaml.gotmpl
@@ -51,8 +51,8 @@ backend:
     AWS_S3_SECRET_ACCESS_KEY: password
     AWS_STORAGE_BUCKET_NAME: impress-media-storage
     STORAGES_STATICFILES_BACKEND: django.contrib.staticfiles.storage.StaticFilesStorage
-    CONVERSION_API_URL: http://impress-y-provider:443/api/convert-markdown/
-    CONVERSION_API_KEY: my-secret
+    Y_PROVIDER_API_BASE_URL: http://impress-y-provider:443/api/
+    Y_PROVIDER_API_KEY: my-secret
 
   migrate:
     command:
--- a/src/helm/env.d/dev/values.impress.yaml.gotmpl
+++ b/src/helm/env.d/dev/values.impress.yaml.gotmpl
@@ -105,6 +105,7 @@ yProvider:
     COLLABORATION_LOGGING: true
     COLLABORATION_SERVER_ORIGIN: https://impress.127.0.0.1.nip.io
     COLLABORATION_SERVER_SECRET: my-secret
+    Y_PROVIDER_API_KEY: my-secret
 
 ingress:
   enabled: true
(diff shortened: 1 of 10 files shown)


qbey (approved):

src/helm/env.d/dev/values.impress.yaml.gotmpl
```diff
@@ -50,6 +51,8 @@ backend:
     AWS_S3_SECRET_ACCESS_KEY: password
     AWS_STORAGE_BUCKET_NAME: impress-media-storage
     STORAGES_STATICFILES_BACKEND: django.contrib.staticfiles.storage.StaticFilesStorage
+    CONVERSION_API_URL: http://impress-y-provider:443/api/convert-markdown/
```
AntoLC, line 54: There is no flexibility with this approach, we will have to create a env var for every endpoints if we do that.
Like propose in PM, `YJS_SERVER_API` will fit more what it is.

AntoLC (commented):
Please, let me in review for this kind of PR, Quentin is expert Django, and working on a other repository, if you don't know how is working the collaboration server, you cannot review correctly this kind of PR.

src/helm/env.d/dev/values.impress.yaml.gotmpl
lebaudantoine: Yes discussed IRL. I agree with you. However, at the time we discussed my endpoint, I add no idea if you would add more endpoint. Also, the express server isn't really configured to use a specific base URL … 

I can refactor this part, but I feel I am solving future problems, which wasn't the top priority while working on my urgent topics. This won't be a huge refactoring, I'll try to do it.

src/helm/env.d/dev/values.impress.yaml.gotmpl
```diff
@@ -50,6 +51,8 @@ backend:
...
+    CONVERSION_API_KEY: my-secret
```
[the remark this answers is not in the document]
lebaudantoine: Discussed IRL with Quentin. 

I did not want to couple the collaboration server secret and the other js endpoints microservice your app will require. Duplication isn't inherently bad; prematurely factoring out unrelated logic can lead to unnecessary coupling and reduced clarity.

src/helm/env.d/dev/values.impress.yaml.gotmpl
lebaudantoine: the microservice is name `y-provider`, you suggest `yjs_server`, actually all the code of this microservice is run by a `collaboration server` … IDK what really should be the naming of this var. However, I will follow your suggestions, but I don't want my call the microservices through the ingress

lebaudantoine: updated !

src/helm/env.d/dev/values.impress.yaml.gotmpl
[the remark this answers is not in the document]
lebaudantoine: I've updated the `CONVERSION_API_KEY` TO `YPROVIDER_API_KEY`. I think it's better to rename the existing secret to this naming, and remove the `CONVERSION_API_KEY`. Might be done in another PR.

AntoLC (approved):
Thank you, I find it is better like that.
Don't forget here to add the secret:
https://github.com/numerique-gouv/impress/blob/5a3b67b69aab5fd3f46e0faa2bb91d11bf1884cb/src/helm/env.d/dev/values.impress.yaml.gotmpl#L107

PR #505, opened 2024-12-16
Experimentation with "internal" llm

lebaudantoine:
Please check my first comit. It's draft

### DIFF
--- a/src/backend/core/services/ai_services.py
+++ b/src/backend/core/services/ai_services.py
@@ -35,10 +35,29 @@ AI_ACTIONS = {
     ),
 }
 
+
 AI_TRANSLATE = (
-    "Translate the markdown text to {language:s}, preserving markdown formatting. "
-    'Return JSON: {{"answer": "your translated markdown text in {language:s}"}}. '
-    "Do not provide any other information."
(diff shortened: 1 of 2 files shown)


src/backend/core/services/ai_services.py
AntoLC, line 1: I think it is a good idea to do directly in markdown.

This prompts as well should be improved, they use the same methods:
https://github.com/numerique-gouv/impress/blob/8b60bc57e6c980dfc5eed8932a3cf64244b2079b/src/backend/core/services/ai_services.py#L13-L36

AntoLC (commented):
We are as well finalizing a PR here, where we actually stop to use totally the markdown with the translation actions: https://github.com/numerique-gouv/impress/pull/479

src/backend/core/services/ai_services.py
lebaudantoine: Yes ofc, I mention it in my commit "Please note I haven't updated yet the other prompts, let's discuss it before."

AntoLC:
I got inspired by it, close in favor of: https://github.com/suitenumerique/docs/pull/662

PR #511, opened 2024-12-17
⚡️e2e flakiness

AntoLC:
## Purpose

e2e test start to be very flaky, we fix a bug that create flakiness, plus improve playwright installation.

## Proposal

-  🐛(frontend) fix rerendering when doc is saving
- ⚡️(ci) playwright install
- ⚡️(e2e) set maxFailures with CI

### DIFF
--- a/src/frontend/apps/e2e/__tests__/app-impress/doc-header.spec.ts
+++ b/src/frontend/apps/e2e/__tests__/app-impress/doc-header.spec.ts
@@ -120,13 +120,14 @@ test.describe('Doc Header', () => {
 
     await editor.locator('h1').fill('');
 
+    // eslint-disable-next-line playwright/no-wait-for-timeout
+    await page.waitForTimeout(500);
+
     await docHeader
       .getByRole('heading', { name: 'Top World', level: 2 })
       .fill(' ');
 
-    await page.getByText('Created at').click({
-      delay: 200,
-    });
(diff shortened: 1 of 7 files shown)


src/frontend/apps/e2e/__tests__/app-impress/doc-header.spec.ts
```diff
@@ -125,7 +125,7 @@ test.describe('Doc Header', () => {
       .fill(' ');
 
     await page.getByText('Created at').click({
-      delay: 200,
+      delay: 1000,
     });
```
PanchoutNathan, line 129: remove this, by default we have 5.000 ms

PanchoutNathan (approved):

src/frontend/apps/e2e/__tests__/app-impress/doc-header.spec.ts
AntoLC: I removed this delay.
I can clearly reproduce the flakiness, I didn't find better than adding that:
https://github.com/numerique-gouv/impress/pull/511/commits/19224026f02e622420c93213b253c28fcf95607d#diff-8c5f9b668095c002522e065079a61f46ea012038e2eec0b6d88e8b2411d55e6dR124

Lot of states are going on very quickly, it needs a delay to stabilize.

Issue #435, opened 2024-11-20
Subpages


virgile-dev:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
Users want a way to build collections of pages in order to organize information as wiki or in folder.

**Describe the solution you'd like**
Backend 
- Allow users to add subpages. 
- It will possible to add subpages to subpages.
- We'll need to deal with permissions. By default a subpage has the same permissions as its parent page. 
- Users  should be able to override permissions that were inherited by default.
- A user can't have less permission on a child page than on a parent page
Frontend
- Subpages will display in the left side bar
  - Users will be able to move around subpages
  - A toggle allow to hide or display subpages in the sidebar
  - On hover action will be display to add a subpage and a ... menu 

**Mockups**
![Image](https://github.com/user-attachments/assets/26fe056d-92dc-4a50-a7c9-8c7d7c24dd70)

sampaccoud:
As of today, everything that is specified in this issue is included in PR https://github.com/numerique-gouv/impress/pull/516.

PR #516, opened 2024-12-18
Allow organizing documents in a tree structure

sampaccoud:
## Purpose

We want to allow nesting documents in a tree structure.

## Proposal

This pull request introduces a hierarchical organization for documents by implementing a tree structure. It integrates the [django-treebeard](https://django-treebeard.readthedocs.io/en/latest/) library to manage the tree hierarchy, allowing documents to be nested within parent documents. This enhancement will turn Docs into a knowledge management platform.

### Key Changes:

- [x] Add the django-treebeard library to dependencies to support tree structures for documents
- [x] Update document models to include parent-child relationships. 
- [x] Modify and optimize the document retrieval logic to take into account the tree structure, ensuring that only the first visible parent document for a user is shown in list views.
- [x] Add endpoint to list and create children below a document 
- [x] Add soft delete to allow trashbin restore (important when deleting complete subtrees)
- [x] Allow moving a document (and the whole subtree below it) to another place in the tree structure 
- [x] Improve admin interface and activate treebeard's nested views
 - [x] Modify the test suites to take into account the tree structure of documents

:warning: Note: Existing documents are transformed to root nodes in the new tree structure.

Resolves https://github.com/suitenumerique/docs/issues/435

Side improvements were also done along the way:
- [x] Make document/template creation atomic
- [x] Generalize use of SerializerPerActionMixin in viewsets
- [x] Bump maximum page size to 200 for new UI
- [x] Add user roles as field in the document API representation
- [x] Remove "content" field for document list views serializer and add "excerpt" field instead


src/backend/core/tests/documents/test_api_documents_list.py
```diff
@@ -59,7 +59,6 @@ def test_api_documents_list_format():
     assert len(results) == 1
     assert results[0] == {
         "id": str(document.id),
-        "content": document.content,
```
AntoLC, line 65: `excert` is not a value that we should get in the document_list ?

src/backend/core/models.py
```diff
@@ -326,6 +326,7 @@ class Document(MP_Node, BaseModel):
     """Pad document carrying the content."""
 
     title = models.CharField(_("title"), max_length=255, null=True, blank=True)
+    excerpt = models.TextField(_("excerpt"), max_length=300, null=True, blank=True)
```
AntoLC, line 329: Ok, so it is the front setting this var during saving as well I guess.

src/backend/core/api/serializers.py
```diff
@@ -165,7 +164,6 @@ class Meta:
         ]
         read_only_fields = [
             "id",
-            "abilities",
```
AntoLC, line 168: It will break the frontend if you remove it. We use the abilities in the data grid, to know if the current user can delete or not a document.
We use it as well to compute the current role of the user on the document (not sure we still do it with the new design) though.

src/backend/demo/management/commands/create_demo.py
```diff
@@ -135,9 +135,14 @@ def create_demo(stdout):
...
 
     with Timeit(stdout, "Creating documents"):
-        for _ in range(defaults.NB_OBJECTS["docs"]):
+        for i in range(defaults.NB_OBJECTS["docs"]):
+            # pylint: disable=protected-access
+            key = models.Document._int2str(i)  # noqa: SLF001
+            padding = models.Document.alphabet[0] * (models.Document.steplen - len(key))
             queue.push(
                 models.Document(
+                    depth=1,
```
AntoLC, line 144: Is it possible to add some children docs, with different depth as well ?

AntoLC (commented):
That is nice!!

It is a bit hard to review from a frontend side perspective without too much context, but it looks what we need to do our tree structure.
For me, if you put back `abilities`, we could merge it because except that, it does not seems to impact negatively the frontend; then we will iterate when we will start to integrate the tree on the frontend.

src/backend/core/api/serializers.py
```diff
@@ -155,21 +155,27 @@ class Meta:
             "content",
             "created_at",
             "creator",
+            "depth",
             "is_favorite",
             "link_role",
             "link_reach",
             "nb_accesses",
+            "numchild",
+            "path",
```
AntoLC, line 163: What is the purpose of `path` ?

src/backend/core/tests/documents/test_api_documents_children_list.py
```diff
@@ -0,0 +1,496 @@
...
+    assert response.json() == {
+        "detail": "You do not have permission to perform this action."
+    }
+
+
+def test_api_documents_children_list_authenticated_related_direct():
+    """
+    Authenticated users should be allowed to retrieve the children of a document
+    to which they are directly related whatever the role.
+    """
```
AntoLC, line 275: I am wondering:
- I am a member of a child document
- I am not a member of a parent document
Do I have the right to see the parent documents ?

If not, will this child document appear in the endpoint `GET /api/v1.0/documents` ?

virgile-dev:
Linked : https://github.com/orgs/numerique-gouv/projects/13/views/11?filterQuery=-status%3ADone++sub&pane=issue&itemId=86564324&issue=numerique-gouv%7Cimpress%7C435

src/backend/core/tests/documents/test_api_documents_list.py
sampaccoud: You're right. I brought it back.

src/backend/core/models.py
sampaccoud: Yes. We discussed with @PanchoutNathan and @rl-83. Two choice:
1. The frontend sets it automatically with the start of the document content: not good as the document could start by bullet points or...
2. The frontend expects users to fill in a description. If nothing is set, the result is displayed with just the title. This is what Github does and my favorite option.
3. The backend runs an LLM in the background and fills in description by summarizing the document content.

2 and 3 could be combined...

src/backend/core/api/serializers.py
sampaccoud: I have brought back the "abilities" field and introduced a new field "user_roles" which lists all roles applied to the user on the document. I leave it up to the frontend to decide what to display from this list. WDYT?

src/backend/core/api/serializers.py
sampaccoud: Path is how the tree is built. There is no foreign key between a child and its parent.
Root documents have a path that looks like `0000001`, `0000002`, `0000003`, etc.
Children of the first root document have a path that looks like: `00000010000001`, `00000010000002`, `00000010000003`, etc.
etc.

It makes it very efficient to get a document's ancestors or descendants by filtering this way:
```python
descendants = Document.objects.filter(path__startswith=parent.path)
children = Document.objects.filter(path__startswith=parent.path, depth=parent.depth+1)
parent = Document.objects.get(path=child.path[-7:])
...
```

src/backend/core/tests/documents/test_api_documents_children_list.py
sampaccoud: Nope as tested here: https://github.com/numerique-gouv/impress/pull/516/files#diff-f0e071eb4146cdb8b7cec94353ffe92734df4836f6d6df0d48e21e659de19d17R102

No. Only the highest accessible ancestor in a subtree appears in list views...
If you have access to `document1` you will automatically have access to its children and to the children of its children etc.
More precisely, if you have `administrator` access to `document1` you will, at the minimum, get `administrator` access to all its descendants. Extra rights can be added on a document below `document1` and these extra rights will then apply to all the descendants of this document...

AntoLC (commented):
Except the collation problem, all seems ok.

It exists the extension `pg_stat_statements`, it could be maybe nice to add it, it would help us to track slow queries. Can be in another PR though.

src/backend/core/api/viewsets.py
```diff
@@ -338,66 +412,148 @@ class DocumentViewSet(
...
+        # deleted as well.
+        ancestors_subquery = queryset.filter(
+            path=Left(db.OuterRef("path"), Length("path"))
+        )
 
-        page = self.paginate_queryset(queryset)
-        if page is not None:
-            serializer = self.get_serializer(page, many=True)
-            return self.get_paginated_response(serializer.data)
+        # Annotate with the oldest `deleted_at`` date among ancestors using the `Min` function`
```
AntoLC, line 510:
```suggestion
        # Annotate with the oldest `deleted_at` date among ancestors using the `Min` function`
```

src/backend/core/tests/documents/test_api_documents_retrieve.py
```diff
@@ -43,12 +51,90 @@ def test_api_documents_retrieve_anonymous_public():
...
+        "excerpt": document.excerpt,
         "is_favorite": False,
         "link_reach": "public",
         "link_role": document.link_role,
         "nb_accesses": 0,
+        "numchild": 0,
+        "path": document.path,
+        "title": document.title,
+        "updated_at": document.updated_at.isoformat().replace("+00:00", "Z"),
+        "user_roles": [],
```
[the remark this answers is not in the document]
sampaccoud: But you have this information on link_reach/link_role so including it here would be less information for you...
It is a list because you could have been given "editor" access on the root node, "administrator" access on its child and be the owner of the leaf document... In this case the `user_roles` field on the leaf would equal ["editor", "administrator", "owner"]. Which for you is the same as saying the the user has owner rights but that would be more work for me and less information for you...

src/backend/core/migrations/0014_set_path_on_existing_documents.py
```diff
@@ -0,0 +1,52 @@
...
+    dependencies = [
+        ('core', '0013_add_tree_structure_to_documents'),
+    ]
+
+    operations = [
+        migrations.RunPython(set_path_on_existing_documents, reverse_code=migrations.RunPython.noop),
+        migrations.AlterField(
+            model_name='document',
+            name='path',
+            field=models.CharField(max_length=255, unique=True),
```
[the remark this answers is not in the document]
AntoLC: You need to have a certain amount of document to see the error, the algo has to reach the lowercase values.
If you add this test, it will showcase the problem:

```py
def test_api_documents_create_authenticated_success_check_path():
    """
    Check that the path of the document is correctly set when creating a document.
    """

    client = APIClient()
    client.force_login(user)

    # Create 50 docs
    for i in range(50):
        response = client.post(
            "/api/v1.0/documents/",
            {
                "title": f"my document {i}",
            },
            format="json",
        )

        assert response.json() != ['Document with this Path already exists.']
        assert response.status_code == 201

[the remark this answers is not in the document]
Comment (qbey) [reply src/backend/core/migrations/0014_set_path_on_existing_documents.py:50]: Yep, had the same behavior on https://github.com/numerique-gouv/people/pull/560

Warning, the alphabet must be fixed also, because uppercase are "lower" than lowercase when sorting.

``` python
ALPHABET = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
```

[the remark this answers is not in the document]
qbey: @sampaccoud I suggest to add a test like 

``` python
def test_models_documents_tree_alphabet():
    """Test the creation of teams with treebeard methods."""
    models.Document.load_bulk([
        {
            "data": {
                "title": f"document-{i}",
            }
        }
        for i in range(len(models.Document.alphabet) * 2)
    ])
```

Which will assert the alphabet is good + the collation is properly configured.

src/backend/core/tests/documents/test_api_document_versions.py
```diff
@@ -185,6 +185,83 @@ def test_api_document_versions_list_authenticated_related_pagination(
...
+            user=user,
+            role=random.choice(models.RoleChoices.choices)[0],
+        )
+    elif via == TEAM:
+        mock_user_teams.return_value = ["lasuite", "unknown"]
+        models.DocumentAccess.objects.create(
+            document=grand_parent,
+            team="lasuite",
+            role=random.choice(models.RoleChoices.choices)[0],
+        )
```
qbey, line 221: I'm not a big fan of "logic" in tests, but while this reduces the number of lines a lot, I guess it's ok ^^

src/backend/core/tests/documents/test_api_document_versions.py
```diff
@@ -185,6 +185,83 @@ def test_api_document_versions_list_authenticated_related_pagination(
...
+    document = factories.DocumentFactory(parent=parent)
+    for i in range(3):
+        document.content = f"before {i:d}"
+        document.save()
+
+    if via == USER:
+        models.DocumentAccess.objects.create(
+            document=grand_parent,
+            user=user,
+            role=random.choice(models.RoleChoices.choices)[0],
```
qbey, line 213: Nit

```suggestion
            role=random.choice(models.RoleChoices.values),
```

src/backend/core/tests/documents/test_api_document_versions.py
```diff
@@ -185,6 +185,83 @@ def test_api_document_versions_list_authenticated_related_pagination(
...
+        )
+    elif via == TEAM:
+        mock_user_teams.return_value = ["lasuite", "unknown"]
+        models.DocumentAccess.objects.create(
+            document=grand_parent,
+            team="lasuite",
+            role=random.choice(models.RoleChoices.choices)[0],
```
qbey, line 220: Nit

```suggestion
            role=random.choice(models.RoleChoices.values),
```

qbey (commented):
I guess the only thing to fix is the collation and the alphabet, otherwise this looks good to me :) I took few ideas of this PR for mine ^^

src/backend/core/models.py
```diff
@@ -358,6 +359,10 @@ class Document(BaseModel):
 
     _content = None
 
+    # Tree structure
+    steplen = 7  # nb siblings max: 78,364,164,096 / max depth: 255/7=36
+    node_order_by = None  # Manual ordering
```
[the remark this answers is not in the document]
sampaccoud: No reason. Changed to [] and all tests pass...

src/backend/core/tests/documents/test_api_document_versions.py
sampaccoud: Good point. Let's do it!

src/backend/core/tests/documents/test_api_document_versions.py
sampaccoud: As discussed IRL. In this case, it's not much code in the if/else and it avoids duplicating all our API tests to test for users and for teams. We could think of something more elegant but not in this PR :sweat_smile:

src/backend/core/models.py
```diff
@@ -774,11 +951,27 @@ class Meta:
     def __str__(self):
         return self.title
 
+    def get_roles(self, user):
+        """Return the roles a user has on a resource."""
+        if not user.is_authenticated:
+            return []
+
+        try:
+            roles = self.user_roles or []
```
qbey, line 960: Is the `or []` necessary? because if the attribute is present it should be `[]` already, no?

qbey (approved):
:exploding_head: seems very good to me, nice work!

src/backend/core/models.py
```diff
@@ -541,48 +561,124 @@ def delete_version(self, version_id):
...
+    def nb_accesses(self):
+        """Calculate the number of accesses."""
+        cache_key = self.get_nb_accesses_cache_key()
+        nb_accesses = cache.get(cache_key)
+
+        if nb_accesses is None:
+            nb_accesses = DocumentAccess.objects.filter(
+                document__path=Left(models.Value(self.path), Length("document__path")),
+            ).count()
+            cache.set(cache_key, nb_accesses, timeout=3600)
```
lunika, line 578: What about managing this timeout with a setting ? It's still faster to change it in production if needed than releasing a new version

sampaccoud: Defining timeout is already handled in Django's CACHES setting. I'll add a setting to configure its TIMEOUT and remove it from here.

src/backend/core/models.py
```diff
@@ -643,6 +739,77 @@ def send_invitation_email(self, email, role, sender, language=None):
...
+
+        # Mark all descendants as soft deleted
+        self.get_descendants().filter(ancestors_deleted_at__isnull=True).update(
+            ancestors_deleted_at=self.ancestors_deleted_at
+        )
+
+    @transaction.atomic
+    def restore(self):
+        """Cancelling a soft delete with checks."""
+        # This should not happen
```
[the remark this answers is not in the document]
sampaccoud: I still don't see how it could happen :smile:

src/backend/core/models.py
```diff
@@ -643,6 +739,77 @@ def send_invitation_email(self, email, role, sender, language=None):
 
         self.send_email(subject, [email], context, language)
 
+    @transaction.atomic
+    def soft_delete(self):
+        """
+        Soft delete the document, marking the deletion on descendants.
+        We still keep the .delete() method untouched for programmatic purposes.
```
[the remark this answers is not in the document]
sampaccoud: No. I still want to allow calling `delete` in the code while treating deletes coming from the API as soft deletes.

src/backend/demo/management/commands/create_demo.py
sampaccoud: we can do this later as it requires a refacto and I'm already doing too much in this PR!

src/backend/core/api/viewsets.py
```diff
@@ -301,100 +296,244 @@ def determine_metadata(self, request, view):
...
             queryset = queryset.filter(
-                db.Q(accesses__user=user)
-                | db.Q(accesses__team__in=user.teams)
+                db.Q(id__in=access_documents_ids)
                 | (
-                    db.Q(link_traces__user=user)
+                    db.Q(id__in=traced_documents_ids)
                     & ~db.Q(link_reach=models.LinkReachChoices.RESTRICTED)
                 )
             )
```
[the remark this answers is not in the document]
sampaccoud: the diff of your proposal is weird :thinking: 
This filter on RESTRICTED is necessary to exclude documents on which you have no specific access by that you may have visited when they were not restricted yet (so you have a LinkTrace on it) and were later restricted!

src/backend/core/api/viewsets.py
```diff
@@ -338,66 +412,148 @@ class DocumentViewSet(
...
-        Use ListDocumentSerializer for list actions, otherwise use DocumentSerializer.
+        Use ListDocumentSerializer for list actions; otherwise, use DocumentSerializer.
         """
-        if self.action == "list":
-            return serializers.ListDocumentSerializer
-        return self.serializer_class
+        return (
+            serializers.ListDocumentSerializer
+            if self.action == "list"
+            else self.serializer_class
```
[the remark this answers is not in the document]
sampaccoud: Done

src/backend/core/api/viewsets.py
```diff
@@ -301,100 +296,244 @@ def determine_metadata(self, request, view):
...
 
-    def list(self, request, *args, **kwargs):
-        """Restrict resources returned by the list endpoint"""
-        queryset = self.filter_queryset(self.get_queryset())
+    def get_queryset(self):
+        """Get queryset performing all annotation and filtering on the document tree structure."""
         user = self.request.user
+        queryset = super().get_queryset()
+
+        if not self.detail:
```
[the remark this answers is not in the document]
sampaccoud: Much better thanks!
Added some comments as well.

src/backend/core/models.py
sampaccoud: it was set to None and we want to be able to use it with "in" statements.

src/backend/core/api/viewsets.py
```diff
@@ -405,32 +544,72 @@ def retrieve(self, request, *args, **kwargs):
...
+        obj = models.Document.add_root(
+            creator=self.request.user,
+            **serializer.validated_data,
+        )
+        serializer.instance = obj
         models.DocumentAccess.objects.create(
             document=obj,
             user=self.request.user,
             role=models.RoleChoices.OWNER,
         )
```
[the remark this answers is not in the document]
sampaccoud: I think it's better to add it yes.

src/backend/core/api/viewsets.py
```diff
@@ -405,32 +544,72 @@ def retrieve(self, request, *args, **kwargs):
...
-                    user=self.request.user,
-                )
-            except ValidationError:
-                # The trace already exists, so we just pass without doing anything
-                pass
+        if (
+            user.is_authenticated
+            and not instance.link_traces.filter(user=user).exists()
+        ):
+            models.LinkTrace.objects.create(document=instance, user=request.user)
```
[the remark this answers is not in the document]
sampaccoud: I tried but the create generates 5 queries... so since I expect the user to access the document many time after the first visit, I should focus on making the subsequent reads as fast as possible which is what the `exists` query does!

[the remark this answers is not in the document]
sampaccoud: Added a comment to explain this.

src/backend/core/models.py
```diff
@@ -541,48 +561,124 @@ def delete_version(self, version_id):
...
 
-        is_owner_or_admin = bool(
-            roles.intersection({RoleChoices.OWNER, RoleChoices.ADMIN})
-        )
-        can_get = bool(roles)
-        can_update = is_owner_or_admin or RoleChoices.EDITOR in roles
+        if user.is_authenticated:
+            for lr in links_definitions:
+                if lr["link_reach"] == LinkReachChoices.AUTHENTICATED:
+                    roles.add(lr["link_role"])
```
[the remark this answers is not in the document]
sampaccoud: ok let's do that!

sampaccoud:
@AntoLC @qbey @lunika I'm done with your reviews. The PR should be ready to merge if Anto confirms that the impact he noted on frontend are not solved.

AntoLC (approved):
I just tried again, seems fine frontend side.
Let's go

PR #517, opened 2024-12-18
✨Collaboration long polling fallback

### LINKED ISSUE #437: Switch from Web Sockets to Long polling
## Bug Report

**Problematic behavior**
Websockets are not compatible with internal networks and security configuration.

**Expected behavior/code**
Replace websocket with Long Polling. 
Which means we need to make changes to hoccus poccus and Y.js

### LINKED ISSUE #456: Collaboration when websockets disable overides content
## Bug Report

**Problematic behavior**
A user that blocks websockets that is working on a collaborative documents will overide the other user content (or his content will get erased) when refreshing the page.

This is an issue because the user without websockets saves automatically in the Django data base.

**Expected behavior/code**
I shouldn't overide content when collaborating.

**Steps to Reproduce**
1. Disable web sockets
2. Open 2 tabs with the same doc (or share the link with someone with websockets)
3. Write content on the 3 first blocks on both tabs
4. Wait 30 sec and refresh the page
5. See that one version was kept over the other one. One was overidden.

**Environment**
- Impress version:
- Platform:

**Possible Solution**
*   Detect when a user doesn't have access to web sockets and display a banner "Your laptop configuration prevents live collaboration, if you continue editing the document you might override others' work and vice versa.
    *   The banner should be displayed when the doc is shared (specific invite + link sharing + edit permission)
*   We could disable auto-saving if the user is disconnected from websocket and add a manual saving button that syncs the edits and does a merge
*   Attributing a version to a user we could see if the user editing is the one who last saved
*   The hocus pocus server could be configured in a way where it centralizes the edits and prevent conflicts
*  Add long polling support when we detect there are no web sockets support

### LINKED ISSUE #486: When no websockets and multi-editors switch user rights to read-only
## Feature Request

Here is the related issue : https://github.com/numerique-gouv/impress/issues/456

When a user has websockets blocked AND there are multiple possible editors (the document is shared with other people or there an edit link that allow collaboration) they should be able to only read the document and an alert banner should display to explain why.

This is until we implement some kind of alternative solution like polling. (https://github.com/numerique-gouv/impress/issues/437)

CleanShot 2024-12-09 at 18 26 13@2x
(full mockup in the "In progress" area.

AntoLC:
## Purpose

Some users have their websockets blocked, so they cannot collaborate.
If they are connected with other collaborators at the same time, it will create constant conflict in the document.

## Proposal

We have succeeded to propose an experience almost as good as with websocket.

- We will use a http fallback when the websocket is not able to connect.
- We are still using the Hocus Pocus mechanism, so push and pull are trigger by the Hocus Pocus provider and server.
- By using the Hocus Pocus mechanism, we are still using `y-protocols/sync` making our request very light (a few bytes).
- We are using the SSE (https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events)
to pull data, to minimize the requests amount and keep as much as possible our documents sync between each others.

## Cases we solved:
- [x] connect even without websockets users altogether
- [x] keep rights (can edit / can view) by using the same mechanism as with the WS
- [x] keep the awareness (cursor), sync and doc update
- [x] keep our requests light
- [x] add a nginx auth cache system - query the backend 1 time every 30 seconds
- [x] test what I could

## Architecture

```mermaid
flowchart TD
    title1[WebSocket Success]-->Client1(Client)<--->|WebSocket Success|WS1(Websocket) --> Nginx1(Ngnix) <--> Auth1("Auth Sub Request (Django)") --->|With the good right|YServer1("Hocus Pocus Server")
  YServer1 --> WS1
  YServer1 <--> clients(Dispatch to clients)
  title2[WebSocket Fails - Push data]-->Client2(Client)---|WebSocket fails|HTTP2(HTTP) --> Nginx2(Ngnix) <--> Auth2("Auth Sub Request (Django)")--->|With the good right|Express2(Express) --> YServer2("Hocus Pocus Server") --> clients(Dispatch to clients)
  title3[WebSocket Fails - Pull data]-->Client3(Client)<--->|WebSocket fails|SSE(SSE) --> Nginx3(Ngnix) <--> Auth3("Auth Sub Request (Django)") --->|With the good right|Express3(Express) --> YServer3("Listen Hocus Pocus Server")
  YServer3("Listen Hocus Pocus Server") --> SSE
  YServer3("Listen Hocus Pocus Server") <--> clients(Data from clients)
```


virgile-dev:
Great job @AntoLC !

YousefED (commented):
@AntoLC Nice you got this working. If I see it correctly, you created a new endpoint over which you're always syncing the entire Y.Doc to and from the server. 

If I'm not mistaken, normally the Y.js sync protocol is more efficient than this and syncs the exact updates required. What's the reason you went for this approach (new endpoint, syncing entire doc) instead of the proxy approach? I think the proxy approach has some potential advantages:

- We can keep the same sync protocol, but just switch to a different transport (more efficient and awareness would still work)
- The HocusPocus side can stay the same, our "fix" would be isolated to a separate layer) (less code complexity and smaller chance of bugs or security issues)

I might be missing some advantages of your current approach, but my concern is mainly that it adds more "custom code" that's another surface we need to test, maintain and secure. The proxy approach would isolate / limit this more I think

AntoLC:
You can test this PR before it is merged on https://docs-ia.beta.numerique.gouv.fr/.
To deactivate the websocket add the query param `withoutWS=true`

Example public doc: https://docs-ia.beta.numerique.gouv.fr/docs/481a9933-3514-4aeb-9877-c21be1388877/?withoutWS=true

docker/files/etc/nginx/conf.d/default.conf
```diff
@@ -34,6 +41,10 @@ server {
     }
 
     location /collaboration-auth {
+        proxy_cache auth_cache;
+        proxy_cache_key "$http_authorization";
```
lunika, line 45: Maybe add something more specific to avoid sharing the same cache key later with an other location

```suggestion
        proxy_cache_key "$http_authorization$request_uri";
```

src/helm/impress/values.yaml
```diff
@@ -82,7 +82,11 @@ ingressCollaborationWS:
   ## @param ingressCollaborationWS.annotations.nginx.ingress.kubernetes.io/proxy-send-timeout
   ## @param ingressCollaborationWS.annotations.nginx.ingress.kubernetes.io/upstream-hash-by
   annotations:
+    nginx.ingress.kubernetes.io/auth-cache-key: "$http_authorization"
```
lunika, line 85: Same here ? `"$http_authorization$request_uri";`

docker/files/etc/nginx/conf.d/default.conf
AntoLC: Yes you'r totally right

src/frontend/apps/impress/src/features/docs/doc-management/libs/CollaborationProvider.ts
```diff
@@ -0,0 +1,333 @@
...
+      setTimeout(() => {
+        this.initCollaborationSSE();
+      }, 5000);
+    };
+  }
+
+  /**
+   * Sync the document with the server.
+   *
+   * In some rare cases, the document may be out of sync.
```
YousefED, line 285: Similar to above, could you also call the existing HP `forceSync` and let the Yjs sync protocol handle this? passing Yjs documents and updates around seems a little dangerous to me (at least it's difficult for me to verify if this is correct or not)

src/frontend/apps/impress/src/features/docs/doc-management/libs/CollaborationProvider.ts
```diff
@@ -0,0 +1,333 @@
...
+    return false;
+  }
+
+  /**
+   * Outgoing message event
+   *
+   * Sent to the server the message to
+   * be sent to the other users
+   */
+  public async onPollOutgoingMessage({ message }: onOutgoingMessageParameters) {
```
YousefED, line 188: I got confused at some parts when checking the code, I think it might be helpful to explain or improve the naming a little bit.

Technically this method is not "polling" anymore, right? When you're polling, you retrieve an update from somewhere, but here we're just sending a message, like a regular POST request.

---

Also, you're not using Long-Polling at all, because long-polling means (afaik) that you send a request to the server, which keeps the connection open until the server has something interesting to send to you. I don't think this is the case, as you're using SSE for events from server -> client.

Which is fine, but maybe better to avoid the term long-polling then to avoid confusion down the line

src/frontend/apps/impress/src/features/docs/doc-management/libs/CollaborationProvider.ts
```diff
@@ -0,0 +1,333 @@
...
+
+        this.treatedAwarenessMessage = awareness64;
+        const awareness = Buffer.from(awareness64, 'base64');
+
+        this.onMessage({
+          data: awareness,
+        } as MessageEvent);
+      }
+
+      if (updatedDoc64) {
```
YousefED, line 256: Why do we have our own logic to apply the message? Isn't it enough (and easier), to let the `onMessage` function handle this for all cases? Or is there a reason that doesn't work?

---

Zoomed out; I'm a little concerned by all the manual Yjs operations you have to do. I was hoping you could just use the existing sync-protocol (and code that handles that), but only over a different transport layer. My guess is you ran into issues with this and came up with some workarounds? That does make the code a little more difficult to review (especially when I don't have the context of why which workarounds are necessary).

PR #520, opened 2024-12-19
🐛(create_demo) fix failing create_demo

AntoLC:
## Purpose

When we do `create_demo` on a database with already dummy data, the script fails because of unicity constraints. 

## Proposal

This commit fixes this by checking if the data are already present before creating them.

### DIFF
--- a/src/backend/demo/management/commands/create_demo.py
+++ b/src/backend/demo/management/commands/create_demo.py
@@ -117,19 +117,20 @@ def create_demo(stdout):
         last_names = [fake.last_name() for _ in range(name_size)]
         for i in range(defaults.NB_OBJECTS["users"]):
             first_name = random.choice(first_names)
-            queue.push(
-                models.User(
-                    admin_email=f"user{i:d}@example.com",
-                    email=f"user{i:d}@example.com",
-                    password="!",
-                    is_superuser=False,
-                    is_active=True,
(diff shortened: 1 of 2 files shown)


sampaccoud:
This is not a good approach. Create demo is designed to test with tens of millions of objects. The proposed modification would make it too slow.

src/backend/demo/management/commands/create_demo.py
```diff
@@ -117,19 +117,20 @@ def create_demo(stdout):
...
-                    admin_email=f"user{i:d}@example.com",
-                    email=f"user{i:d}@example.com",
-                    password="!",
-                    is_superuser=False,
-                    is_active=True,
-                    is_staff=False,
-                    short_name=first_name,
-                    full_name=f"{first_name:s} {random.choice(last_names):s}",
-                    language=random.choice(settings.LANGUAGES)[0],
+            if not models.User.objects.filter(admin_email=f"user{i:d}@example.com").exists():
```
sampaccoud, line 120: Can you explain how the admin email can be duplicate since it is computed with the index of the loop. My understanding is that it is necessary unique per user :thinking:

sampaccoud:
Closing this PR as it would slow down the code and break our main use case which is to test our backend application with tens of millions of records in database.

PR #525, opened 2024-12-21
✨(backend) add option to configure list of required OIDC claims

sampaccoud:
## Purpose

We want to be able to refuse connection for users who have missing claims from a list of required keys.

## Proposal

Add option to configure list of required OIDC claims: USER_OIDC_REQUIRED_CLAIMS.
Defaults to an empty list.

### DIFF
--- a/src/backend/core/authentication/backends.py
+++ b/src/backend/core/authentication/backends.py
@@ -57,6 +57,18 @@ class OIDCAuthenticationBackend(MozillaOIDCAuthenticationBackend):
                     _("Invalid response format or token verification failed")
                 ) from e
 
+        # Validate required claims
+        missing_claims = [
+            claim
+            for claim in settings.USER_OIDC_REQUIRED_CLAIMS
+            if claim not in userinfo
+        ]
+        if missing_claims:
+            raise SuspiciousOperation(
+                _("Missing required claims in user info: %(claims)s")
+                % {"claims": ", ".join(missing_claims)}
+            )
+
         return userinfo
 
     def get_or_create_user(self, access_token, id_token, payload):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to
 
 ## Added
 
+🔧(backend) add option to configure list of required OIDC claims #525
 🔧(helm) add option to disable default tls setting by @dominikkaminski #519
 
 ## Changed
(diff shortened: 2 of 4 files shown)


dominikkaminski (approved):

AntoLC (approved):

src/backend/core/authentication/backends.py
```diff
@@ -57,6 +57,18 @@ def get_userinfo(self, access_token, id_token, payload):
...
+            claim
+            for claim in settings.USER_OIDC_REQUIRED_CLAIMS
+            if claim not in userinfo
+        ]
+        if missing_claims:
+            raise SuspiciousOperation(
+                _("Missing required claims in user info: %(claims)s")
+                % {"claims": ", ".join(missing_claims)}
+            )
+
```
lebaudantoine, line 71: according to the spec, these claims should be referred as essential claims. Please have a look here https://openid.net/specs/openid-connect-core-1_0.html#IndividualClaimsRequests.

lebaudantoine: You might prefer to override this method, and call it in `get_or_create_user`, as proposed in the initial implementation. `get_userinfo` should have a single responsibility, getting the data. 

```
    def verify_claims(self, claims):
        """Verify the provided claims to decide if authentication should be allowed."""

        # Verify claims required by default configuration
        scopes = self.get_settings("OIDC_RP_SCOPES", "openid email")
        if "email" in scopes.split():
            return "email" in claims

        LOGGER.warning(
            "Custom OIDC_RP_SCOPES defined. "
            "You need to override `verify_claims` for custom claims verification."
        )

        return True
```

https://github.com/mozilla/mozilla-django-oidc/blob/2c2334fdc9b2fc72a492b5f0e990b4c30de68363/mozilla_django_oidc/auth.py#L84

lebaudantoine: Capture d’écran 2024-12-27 à 16 40 05

sampaccoud: Interesting! :pray: 
Note that checking the presence of the "sub" claim was already wrong then :thinking: 
Let's fix that before it is released!

sampaccoud: @lebaudantoine Here you go: https://github.com/numerique-gouv/impress/pull/531

lebaudantoine: Kind of, checking the presence of "sub" was naive and simple. 
As soon as we introduce more complexity (a list check, a new setting, …) I would stick with the mozilla django oidc philosophy

PR #528, opened 2024-12-23
♻️Collaboration process

AntoLC:
## Purpose

It is the first step to introduce the long polling feature (see: #517)

We improve the architecture of the collaboration server, it will help us to introduce new routes without impacting the other routes.
We create `useProviderStore` on the frontend side, it will be dedicated to the provider. 
We add as well `useCollaboration` hook, that will manage the collaboration connections.

## Proposal

- [x] 🏗️(y-provider) organize yjs server
- [x] 🔒️(y-provider) add cors middlewares
- [x] ♻️(frontend) update already existing tasks
- [x] ♻️(frontend) create useProviderStore


src/frontend/apps/impress/src/stores/useBroadcastStore.tsx
```diff
@@ -25,26 +38,47 @@ export const useBroadcastStore = create<BroadcastState>((set, get) => ({
...
+      if (!isInitializing) {
+        action();
+      }
+    };
+
+    task.observe(observer);
+
+    setTimeout(() => {
+      isInitializing = false;
+    }, 1000);
```
PanchoutNathan, line 68: why ?

AntoLC: `task.observe` triggers the task lightly after you bind it to it, it seems like a bug, it is due to the architecture of y.js apparently. We want to exec the task only when we trigger it from broadcast method, so `isInitializing` is a trick to blocked this initial trigger.

https://github.com/numerique-gouv/impress/blob/375418df5438f36f48e8de21697aa650310137c2/src/frontend/apps/impress/src/stores/useBroadcastStore.tsx#L57-L68

PR #529, opened 2024-12-24
📸(helm) production-example

AntoLC:
## Purpose

We add a "production-example" environment to the helm chart. 
We have the "dev" environment already, but this one can be mistaken for a production, so we add a "production-example" to make it clear.  "dev" is for development, it is used by our Tilt stack.


AntoLC:
Closed in favor of https://github.com/suitenumerique/docs/pull/855

PR #530, opened 2024-12-24
Helm refactoring and documentation about self-hosted installation

rouja:
- add github ci to publish a helm chart
- add documentation for self-hosted deployment (without AI feature


AntoLC (approved):
So everytime we do a change we increment the version, and automatically it will be deploy ?

PR #543, opened 2025-01-10
Fix create document for user when sub does not match but email does

sampaccoud:
## Purpose

When creating a document on behalf of a user via the server-to-server API, a special edge case was broken that should should never happen but happens in our OIDC federation because one of the provider modifies the users "sub" each time they login.
    
We ended-up with existing users for who the email matches but not the sub. They were not correctly handled.

## Proposal

Get the user by its email like we do during authentication if activated in settings.
    
Made a few additional fixes and improvements to the endpoint.

### DIFF
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -264,12 +264,16 @@ class ServerCreateDocumentSerializer(serializers.Serializer):
         """Create the document and associate it with the user or send an invitation."""
         language = validated_data.get("language", settings.LANGUAGE_CODE)
 
-        # Get the user based on the sub (unique identifier)
+        # Get the user on its sub (unique identifier). Default on email if allowed in settings
         try:
             user = models.User.objects.get(sub=validated_data["sub"])
-        except (models.User.DoesNotExist, KeyError):
-            user = None
+        except (models.User.DoesNotExist, KeyError) as err:
+            if not settings.OIDC_FALLBACK_TO_EMAIL_FOR_IDENTIFICATION:
+                raise serializers.ValidationError(
+                    {"sub": ["Could not find user with this sub."]}
+                ) from err
             email = validated_data["email"]
+            user = models.User.objects.filter(email=email).first()
         else:
             email = user.email
             language = user.language or language
@@ -279,7 +283,9 @@ class ServerCreateDocumentSerializer(serializers.Serializer):
                 validated_data["content"]
             )
         except ConversionError as err:
-            raise exceptions.APIException(detail="could not convert content") from err
+            raise serializers.ValidationError(
+                {"content": ["Could not convert content"]}
+            ) from err
 
         document = models.Document.objects.create(
             title=validated_data["title"],
@@ -302,7 +308,11 @@ class ServerCreateDocumentSerializer(serializers.Serializer):
                 role=models.RoleChoices.OWNER,
             )
 
-        # Notify the user about the newly created document
+        self._send_email_notification(document, validated_data, email, language)
+        return document
+
+    def _send_email_notification(self, document, validated_data, email, language):
+        """Notify the user about the newly created document."""
         subject = validated_data.get("subject") or _(
             "A new document was created on your behalf!"
         )
@@ -312,8 +322,6 @@ class ServerCreateDocumentSerializer(serializers.Serializer):
             "title": subject,
         }
         document.send_email(subject, [email], context, language)
-
-        return document
 
     def update(self, instance, validated_data):
         """
--- a/bin/_config.sh
+++ b/bin/_config.sh
@@ -7,7 +7,7 @@ UNSET_USER=0
 
 TERRAFORM_DIRECTORY="./env.d/terraform"
 COMPOSE_FILE="${REPO_DIR}/docker-compose.yml"
-COMPOSE_PROJECT="impress"
+COMPOSE_PROJECT="docs"
 
 
 # _set_user: set (or unset) default user id used to run docker commands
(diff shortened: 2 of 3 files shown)


AntoLC (approved):

src/backend/core/api/serializers.py
```diff
@@ -264,12 +264,16 @@ def create(self, validated_data):
...
             user = models.User.objects.get(sub=validated_data["sub"])
-        except (models.User.DoesNotExist, KeyError):
-            user = None
+        except (models.User.DoesNotExist, KeyError) as err:
+            if not settings.OIDC_FALLBACK_TO_EMAIL_FOR_IDENTIFICATION:
+                raise serializers.ValidationError(
+                    {"sub": ["Could not find user with this sub."]}
+                ) from err
             email = validated_data["email"]
+            user = models.User.objects.filter(email=email).first()
```
AntoLC, line 276: user could be null actually here, so the creator could be null when you create the document, isn't it ?

sampaccoud: yep. This is being handled. When the creator is null it will be filled-in when the invitation completes...

qbey (approved):
Nice catch :)

src/backend/core/api/serializers.py
```diff
@@ -279,7 +283,9 @@ def create(self, validated_data):
                 validated_data["content"]
             )
         except ConversionError as err:
-            raise exceptions.APIException(detail="could not convert content") from err
+            raise serializers.ValidationError(
+                {"content": ["Could not convert content"]}
+            ) from err
```
lebaudantoine, line 288: 👍

src/backend/core/api/serializers.py
lebaudantoine: Question on my side, why not calling : 
```
User.objects.get(email__iexact=email)
```
As [here](https://github.com/suitenumerique/docs/blob/f12c06e975720682c598d12e0d0927d2c12d6918/src/backend/core/authentication/backends.py#L122)

src/backend/impress/settings.py
```diff
@@ -473,6 +473,11 @@ class Base(Configuration):
         environ_name="OIDC_FALLBACK_TO_EMAIL_FOR_IDENTIFICATION",
         environ_prefix=None,
     )
+    OIDC_ALLOW_DUPLICATE_EMAILS = values.BooleanValue(
+        default=False,
+        environ_name="OIDC_ALLOW_DUPLICATE_EMAILS",
+        environ_prefix=None,
+    )
```
lebaudantoine, line 480: nit: 
```
    ...
    description="WARNING: Enabling this setting allows multiple user accounts to share the same email address. "
                "This may cause security issues and is not recommended for production use if email is used for identification.",
    ...
```

lebaudantoine: btw, isn't risky to allow duplicate emails? Should it be enabled in production if `OIDC_FALLBACK_TO_EMAIL_FOR_IDENTIFICATION` is true?

I am not sure to fully understand which problem you are trying to solve with this setting

src/backend/core/models.py
```diff
@@ -126,6 +137,32 @@ def save(self, *args, **kwargs):
         super().save(*args, **kwargs)
 
 
+class UserManager(auth_models.UserManager):
```
lebaudantoine, line 140: love it!

src/backend/core/models.py
```diff
@@ -126,6 +137,32 @@ def save(self, *args, **kwargs):
...
+        except self.model.DoesNotExist as err:
+            if email and settings.OIDC_FALLBACK_TO_EMAIL_FOR_IDENTIFICATION:
+                try:
+                    return self.get(email=email)
+                except self.model.DoesNotExist:
+                    pass
+            elif (
+                self.filter(email=email).exists()
+                and not settings.OIDC_ALLOW_DUPLICATE_EMAILS
+            ):
```
lebaudantoine, line 156: (I might have misunderstood the code)

The email=None handling has a bug that can lead to unexpected behavior:
- When email=None, first branch is skipped via short-circuit (`if email and...`)
- But second branch still executes and checks for email=None in db
- This raises DuplicateEmailError if any user has email=None, despite never attempting email lookup

The error message about "email already associated" is misleading in this case since email is None. We should either skip duplicate check for None emails or handle them explicitly.

src/backend/core/api/serializers.py
```diff
@@ -265,16 +265,16 @@ def create(self, validated_data):
...
-                    {"sub": ["Could not find user with this sub."]}
-                ) from err
-            email = validated_data["email"]
-            user = models.User.objects.filter(email=email).first()
-        else:
+            user = models.User.objects.get_user_by_sub_or_email(
+                validated_data["sub"], email
+            )
+        except models.DuplicateEmailError as err:
+            raise serializers.ValidationError({"sub": [err.message]}) from err
```
lebaudantoine, line 275: Is the 'sub' key truly appropriate here? Reading the code, it seems misleading—the raised error is a DuplicateEmailError, but the serialized error pertains to 'sub'.

sampaccoud: True. Should be email.

src/backend/core/models.py
sampaccoud: Good catch thank you. 
Fixed it by passing when email is not set which will lead to creating a new user.

src/backend/impress/settings.py
sampaccoud: Agreed, duplicate email should not be activated when fallback is. 
But when fallback is false it could be activated or not... 
I should add a check that we don't have True+True.

lebaudantoine (approved):

PR #552, opened 2025-01-14
🏷️(backend) add content-type to uploaded files

AntoLC:
## Purpose

All the uploaded files had the content-type set to `application/octet-stream`. It create issues when the file is downloaded from the frontend because the browser doesn't know how to handle the file.

image

## Proposal

We improved the mimetype database and we now determine the content-type of the file and set it to the file object.


src/backend/core/api/viewsets.py
```diff
@@ -604,8 +605,19 @@ def attachment_upload(self, request, *args, **kwargs):
         extension = serializer.validated_data["expected_extension"]
         key = f"{document.key_base}/{ATTACHMENTS_FOLDER:s}/{file_id!s}.{extension:s}"
 
+        # Determine the content type of the file
+        file = serializer.validated_data["file"]
+        content_type, _ = mimetypes.guess_type(file.name)
```
sampaccoud, line 610: This was already done in the serializer. You don't need to do it here again : https://github.com/suitenumerique/docs/blob/main/src/backend/core/api/serializers.py#L367

Dockerfile
```diff
@@ -76,6 +76,8 @@ RUN apk add \
   pango \
   shared-mime-info
 
+RUN wget https://svn.apache.org/repos/asf/httpd/httpd/trunk/docs/conf/mime.types -O /etc/mime.types
```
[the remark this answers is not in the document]
AntoLC: Magic is to search the correct mime types I think, but it does not update the mime types database. 
Our image is an alpine, when the CI is the latest Ubuntu, so the mime types databases are different, it breaks the tests.

image

By updating the mime types database on our image and on the CI, we are sure that we are using the same mime types database.

src/backend/core/api/viewsets.py
```diff
@@ -605,7 +605,11 @@ def attachment_upload(self, request, *args, **kwargs):
         key = f"{document.key_base}/{ATTACHMENTS_FOLDER:s}/{file_id!s}.{extension:s}"
 
         # Prepare metadata for storage
-        extra_args = {"Metadata": {"owner": str(request.user.id)}}
+        content_type = serializer.validated_data["content_type"]
+        extra_args = {
+            "Metadata": {"owner": str(request.user.id)},
+            "ContentType": content_type,
+        }
```
lunika, line 612:
```suggestion
        extra_args = {
            "Metadata": {"owner": str(request.user.id)},
            "ContentType": serializer.validated_data["content_type"],
        }
```

src/backend/core/api/serializers.py
```diff
@@ -378,6 +378,10 @@ def validate_file(self, file):
         if extension_mime_type != magic_mime_type:
             self.context["is_unsafe"] = True
 
+        # mime_type fallback
+        if extension_mime_type is None:
```
lunika, line 382: `extension_mime_type` is a confusing name. `content_type` or `mime_type` is less error prone IMO.

sampaccoud: This was to differentiate between the mimetype inferred from the file extension and the mimetype inferred from the file content. I believe this distinction is important. Please take a look at the code above.

lunika: Yes but at the end it is this one used by the serializer as a validated data. The one provided by magic is only used to set a `is_unsafe` data and then seems never used.

sampaccoud (approved):

src/backend/core/migrations/00013_fix_content_types.py
```diff
@@ -0,0 +1,100 @@
+import magic
+from django.db import migrations
+from django.core.files.storage import default_storage
+
+def fix_attachments_content_type(apps, schema_editor):
```
lunika, line 5: All of this should be made in a management command. It is a not related to a database modification and tests should be added IMO

lunika: And so we have to create a job in the helm chart to run this kind of management command. I will made it in this branch if you are ok

AntoLC: Yes feel free to do it. 🙏

lunika:
@AntoLC I updated the PR accordingly with my last suggestion.

PR #553, opened 2025-01-14
✨(frontend) add home page

PanchoutNathan:
## Purpose

We currently do not have a homepage for the app. We're adding one by changing the login workflow a bit.

Capture d’écran 2025-01-30 à 09 59 47

Capture d’écran 2025-01-30 à 10 00 37


src/frontend/apps/impress/src/components/ProConnectButton.tsx
```diff
@@ -0,0 +1,33 @@
+import { css } from 'styled-components';
```
PanchoutNathan, line 1: I copied the one from visio, because as it is already online, it has therefore been approved. This is a component that I should put in the "ui-kit" package

AntoLC:
Close this pull request in favor of this one: https://github.com/suitenumerique/docs/pull/608

PR #562, opened 2025-01-16
✨(backend) add prometheus metrics

lindenb1:
Provides Prometheus metrics, a custom metrics exporter, a CIDR filter for monitoring targeted views,
and readiness/liveness probe endpoints for Kubernetes.

## Purpose

Simplify operational observability and health checking in containerized environments. By exposing both standard and custom metrics from Django for Prometheus, you can gain deeper insights into application performance. The integrated CIDR filter ensures that only allowed IPs can access the monitoring and probe endpoints of the application, and the built-in readiness/liveness probes provide health checks for Kubernetes to manage pod lifecycles effectively.

## Proposal

Integrate these changes to benefit from metrics collection, and streamlined health checks. By centralizing these responsibilities, teams can reduce complexity and maintain a clearer view of application behavior within Kubernetes clusters and similar environments.


lindenb1:
@virgile-dev Hey Virgile, could you please review the Prometheus implementation? At the moment, it is designed as an on/off switch, so it should not interfere significantly (if at all) with your existing code.

In addition to the Prometheus metrics, I have added the standard probes required by Kubernetes (readiness and liveness checks). I also included a CIDR filter decorator so that the metrics, readiness, and liveness endpoints are accessible only from specific networks, such as the private Kubernetes network where the application is running.

securitykernel:
@sampaccoud Can you or @AntoLC please do a review?

src/backend/impress/urls.py
```diff
@@ -1,22 +1,55 @@
...
+            name="liveness-probe",
+        ),
+    )
+    urlpatterns.append(
+        path(
+            "probes/readiness/",
+            monitoring_cidr_protected_view(readiness_check),
+            name="readiness-probe",
+        ),
+    )
```
sampaccoud, line 51: My understanding is that this would be redundant with what we already have on /__lbheartbeat__ and /__heartbeat__ with [Mozilla's dockerflow](https://github.com/mozilla-services/Dockerflow). 
Can you check before I review the rest?

src/backend/core/api/custom_metrics_exporter.py
```diff
@@ -0,0 +1,198 @@
...
+                "Number of active documents in the last 30 days",
+                value=doc_metrics["active_docs_last_30_days"],
+            )
+        )
+
+        # -- Document oldest/newest timestamps
+        if doc_ages["oldest"]:
+            metrics.append(
+                GaugeMetricFamily(
+                    prefixed_metric_name("oldest_document_date"),
```
simonpasquier, line 170: It's recommended to include "timestamp"in the metric name and have a _seconds suffix (from https://prometheus.io/docs/practices/naming/#metric-names)
```suggestion
                    prefixed_metric_name("oldest_document_timestamp_seconds"),
```

src/backend/core/api/custom_metrics_exporter.py
```diff
@@ -0,0 +1,198 @@
...
+                    "Timestamp of the oldest document creation date",
+                    value=doc_ages["oldest"].timestamp(),
+                )
+            )
+        if doc_ages["newest"]:
+            metrics.append(
+                GaugeMetricFamily(
+                    prefixed_metric_name("newest_document_date"),
```
simonpasquier, line 178:
```suggestion
                    prefixed_metric_name("newest_document_timestamp_seconds"),
```

src/backend/core/api/custom_metrics_exporter.py
```diff
@@ -0,0 +1,198 @@
...
+                    "Timestamp of the newest document creation date",
+                    value=doc_ages["newest"].timestamp(),
+                )
+            )
+
+        # -- User document distribution
+        user_distribution_metric = GaugeMetricFamily(
+            prefixed_metric_name("user_document_distribution"),
```
simonpasquier, line 186: (suggestion) not sure that the  "_distribution" is meaningful?
```suggestion
            prefixed_metric_name("user_documents"),
```

rvveber:
I'm re-opening this PR to take over.

- [x] Getting it up to date.
- [x] I'm talking to monitoring expert at B1 to see if we need anything else to fulfil best practice.
- [ ] Implement change requests
- [ ] Successful Review

src/backend/impress/urls.py
rvveber: Yes they are kind of redundant:

- The liveness probe `liveness_check` is essentially the same as the dockerflow heartbeat.
- But the readiness probe `readiness_check` reaches more deeply into the application, checking DB, cache and media backend before returning 200 so it catches failures Dockerflow would miss.

Options:
A: Remove these probes in favor for the less thorough heartbeat ones
B: Switch probing (`src/helm/impress/values.yaml:311`) to the custom probes

src/backend/core/api/custom_metrics_exporter.py
rvveber: Is now `user_documents_total`

src/backend/core/api/custom_metrics_exporter.py
rvveber: Is now  `documents_created_timestamp_seconds`
with `oldest | newest` as possible values for the label `bound`

src/backend/core/api/custom_metrics_exporter.py
rvveber: Is now  `documents_created_timestamp_seconds`
with `oldest | newest` as possible values for the label `bound`

src/backend/core/api/custom_metrics_exporter.py
simonpasquier: you shouldn't use a `_total` suffix for gauges, it's normally reserved for counters: https://prometheus.io/docs/specs/om/open_metrics_spec_1_1/#suffixes

rvveber: Oh okay
But it's an accumulated integer count.
So it should not be a gauge then - (code currently exclusively uses gauges).

I'll correct this, thanks

rvveber: Ah, apparently, they are not supposed to be counters, because they are not true counters in the sense that they do not reset on restart. Well then they need to stay gauges and we remove the _total suffix as you said

EDIT: Done

simonpasquier (commented):
LGTM. I just have one concern about the `user_documents` metric which may lead to excessive cardinality if values for the `user_email` label are unbounded.

rvveber:
> LGTM. I just have one concern about the `user_documents` metric which may lead to excessive cardinality if values for the `user_email` label are unbounded.

Yes, i agree, and i think its too much of an intrusive metric anyways.
We shouldn't include private data like E-Mails in metrics, so lets remove it.

Thank you for Reviewing!

src/backend/impress/urls.py
rvveber: @lunika can you help me with the decision?

lunika: Yes, reviewing this PR is in my todo list of this week.

src/backend/core/api/decorators.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    Decorator to restrict access to a view based on CIDR ranges.
+
+    Checks the client's IP address against allowed CIDR ranges specified
+    in the MONITORING_ALLOWED_CIDR_RANGES environment variable. If the
+    IP address is not within the allowed ranges, access is denied.
+    """
+
+    def wrapped_view(request, *args, **kwargs):
+        cidr_env_raw = os.environ.get("MONITORING_ALLOWED_CIDR_RANGES", "")
```
Ash-Crow: Environment variables should be centralized in the `settings.py` file (please see reasoning for this in https://adamj.eu/tech/2022/11/24/django-settings-patterns-to-avoid/#avoid-creating-dynamic-module-constants-instead-of-settings )

src/backend/impress/urls.py
```diff
@@ -1,22 +1,55 @@
...
+from core.api.custom_probe_views import liveness_check, readiness_check
+from core.api.decorators import monitoring_cidr_protected_view
+
 urlpatterns = [
     path("admin/", admin.site.urls),
     path("", include("core.urls")),
 ]
 
+# Conditionally add Prometheus Exporter endpoint
+if os.environ.get("MONITORING_PROMETHEUS_EXPORTER", "False").lower() == "true":
```
Ash-Crow: Same comment as before: `MONITORING_PROMETHEUS_EXPORTER` and `MONITORING_PROBING` should be pulled from settings through `django.conf`

src/backend/core/api/custom_metrics_exporter.py
```diff
@@ -0,0 +1,163 @@
...
+    def collect(self):
+        """
+        Collect and yield Prometheus metrics for user activity, document activity,
+        and document statistics over various time periods.
+        """
+
+        namespace = getattr(settings, "PROMETHEUS_METRIC_NAMESPACE", "")
+
+        def metric_name(name):
+            return f"{namespace}_{name}" if namespace else name
```
lunika: Should be extracted in a private method, before declaring the `collect` method

```suggestion
def _get_metric_name(self, name):
    namespace = settings.PROMETHEUS_METRIC_NAMESPACE
    return f"{namespace}_{name}" if namespace else name
```

and then call `self._get_metric_name` everywhere you need it inside this class.

src/backend/core/api/custom_metrics_exporter.py
```diff
@@ -0,0 +1,163 @@
...
+        doc_metrics = {
+            "count": models.Document.objects.count(),
+            "shared": (
+                models.Document.objects.annotate(access_count=Count("accesses"))
+                .filter(access_count__gt=1)
+                .count()
+            ),
+            "active_today": models.Document.objects.filter(
+                updated_at__gte=times["today_start_utc"],
+                updated_at__lt=times["today_start_utc"] + timedelta(days=1),
```
lunika: Not sure why this line is needed. And if it is, why not doing the same for the users.

src/backend/core/api/custom_probe_views.py
```diff
@@ -0,0 +1,85 @@
...
+from botocore.exceptions import BotoCoreError, ClientError
+
+
+def liveness_check(request):
+    """
+    Liveness probe endpoint.
+    Returns HTTP 200 if the application is alive and running.
+    """
+
+    return JsonResponse({"status": "OK"}, status=200)
```
lunika: This already exists in the `dockerflow` package

src/backend/core/api/custom_probe_views.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def readiness_check(request):
```
lunika: Instead of recreating a readiness endpoint, you should use what is provided by the `dockerflow` library.

The `dockerflow` library use the [django checks](https://docs.djangoproject.com/en/5.2/topics/checks/) and register there own custom checks. When the `hertbeat` endpoint is called, all checks are run, if one fail, the heartbeat returns a 500.

I suggest you to register the checks you wrote in the `DOCKERFLOW_CHECKS` settings and dockerflow will use them in the `heartbeat` endpoint. 

See the dockerflow documentation to understand how it works and how to extend it https://python-dockerflow.readthedocs.io/en/main/django.html

src/backend/core/api/custom_probe_views.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    Readiness probe endpoint.
+    Checks database, cache, media storage, and OIDC configuration.
+    Returns HTTP 200 with JSON status "OK" if all checks pass,
+    or HTTP 500 with JSON status "Error" and an error message.
+    """
+
+    def check_database():
```
lunika: If you consider my previous comment, this check already exists in the `dockerflow` library

src/backend/core/api/decorators.py
```diff
@@ -0,0 +1,50 @@
...
+                    ip_network(c.strip().strip('"').strip("'"))
+                    for c in cidr_env_stripped.split(",")
+                    if c.strip()
+                ]
+            except ValueError as e:
+                raise ValueError(
+                    f"Invalid CIDR range in MONITORING_ALLOWED_CIDR_RANGES: {e}"
+                ) from e
+
+        client_ip = request.META.get("REMOTE_ADDR")
```
lunika: We can't rely on this security. First this value will be probably wrong if the django application is behind multiple proxy. Also it is possible to spoof the ip.

I look how to secure a prometheus metrics endpoint and I found that it can be secure using a basic auth mechanisms. DRF provides a BaseAuth backend : https://www.django-rest-framework.org/api-guide/authentication/#basicauthentication

src/backend/core/tests/monitoring/test_probes_cidr.py
```diff
@@ -0,0 +1,179 @@
+"""Test liveness and readiness CIDR protected probes of impress's core app."""
+
+import os
+from unittest.mock import patch
+
+from django.test import TestCase
+
+
+class ProbesCidrProtectionTest(TestCase):
```
lunika: We are not using django TestCase in this project. We are using pytest django instead. See how other tests are implemented

src/backend/impress/wsgi.py
```diff
@@ -11,7 +11,26 @@
...
+    ):
+        REGISTRY.register(CustomMetricsExporter())
+
+
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "impress.settings")
 os.environ.setdefault("DJANGO_CONFIGURATION", "Development")
 
+if os.environ.get("MONITORING_PROMETHEUS_EXPORTER", "False").lower() == "true":
+    register_prometheus_exporter()
+
```
lunika: Not sure to understand why you are doing all of this here. Why not relying with what is made in the `django-prometheus` library ?

lunika (commented):
We have to find an other way to expose the `/metrics` endpoint. Filtering by ip is not a secure solution IMO.
Apparently the way to secure a prometheus metrics endpoint is to use basic auth.

As mention by @Ash-Crow you have to rely on django settings instead of doing `os.getenv` everywhere you need the environment variables.

lunika:
To securize the `/metrics` endpoint, I think a new view useing DRF View classes should be created, on this view the basic auth should be configured as authenticator class.
In this view you should recreate what is made in the django-prometheus [exporter](https://github.com/django-commons/django-prometheus/blob/df29fd3d30a51678382906848482ac3b3bad85e8/django_prometheus/exports.py#L111) and return a DRF response instead of a django one.
And this view should be declared in the url.

Doing this I think you dooing this and followin django-promotheus documentation on how to install the project I think you will not need anything more what is made in the `wsgi` module. This module should leave untouch.

src/backend/core/api/custom_metrics_exporter.py
rvveber: You're right, it should not be needed under normal circumstances it should never happen that a user/doc is created/updated with a future `updated_at` time.

src/backend/impress/wsgi.py
rvveber: `django_prometheus` just handles default exporters, and provides decorated backends for Cache and Database for the Django Framework.
[Adding custom business metrics like ours, we need to use `prometheus_client` directly](https://github.com/django-commons/django-prometheus/blob/master/README.md#adding-your-own-metrics)

Doing the registration in WSGI is a valid way to do it, because it's the entrypoint, and one way to make sure the registration of the `prometheus_client` is happening only once per process.

But in Django it's also possible and maybe better to register lazily in `apps.py`
i did that instead, see if you like it more, i think it's cleaner

src/backend/impress/urls.py
rvveber: The custom probes have been removed.

src/backend/core/api/custom_probe_views.py
rvveber: Thanks for your help and guidance.
I have removed the custom probes and expanded the dockerflow probes with django checks like you recommended.
Much cleaner!

src/backend/core/tests/monitoring/test_probes_cidr.py
rvveber: Right. Fixed.

src/backend/core/api/decorators.py
rvveber: I removed CIDR protection and switched to a BasicAuth implementation.
Looks like DRF needs a user for BaseAuth, so i created a dummy user `MonitoringUser`.
Indeed its way better to like this 👍🏼

rvveber:
@lunika Could you review the updated version?
I've implemented your change requests.
In addition i fully utilized everything that django_prometheus allows to do:

- Using the Mixins to get generic stats for the most relevant Entities
- Using the Decorated Backends for any type of configured DB and Cache
- - Workaround for some Backends that are broken from django_prometheus upstream + document when and how to remove

src/backend/core/api/custom_collector.py
```diff
@@ -0,0 +1,201 @@
...
+    - leading application (namespace) prefix configured via settings
+    - snake_case names with explicit unit/type suffix where relevant
+    - use labels inside the metric name
+    """
+
+    def _get_metric_name(self, name):
+        """
+        Get the Prometheus metric if PROMETHEUS_METRIC_NAMESPACE is set.
+        """
+        namespace = getattr(settings, "PROMETHEUS_METRIC_NAMESPACE", "")
```
lunika, line 30:
```suggestion
        namespace = settings.PROMETHEUS_METRIC_NAMESPACE
```

The setting exists in the settings module, you can directly use it.

src/backend/core/checks.py
```diff
@@ -0,0 +1,79 @@
...
+from django.core.checks import Error, register
+from django.core.exceptions import SuspiciousFileOperation
+from django.core.files.base import ContentFile
+from django.core.files.storage import default_storage
+from django.db import connections
+from django.db.utils import OperationalError
+
+
+@register()
+def database_check(app_configs, **kwargs):
```
lunika, line 16: This check already exists in dockerflow and I think they are checking more things : https://github.com/mozilla-services/python-dockerflow/blob/3221444e8994e38cd77e520c87e2a7cc523003a4/src/dockerflow/django/checks.py#L14

src/backend/impress/settings.py
```diff
@@ -76,6 +77,34 @@ class Base(Configuration):
     ROOT_URLCONF = "impress.urls"
     WSGI_APPLICATION = "impress.wsgi.application"
 
+    # Monitoring
+    DOCKERFLOW_CHECKS = [
```
lunika, line 81: IMO you also have to keep existing checks from dockerflow : 

```
"dockerflow.django.checks.check_database_connected",
"dockerflow.django.checks.check_migrations_applied",
```

src/backend/impress/settings.py
```diff
@@ -846,6 +876,48 @@ def post_setup(cls):
...
+                )
+            cls.MIDDLEWARE = [
+                "django_prometheus.middleware.PrometheusBeforeMiddleware",
+                *cls.MIDDLEWARE,
+                "django_prometheus.middleware.PrometheusAfterMiddleware",
+            ]
+            
+            _PROMETHEUS_DB_BACKENDS = {
+                "django.db.backends.sqlite3": "django_prometheus.db.backends.sqlite3",
+                "django.db.backends.postgresql": "django_prometheus.db.backends.postgresql",
```
lunika, line 894: We are not supporting other backends than postgresql, I don't know if it's needed to have all the engine here.

kosssi:
This morning I upgrade Docs in 1.8.2 I forgot to perform the DB migration, which caused the API-side instance to return a 500 error. My monitoring didn't flag the error (frontend is 200). Could this issue have alerted me to it?

lunika:
Not sure to understand about what issue you are talking about.
Is your backend deployment using probes to determine if the application is ready to accept traffic ?

PR #566, opened 2025-01-17
✨(frontend) add Alert, Quote, and Divider blocks to the editor

arnaud-robin:
## Purpose

Add new block types to better reproduce current text editor experience that user need.

## Proposal

- Introduced new Alert, Quote, and Divider components for enhanced document editing.
- Updated BlockNoteEditor to include these new block types in the schema.
- Implemented corresponding slash menu items for easy insertion of these blocks.
- Added translations for new block types and their descriptions in the i18n files.


AntoLC:
We decided to improve the Alert block, we will split this PR, works started here: #646

PR #585, opened 2025-01-28
✅(e2e) fix flakiness

AntoLC:
## Purpose

Some flakiness appeared in the e2e tests. It started to impact many pull requests. Time to fix them.


src/frontend/apps/e2e/__tests__/app-impress/doc-search.spec.ts
```diff
@@ -24,91 +25,59 @@ test.describe('Document search', () => {
...
+    await expect(rowdoc.getByText('keyboard_return')).toBeVisible();
+    await expect(rowdoc.getByText(/seconds? ago/)).toBeVisible();
   });
 
   test('it checks search for a document', async ({ page, browserName }) => {
-    const id = Math.random().toString(36).substring(7);
-
-    const doc1 = await createDoc(page, `My super ${id} doc`, browserName, 1);
-    await verifyDocName(page, doc1[0]);
-    await page.goto('/');
```
PanchoutNathan, line 37: I don't think this step should be deleted. Because potentially, this button will not have the same behavior with multi-page for example. Or maybe a future screen. We might as well go back to the place where we are sure that this is the desired behavior. WDYT ?

src/frontend/apps/e2e/__tests__/app-impress/doc-search.spec.ts
```diff
@@ -24,91 +25,59 @@ test.describe('Document search', () => {
...
-
-    response = await responsePromisePage;
-    result = (await response.json()) as { results: SmallDoc[] };
-    docs = result.results;
+    await inputSearch.fill('My doc search super');
 
-    expect(docs.length).toEqual(1);
+    await expect(
+      listSearch.getByRole('option').getByText(doc2Title),
+    ).toBeHidden();
```
PanchoutNathan, line 81: Maybe check that the doc "My doc search super" is present

src/frontend/apps/e2e/__tests__/app-impress/common.ts
```diff
@@ -91,6 +92,18 @@ export const addNewMember = async (
   return users[index].email;
 };
 
+export const getGridRow = async (page: Page, title: string) => {
+  const docsGrid = page.getByRole('grid');
+  await expect(docsGrid).toBeVisible();
+  await expect(docsGrid.getByTestId('grid-loader')).toBeHidden();
```
PanchoutNathan, line 98: This loader is not present directly in the grid, use page. can you do the same for "goToGridDoc" method ? 

```
  await expect(page.getByTestId('grid-loader')).toBeHidden();
```

src/frontend/apps/e2e/__tests__/app-impress/common.ts
```diff
@@ -40,7 +40,8 @@ export const createDoc = async (
       })
       .click();
 
-    const input = page.getByRole('textbox', { name: 'doc title input' });
+    const input = page.getByLabel('doc title input');
+    await expect(input).toHaveText('');
```
PanchoutNathan, line 44: why test this ? can be better to test if the input is visible

src/frontend/apps/e2e/__tests__/app-impress/common.ts
```diff
@@ -91,6 +92,18 @@ export const addNewMember = async (
...
+export const getGridRow = async (page: Page, title: string) => {
+  const docsGrid = page.getByRole('grid');
+  await expect(docsGrid).toBeVisible();
+  await expect(docsGrid.getByTestId('grid-loader')).toBeHidden();
+
+  const rows = docsGrid.getByRole('row');
+
+  return rows.filter({
+    hasText: title,
+  });
```
PanchoutNathan, line 104: To avoid going further, I would test if the rose is well defined before returning it. WDYT ?

src/frontend/apps/e2e/__tests__/app-impress/doc-editor.spec.ts
```diff
@@ -220,17 +220,19 @@ test.describe('Doc Editor', () => {
...
     await expect(editor.getByText('Hello World Doc persisted 1')).toBeVisible();
 
-    const secondDoc = await goToGridDoc(page, {
-      nthRow: 2,
-    });
+    const [secondDoc] = await createDoc(
+      page,
+      'doc-saves-change-other',
+      browserName,
+    );
```
PanchoutNathan, line 235: Agreed with this, WP :)

src/frontend/apps/e2e/__tests__/app-impress/doc-member-create.spec.ts
```diff
@@ -40,7 +40,9 @@ test.describe('Document create member', () => {
     await expect(
       list.getByTestId(`doc-share-add-member-${users[0].email}`),
     ).toBeVisible();
-    await expect(list.getByText(`${users[0].full_name}`)).toBeVisible();
+    await expect(
+      list.getByText(`${users[0].full_name || users[0].email}`),
+    ).toBeVisible();
```
PanchoutNathan, line 45: nice catch

src/frontend/apps/e2e/__tests__/app-impress/doc-search.spec.ts
```diff
@@ -1,20 +1,21 @@
...
-  updated_at: string;
-};
-
 test.beforeEach(async ({ page }) => {
   await page.goto('/');
 });
 
 test.describe('Document search', () => {
-  test('it checks all elements are visible', async ({ page }) => {
+  test('it checks all elements are visible', async ({ page, browserName }) => {
```
PanchoutNathan, line 10: maybe merge the two tests because we are duplicating the same test actions a bit, I find it a bit of a shame to take time for nothing. I don't have a strong opinion. Just a note

PanchoutNathan (approved):
Well done! Several good catches, it's really cool :)

src/frontend/apps/e2e/__tests__/app-impress/common.ts
AntoLC: When you create a new doc from a doc, the value does not have the time to be clear cos the provider takes a few milliseconds to clear. To add a check that the value is empty, it adds a bit of times for the provider to be clean.

PanchoutNathan: all right

src/frontend/apps/e2e/__tests__/app-impress/doc-search.spec.ts
AntoLC: Done

PR #616, opened 2025-02-06
🌐(CI) build mails when crowdin_upload workflow

AntoLC:
## Purpose

When we were executing the `crowdin_upload` workflow, we were not building the mail template to dispatch it to the backend. 
It resulted in the mail not being totally translated. 

## Proposal

This commit fixes that issue by adding the build mail step to the `crowdin_upload` workflow. 
To do so, we added it to the `dependencies` workflow, a workflow callable by other workflows that need a specific job.

- [x] 🌐(CI) build mails when crowdin_upload workflow
- [x] 🐛(i18n) add back the missing email translations
- [x] 💄(email) fix line height email title

## Prob

image


.github/workflows/crowdin_upload.yml
```diff
@@ -29,6 +31,12 @@ jobs:
       - name: Install development dependencies
         run: pip install --user .
         working-directory: src/backend
+      - name: Restore the mail templates
+        uses: actions/cache@v4
+        id: mail-templates
+        with:
+          path: "src/backend/core/templates/mail"
+          key: mail-templates-${{ hashFiles('src/mail/mjml') }}
```
lunika, line 39: you should probably add `fail-on-cache-miss: true`

lunika (approved):

PR #619, opened 2025-02-07
⚗️Other type cursors

AntoLC:
## Purpose

Test other kind of cursors.

## Demo
image


AntoLC:
It was an experimentation to showcase, not useful anymore.

PR #629, opened 2025-02-11
🛂(frontend) Restore version visibility

AntoLC:
## Purpose

When you are a reader member, you have the right to see the version but you cannot restore them. 
So, we hide the restore button when you are a reader.

## Proposal

- [x] 🩹(docker-compose) remove comment
- [x] 🛂(frontend) hide version restore button when reader

## Where

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to
 
 ## Changed
 
+- 🛂(frontend) Restore version visibility #629
 - 📝(doc) minor README.md formatting and wording enhancements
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -15,7 +15,7 @@ services:
       - "1081:1080"
 
   minio:
-    # user: ${DOCKER_USER:-1000}
+    user: ${DOCKER_USER:-1000}
```
lunika, line 18: Already modified in main branch

lunika (approved):

PR #630, opened 2025-02-11
✨(ai) A11Y Hackathon - Chat with your doc & PDF ! 

NathanVss:


lunika:
What the state of this PR ? Should we finish it or it's just a POC without usage and we can close it ?

AntoLC:
I added the label "keep track", to find interesting PR or issue easily after they are closed.

PR #633, opened 2025-02-12
🐛(backend) race condition create doc

AntoLC:
## Purpose

When 2 docs are created almost at the same time, the second one will fail because the first one.
We get a unicity error on the path key already used ("impress_document_path_key").
This error creates e2e flakiness, we have some cases as well in production.

## Proposal

To fix this issue, we will lock the table the time to create the document, the next query will wait for the lock to be released.


src/backend/core/api/viewsets.py
```diff
@@ -573,6 +573,12 @@ def retrieve(self, request, *args, **kwargs):
     @transaction.atomic
     def perform_create(self, serializer):
         """Set the current user as creator and owner of the newly created object."""
+        with connection.cursor() as cursor:
+            cursor.execute(
+                f'LOCK TABLE "{models.Document._meta.db_table}" '  # noqa: SLF001
+                "IN SHARE ROW EXCLUSIVE MODE;"
```
lunika, line 579: You should add a comment explaining this mode.

sampaccoud: should be done when creating children as well? See [children method on the viewset](https://github.com/suitenumerique/docs/blob/main/src/backend/core/api/viewsets.py#L744).

src/backend/core/tests/documents/test_api_documents_create.py
```diff
@@ -51,6 +52,35 @@ def test_api_documents_create_authenticated_success():
...
+        future1 = executor.submit(create_document, "my document 1")
+        future2 = executor.submit(create_document, "my document 2")
+
+        response1 = future1.result()
+        response2 = future2.result()
+
+        assert response1.status_code == 201
+        assert response2.status_code == 201
+
+
```
sampaccoud, line 83: Add test for the children use case as well?

lunika: Yes we are working on it but it much more difficult to reproduce

src/backend/core/api/viewsets.py
qbey: As discussed with lunika, I don't think the lock should be managed by the viewset, is should be at the model/manager level.
And as you mentioned, the same could happen to children/siblings, it should be covered also.

AntoLC: We did try to add it in the model, see [here](https://github.com/suitenumerique/docs/compare/1140eef52f3ccb0ed94773e5cbfc5544d5e0fe6f?diff=unified&w=#diff-21170f63f26dc81c71cf947970745356a020118624a97a25988b67f4c40fdf3cR819-R833), but it brings instability in our tests with the factory, a fix brings another instability, and so on.

This naive way is much more stable.
Most of my flakiness is gone as well:
image

src/backend/core/tests/documents/test_api_documents_create.py
AntoLC: Done

src/backend/core/tests/documents/test_api_documents_children_create.py
```diff
@@ -249,3 +250,41 @@ def test_api_documents_children_create_force_id_existing():
...
+    """
+
+    user = factories.UserFactory()
+
+    client = APIClient()
+    client.force_login(user)
+
+    document = factories.DocumentFactory()
+
+    factories.UserDocumentAccessFactory(user=user, document=document, role="owner")
```
lunika, line 269: You are not testing concurrency on children creation. But on parent creation. Children will be created on different parent so they have a different path, so you will never have a unique constraint error in you test.

AntoLC: No, look at here: https://github.com/suitenumerique/docs/pull/633/files#diff-719c99bace85294f27e3909b1067f53da483436c448af38d93a7856fb5ae36faR290

lunika (approved):

PR #634, opened 2025-02-12
♻️Stop setting a default title on doc creation

AntoLC:
## Purpose

We were setting a default title to our document during creation, but we should not do that, it created lot of similar titles, lot of documents will show up during search.

## Proposal

- [x] 🩹(frontend) minor fixes
- [x] ♻️(frontend) stop setting a default title on doc creation
- [x] ➕(backend) add django-test-migrations
- [x] 🩹(migration) add migration to update default titles
- [x] 🦺(migration) add back the migration folders to linter


src/backend/core/migrations/0018_update_blank_title.py
```diff
@@ -0,0 +1,22 @@
...
+    ).update(title=None)
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("core", "0017_add_fields_for_soft_delete"),
+    ]
+
+    operations = [
+        migrations.RunPython(update_titles_to_null),
```
sampaccoud, line 21: It's better to add a reverse migration with noop so that migrations run both ways.

src/frontend/apps/impress/src/features/docs/docs-grid/components/DocsGridLoader.tsx
```diff
@@ -27,15 +26,11 @@ export const DocsGridLoader = ({ isLoading }: DocsGridLoaderProps) => {
...
         $width="100%"
-        $maxWidth="960px"
-        $background="rgba(255, 255, 255, 0.3)"
+        $background="rgba(255, 255, 255, 0.5)"
         $zIndex={998}
-        $position="fixed"
-        $css={css`
-          top: ${HEADER_HEIGHT}px;
-        `}
+        $position="absolute"
```
sampaccoud, line 33: Changes related to css would require a separate commit for different explanations?

sampaccoud (approved):

src/frontend/apps/impress/src/features/docs/docs-grid/components/DocsGridLoader.tsx
AntoLC: You'r right was a mistake. You can read front so :D

PR #636, opened 2025-02-13
Fix user search

sampaccoud:
## Purpose

We don't want user search to be triggered for very short query strings.
Also, we are currently returning users who have been disabled which is dysfunctional.

## Proposal

- [x] require a query string of minimum 5 characters for a user search
- [x] Filter inactive users out of search results

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to
 
 ## Fixed
 
+- 🐛(backend) stop returning inactive users on the list endpoint #635
 - 🔒️(backend) require at least 5 characters to search for users #635
 
 ## [2.2.0] - 2025-02-10
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+## Fixed
+
+- 🔒️(backend) require at least 5 characters to search for users #635
+
 ## [2.2.0] - 2025-02-10
 
 ## Added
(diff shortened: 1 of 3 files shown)


src/backend/core/api/viewsets.py
AntoLC, line 1: We should maybe add a limit as well, let's say we have 2000 "michel", we maybe don't want to list all of them, maybe the  50 first, wdyt ?

AntoLC, line 1: It could have been interesting to add a throttle as well , to avoid the bots to pull the data.

The AI system is quite nice, we throttle per doc, it could have been something like `GET document/[id]/users?q=`; longer to put in place though.

src/backend/core/api/viewsets.py
```diff
@@ -160,8 +160,8 @@ def get_queryset(self):
         if document_id := self.request.GET.get("document_id", ""):
             queryset = queryset.exclude(documentaccess__document_id=document_id)
 
-        if not (query := self.request.GET.get("q", "")):
-            return queryset
+        if not (query := self.request.GET.get("q", "")) or len(query) < 5:
```
lunika, line 163: `3` is I think "less" aggressive. 5 probably match for european or "occidental" names but we all have example of name length of only 1 character (like is asia).

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ and this project adheres to
 ## Fixed
 
 - 🐛(back) allow only images to be used with the cors-proxy #781
+- 🐛(backend) stop returning inactive users on the list endpoint #635
+- 🔒️(backend) require at least 5 characters to search for users #635
```
AntoLC, line 23:
```suggestion
- 🐛(backend) stop returning inactive users on the list endpoint #636
- 🔒️(backend) require at least 5 characters to search for users #636
```

lunika (approved):

PR #637, opened 2025-02-14
increase nginx buffer size when proxifying keycloak 

lunika:
## Purpose

Nginx is used to proxify keycloak in our development configuration. When
a new user is created keycloak is send a large amount of headers in its
response and the default nginx config is not enough to handle this
amount of headers. We have to increase the proxy buffer size to handle
them.

We all remove the usage of dockerize in this PR to use healthcheck in docker compose services.
And then change all make run commands. The run command is starting everything. The run-back
command is starting all services needed to use the backend application
and then the run-frontend starts only the frontend-dev service.

## Proposal

- [x] increase nginx buffer size when proxifying keycloak 
- [x] remove usage of dockerize
- [x] make run command starting everything

### DIFF
--- a/.github/workflows/impress-frontend.yml
+++ b/.github/workflows/impress-frontend.yml
@@ -88,28 +88,6 @@ jobs:
       - name: Start Docker services
         run: make bootstrap FLUSH_ARGS='--no-input' cache=
 
-      # Tool to wait for a service to be ready
-      - name: Install Dockerize
-        run: |
-          curl -sSL https://github.com/jwilder/dockerize/releases/download/v0.8.0/dockerize-linux-amd64-v0.8.0.tar.gz | sudo tar -C /usr/local/bin -xzv
-
-      - name: Wait for services to be ready
-        run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/impress-frontend.yml
```diff
@@ -88,28 +88,6 @@ jobs:
...
-          printf "Server collaboration check...\n"
-          dockerize -wait tcp://localhost:4444 -timeout 20s
-          printf "Ngnix check...\n"
-          dockerize -wait tcp://localhost:8083 -timeout 20s
-          printf "DRF check...\n"
-          dockerize -wait tcp://localhost:8071 -timeout 20s
-          printf "Postgres Keyclock check...\n"
-          dockerize -wait tcp://localhost:5433 -timeout 20s
-          printf "Postgres back check...\n"
-          dockerize -wait tcp://localhost:15432 -timeout 20s
```
AntoLC, line 111: It solves some flakiness, but let's see.

AntoLC (approved):

PR #639, opened 2025-02-14
♻️(front) improve table pdf rendering

NathanVss:
## Purpose

The previous way of rendering table was causing issues when tables could not fit on one page. I then came accross this discussion https://github.com/diegomura/react-pdf/issues/2343. The author created a lib to improve the rendering of table, it's better, but still not perfect maybe.

> Improving PDF is quite cumbersome, so I encourage you to test on your side to see if the rendering is correct. Otherwise we'll need to put more effort into this by going deep in the implementation, not sure we want to put that much of energy in it. Let's try this lib before :)

### The Doc
Capture d’écran 2025-02-14 à 15 32 46

### Before 
Capture d’écran 2025-02-14 à 15 32 22

### After
Capture d’écran 2025-02-14 à 15 33 05

### DIFF
--- a/src/frontend/apps/impress/src/features/docs/doc-header/components/ModalExport.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-header/components/ModalExport.tsx
@@ -27,6 +27,8 @@ import { Doc } from '@/features/docs/doc-management';
 import { TemplatesOrdering, useTemplates } from '../api/useTemplates';
 import { downloadFile, exportResolveFileUrl } from '../utils';
 
+import { Table } from './blocks/Table';
+
 enum DocDownloadFormat {
   PDF = 'pdf',
   DOCX = 'docx',
@@ -114,6 +116,7 @@ export const ModalExport = ({ onClose, doc }: ModalExportProps) => {
                     : 1.17;
               return (
                 <PDFText
+                  key={block.id}
                   style={{
                     fontSize: fontSizeEM * FONT_SIZE * PIXELS_PER_POINT,
                     fontWeight: 700,
@@ -150,6 +153,9 @@ export const ModalExport = ({ onClose, doc }: ModalExportProps) => {
                   {exporter.transformInlineContent(block.content)}
                 </PDFText>
               );
+            },
+            table: (block, transformer) => {
+              return <Table data={block.content} transformer={transformer} />;
             },
           },
         },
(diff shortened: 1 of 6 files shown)


src/frontend/apps/impress/src/features/docs/doc-header/components/ModalExport.tsx
```diff
@@ -151,6 +153,9 @@ export const ModalExport = ({ onClose, doc }: ModalExportProps) => {
                 </Text>
               );
             },
+            table: (block, t) => {
+              return <Table data={block.content} transformer={t} />;
+            },
```
AntoLC, line 158: nit: `t` is often referenced in the project as translation, could be more specific to avoid confusion ?

```suggestion
            table: (block, transformer) => {
              return <Table data={block.content} transformer={transformer} />;
            },
```

src/frontend/apps/impress/src/features/docs/doc-header/components/blocks/Table.tsx
```diff
@@ -0,0 +1,60 @@
...
+}) => {
+  return (
+    <TablePDF>
+      {props.data.rows.map((row, index) => {
+        if (index === 0) {
+          return (
+            <TH key={index}>
+              {row.cells.map((cell, index) => {
+                return (
+                  <TD key={index}>
```
AntoLC, line 32: I have a few warning about identical key, maybe it can be fix easily ?

image

AntoLC:
It could be nice to have it integrated by default in Blocknotes wdyt @YousefED ?

src/frontend/apps/impress/src/features/docs/doc-header/components/blocks/Table.tsx
NathanVss: I do not have these errors, can you give me your doc so I can try on my side?

YousefED:
> It could be nice to have it integrated by default in Blocknotes wdyt @YousefED ?

Cool! Great to see this. I'd be happy to add this to react-pdf. @NathanVss are you able to do this? Otherwise we add it to our todos

I see two ways:
a) migrate to react-pdf-table in BlockNote
b) find the part in react-pdf-table that actually fixes the cross-page behavior (or the part in BlockNote that breaks it). I just spend some time to find this, but didn't immediately spot the cause

AntoLC (approved):
Nice improvement ^^

YousefED:
@NathanVss I was wrong in standup, react-pdf depends on pdfkit, not pdfjs

For pdfkit, this issue is in progress: https://github.com/foliojs/pdfkit/pull/1577

Once that lands it should be possible to add it to react-pdf and then to blocknote / docs

PR #642, opened 2025-02-17
🐛(front) improve text rendering in pdf

NathanVss:
## Purpose

The rendered text had unwanted line breaks in middle of them. It was because we were not using the appropriate Text component, the one to be used in the one from react-pdf.

## The doc

Capture d’écran 2025-02-17 à 10 58 05

## PDF Before

Capture d’écran 2025-02-17 à 10 58 23

## PDF After

Capture d’écran 2025-02-17 à 10 58 31


AntoLC (approved):
Well seen !

PR #645, opened 2025-02-18
Adjustments to finalize frontend with tree structure

sampaccoud:
## Purpose

We recently added document tree structure to the backend but the frontend is still a work in progress :construction: 
While working on the frontend, we identified some issues and missing features on the backend.

- We want to be able to make a search query inside a hierarchical document.
- If a document already gets a link reach/role inheriting from one of its ancestors, we should not propose setting link reach/role on the document that would be more restrictive than what we inherited from ancestors.
- Filtering does not work on the children view
- We need to be able to search on all descendants of a given document
- We need to be able to retrieve the opened ancestors tree for the current document in order to display the tree structure in the left side bar, opened at the current document.
- when getting documents or its children/descendants, we want to be able to force page size to a greater value than 20.

## Proposal

- Add a "link_select_options" parameter to the documents abilities dictionary from which the frontend can propose a limited set of choices for link reach and link role.
- allow getting descendants of a document: add new "descendants" action to document detail endpoint (it's elegant to do it as a document detail action so that we benefit from the document's access control)
- allow getting the ancestors tree of a document, opened at the current document
- fix filtering on the `children` view and make sure it works as well on the new `descendants` view
- allow setting page size on the document list endpoints (up to 200 max).
    
TODO: In a future Pull Request, we should take control of the "get_object"  and apply specific filtering and annotations to it that is not coupled to what we do on other lists views...


lunika (commented):
Quite a nightmare to review

src/backend/core/api/utils.py
```diff
@@ -11,6 +11,33 @@
 from rest_framework.throttling import BaseThrottle
 
 
+def flat_to_nested(documents):
```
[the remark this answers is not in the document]
qbey: Did you consider using the same approach as the `dump_bulk` method? https://github.com/django-treebeard/django-treebeard/blob/master/treebeard/mp_tree.py#L625
I'm not saying it's better, I'm just wondering.

src/backend/core/api/viewsets.py
```diff
@@ -766,6 +768,71 @@ def children(self, request, *args, **kwargs):
...
+    )
+    def tree(self, request, *args, **kwargs):
+        """
+        List ancestors tree above the document.
+        What we need to display is the tree structure opened for the current document.
+        """
+        lookup_url_kwarg = self.lookup_url_kwarg or self.lookup_field
+        filter_kwargs = {self.lookup_field: self.kwargs[lookup_url_kwarg]}
+        current_document = (
+            self.queryset.filter(**filter_kwargs).only("depth", "path").first()
```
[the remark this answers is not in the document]
sampaccoud: I find the logic clearer by using the .first() pattern in this case.
get_queryset can't be used here as it does more. This can be improved after refactoring get_object.

src/backend/core/models.py
```diff
@@ -367,6 +367,51 @@ def _get_abilities(self, resource, user):
...
+            return self.filter(
+                models.Q(accesses__user=user)
+                | models.Q(accesses__team__in=user.teams)
+                | ~models.Q(link_reach=LinkReachChoices.RESTRICTED)
+            )
+
+        return self.filter(models.Q(link_reach=LinkReachChoices.PUBLIC))
+
+
+class DocumentManager(models.Manager):
```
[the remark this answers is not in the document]
sampaccoud: good point! thanks.

src/backend/core/models.py
```diff
@@ -367,6 +367,51 @@ def _get_abilities(self, resource, user):
         }
 
 
+class DocumentQuerySet(models.QuerySet):
```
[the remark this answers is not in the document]
sampaccoud: done.

src/backend/core/models.py
```diff
@@ -367,6 +367,51 @@ def _get_abilities(self, resource, user):
...
+    Custom manager for the Document model, enabling the use of the custom
+    queryset methods directly from the model manager.
+    """
+
+    def get_queryset(self):
+        """
+        Overrides the default get_queryset method to return a custom queryset.
+        :return: An instance of DocumentQuerySet.
+        """
+        return DocumentQuerySet(self.model, using=self._db)
```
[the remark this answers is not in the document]
sampaccoud: done

src/backend/core/models.py
```diff
@@ -367,6 +367,51 @@ def _get_abilities(self, resource, user):
...
+        :return: A queryset of documents readable by the user.
+        """
+        if user.is_authenticated:
+            return self.filter(
+                models.Q(accesses__user=user)
+                | models.Q(accesses__team__in=user.teams)
+                | ~models.Q(link_reach=LinkReachChoices.RESTRICTED)
+            )
+
+        return self.filter(models.Q(link_reach=LinkReachChoices.PUBLIC))
```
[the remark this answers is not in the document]
sampaccoud: True!

src/backend/core/models.py
```diff
@@ -367,6 +367,51 @@ def _get_abilities(self, resource, user):
...
+        Filters the queryset to return documents that the given user has
+        permission to read.
+        :param user: The user for whom readable documents are to be fetched.
+        :return: A queryset of documents readable by the user.
+        """
+        if user.is_authenticated:
+            return self.filter(
+                models.Q(accesses__user=user)
+                | models.Q(accesses__team__in=user.teams)
+                | ~models.Q(link_reach=LinkReachChoices.RESTRICTED)
```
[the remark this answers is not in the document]
sampaccoud: if you are authenticated, a document is readable if you have a specific access on it, either directly via your user or via a team to which you are related.
If you don't have a specific access, you can also access the document if it is has a link reach of PUBLIC or AUTHENTICATED... so we exclude document that have a link reach of RESTRICTED.

src/backend/core/api/utils.py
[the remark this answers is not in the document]
sampaccoud: No I didn't know it existed but I think it's much simpler and more efficient to serialize the documents flat and nest them in the last step.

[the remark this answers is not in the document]
sampaccoud: Ok let's add unit tests (although the method is already battle tested via the API results).

src/backend/core/api/viewsets.py
```diff
@@ -766,6 +768,71 @@ def children(self, request, *args, **kwargs):
...
+        )
+
+        if current_document is None:
+            raise Http404
```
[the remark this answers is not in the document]
lunika: While it's not refator you can probably use it. But you can do the same thing without checking the lookup_filed, etc. Something like this : 

```
@drf.decorators.action(detail=True, methods=["get"])
    def tree(self, request, pk=None):

        document = drf.generics.get_object_or_404(self.queryset, pk=pk)
```

src/backend/core/models.py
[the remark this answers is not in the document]
lunika: Yes, I understood later doing my stuff in drive :+1:

src/backend/core/api/viewsets.py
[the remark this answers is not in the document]
sampaccoud: but that would pull all the fields when I need only "depth" and "path"? Among the unwanted fields is "excerpt".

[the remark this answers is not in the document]
lunika: Yes but you can still remove the manipulation around lookup_field and lookup_url_kwargs. You can directly use the `pk` argument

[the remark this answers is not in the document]
lunika: and you can't do `self.queryset.only("depth", "path").get(pk=pk)` ? With a try/except block ?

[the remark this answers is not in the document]
lunika:
```
try:
            document = self.queryset.only("path").get(pk=pk)
        except models.Document.DoesNotExist:
            raise drf.exceptions.NotFound
```

src/backend/core/tests/test_models_documents.py
```diff
@@ -927,29 +927,78 @@ def test_models_documents_nb_accesses_cache_is_invalidated_on_document_soft_dele
...
+    instance "deleted_at" attributes where tempered with.
+    """
+    document = factories.DocumentFactory()
+
+    if random.choice([False, True]):
+        document.deleted_at = timezone.now()
+    else:
+        document.ancestors_deleted_at = timezone.now()
+
+    with pytest.raises(RuntimeError, match="This document is not deleted."):
```
lunika, line 1000: Nice, I didn't know about this `match` parameter :+1:

src/backend/core/tests/test_api_utils_flat_to_nested.py
```diff
@@ -0,0 +1,109 @@
...
+
+def test_api_utils_flat_to_nested_single_document():
+    """Test that a single document is returned as the only root element."""
+    documents = [{"id": "1", "depth": 0}]
+    expected = [{"id": "1", "depth": 0, "children": []}]
+    assert flat_to_nested(documents) == expected
+
+
+def test_api_utils_flat_to_nested_multiple_root_documents():
+    """Test that multiple root-level documents are correctly added to the root."""
```
lunika, line 20: You can't have two roots element. IMO an exception should be raised if it's the case.

src/backend/core/models.py
```diff
@@ -867,32 +867,33 @@ def send_invitation_email(self, email, role, sender, language=None):
...
-                "Cannot delete this document because one or more ancestors are already deleted."
+                _("This document is already deleted or has deleted ancestors.")
             )
 
         self.ancestors_deleted_at = self.deleted_at = timezone.now()
         self.save()
         self.invalidate_nb_accesses_cache()
 
+        if parent := self.get_parent():
+            self._meta.model.objects.filter(pk=parent.pk).update(
```
lunika, line 893: Why using `self._meta.model.objects` is better than using `Document.objects` ?

src/backend/core/models.py
```diff
@@ -867,32 +867,33 @@ def send_invitation_email(self, email, role, sender, language=None):
...
             raise RuntimeError(
-                "Cannot delete this document because one or more ancestors are already deleted."
+                _("This document is already deleted or has deleted ancestors.")
             )
 
         self.ancestors_deleted_at = self.deleted_at = timezone.now()
         self.save()
         self.invalidate_nb_accesses_cache()
 
+        if parent := self.get_parent():
```
lunika, line 892: You can probably save one query here. if `self` has a parent it means that `self.depth > 0`:

```suggestion
       if self.depth > 0:
            parent = self.get_parent()
            ...
```

WDYT ?

src/backend/core/models.py
```diff
@@ -929,8 +926,13 @@ def restore(self):
...
```
lunika, line 929: Same here.

src/backend/core/api/viewsets.py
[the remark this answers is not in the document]
sampaccoud: ok. Not sure if the try/except is better than the .first pattern but ok.

src/backend/core/tests/test_api_utils_flat_to_nested.py
sampaccoud: yep I modified the method as discussed irl

src/backend/core/models.py
sampaccoud: I like that it can be reached from the args that the method gets and not from something external. Sometimes importing models globally leads to circular imports so I tend to limit them when I can.

src/backend/core/models.py
sampaccoud: ok

src/backend/core/models.py
sampaccoud: done!

src/backend/core/models.py
```diff
@@ -750,19 +873,26 @@ def soft_delete(self):
...
+            self._meta.model.objects.filter(
+                models.Q(deleted_at__isnull=False)
+                | models.Q(ancestors_deleted_at__isnull=False),
+                pk=self.pk,
+            ).exists()
+            or self.get_ancestors().filter(deleted_at__isnull=False).exists()
+        ):
             raise RuntimeError(
-                "Cannot delete this document because one or more ancestors are already deleted."
+                _("This document is already deleted or has deleted ancestors.")
```
qbey, line 885: Is this displayed/user in the frontend? I'm not sure de should translate errors, because it makes "grep" in logs quite complex.

src/backend/core/models.py
```diff
@@ -773,18 +903,14 @@ def soft_delete(self):
     def restore(self):
         """Cancelling a soft delete with checks."""
         # This should not happen
-        if self.deleted_at is None:
-            raise ValidationError({"deleted_at": [_("This document is not deleted.")]})
+        if self._meta.model.objects.filter(
+            pk=self.pk, deleted_at__isnull=True
+        ).exists():
+            raise RuntimeError(_("This document is not deleted."))
```
qbey, line 909: Same comment as above

src/backend/core/models.py
```diff
@@ -773,18 +903,14 @@ def soft_delete(self):
...
-            raise ValidationError(
-                {
-                    "deleted_at": [
-                        _(
-                            "This document was permanently deleted and cannot be restored."
-                        )
-                    ]
-                }
+            raise RuntimeError(
+                _("This document was permanently deleted and cannot be restored.")
```
qbey, line 913: Same here

qbey (approved):
Nice :) 

I only have a question about translated text in errors.

src/backend/core/api/viewsets.py
```diff
@@ -759,12 +777,102 @@ def children(self, request, *args, **kwargs):
...
+            raise drf.exceptions.NotFound from excpt
+
+        ancestors = (
+            (current_document.get_ancestors() | self.queryset.filter(pk=pk))
+            .filter(ancestors_deleted_at__isnull=True)
+            .order_by("path")
+        )
+
+        # Get the highest readable ancestor
+        highest_readable = ancestors.readable_per_se(request.user).only("depth").first()
```
lunika, line 829:
```suggestion
        highest_readable = ancestors.readable_per_se(request.user).only("id", "depth").first()
```

src/backend/core/api/viewsets.py
```diff
@@ -759,12 +777,102 @@ def children(self, request, *args, **kwargs):
...
+        paths_links_mapping = {}
+        ancestors_links = []
+        children_clause = db.Q()
+        for ancestor in ancestors:
+            if ancestor.depth < highest_readable.depth:
+                continue
+
+            children_clause |= db.Q(
+                path__startswith=ancestor.path, depth=ancestor.depth + 1
+            )
```
lunika, line 846:
```suggestion
            if ancestor.path == highest_readable.path:
                children_clause |= db.Q(pk=highest_readable.id)
            else:
                children_clause |= db.Q(
                    path__startswith=ancestor.path, depth=ancestor.depth + 1
                )
```

lunika: When you reach the highest readable you are at the root of your tree. In that case you only want to pick it but not its siblings.

src/backend/core/api/viewsets.py
sampaccoud: From what I tested, "id" is always included by default. Do you have a reason to believe it may have to be added? Or do you just think explicit is better than implicit?

src/backend/core/api/viewsets.py
sampaccoud: but this is not picking the siblings, this is picking the children. According to the tests I think we are already good?

src/backend/core/models.py
sampaccoud: agreed! These types of messages should not be translated :pray:

src/backend/core/api/viewsets.py
lunika: No I just didn't know that it was included by default, probably because it is the primary key.

src/backend/core/api/viewsets.py
lunika: You are right, you select the highest later line 857. You can ignore this suggestion

src/backend/core/api/viewsets.py
```diff
@@ -499,8 +498,38 @@ def get_queryset(self):
...
+        """Return paginated response for the queryset if requested."""
+        page = self.paginate_queryset(queryset)
+        if page is not None:
+            serializer = self.get_serializer(page, many=True)
+            return self.get_paginated_response(serializer.data)
+
+        serializer = self.get_serializer(queryset, many=True)
+        return drf.response.Response(serializer.data)
+
+    def list(self, request, *args, **kwargs):
```
lunika, line 517: For me this PR is ok. 
But @PanchoutNathan has spot an unwanted behavior. The list view should only return highest readable documents. And with this implementation it is not the case. Here a test showing the expected behavior:

```
def test_api_documents_list_filter_title_should_not_returned_children():
    """Authenticated users should be able to search documents by their title but not returning children."""
    client = APIClient()
    client.force_login(user)

    # Create documents with predefined titles
    titles = [
        "Project Alpha Documentation",
        "Project Beta Overview",
        "User Guide",
        "Financial Report 2024",
        "Annual Review 2024",
    ]
    parent = factories.DocumentFactory(title="Parent", users=[user])
    for title in titles:
        factories.DocumentFactory(title=title, users=[user], parent=parent)

    # Perform the search query
    response = client.get(f"/api/v1.0/documents/?title=Alpha")

    assert response.status_code == 200
    results = response.json()["count"]
    assert results == 0
```

**But** I don't really understand why they should not be returned. We should discuss about the expected behavior.

lunika (approved):

src/backend/core/api/viewsets.py
```diff
@@ -499,8 +498,38 @@ def get_queryset(self):
...
+        applied before ordering and returning the response.
+        """
+        queryset = (
+            self.get_queryset()
+        )  # Not calling filter_queryset. We do our own cooking.
+
+        filterset = ListDocumentFilter(
             self.request.GET, queryset=queryset, request=self.request
         )
         filterset.is_valid()
```
lunika, line 535:
```suggestion
        if not filterset.is_valid():
            raise drf.exceptions.ValidationError(filterset.errors)
```

the `is_valid` method only returns a boolean. It doesn't raises an exception like a serializer.

PR #648, opened 2025-02-19
♻️(compose) set compose name in the docker-compose file

lunika:
## Purpose

The helper bin/compose was using the option -p to set the compose project name but this option is not used in the Makefile. This can lead to different way to use the docker compose file definition with different project name. In order to have a consistent name everywhere and for everybody, we set the name in the docker compose file itself.

## Proposal

- [x] set compose name in the docker-compose file


AntoLC (approved):
Nice, thanks

PR #649, opened 2025-02-20
📈(frontend) abstract analytics classes

AntoLC:
## Purpose

Add abstract classes for analytics services.
We will be able to add easily any analytic services.
Our first analytic service use case is Posthog.

## Proposal

- [x] 📈(frontend) abstract analytics classes
- [x] 📈(frontend) get user analytics
- [x] 🚩(frontend) add feature flag on "Copy as HTML"

## Demo

https://github.com/suitenumerique/docs/blob/09cd10222b4992789a9a159b9bfb5e3f6d650171/src/frontend/apps/impress/src/features/docs/doc-header/components/DocToolBox.tsx#L91-L98

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to
 
 ## Added
 
+- 🚩(frontend) feature flag analytic on copy as html #649
 - ✨(frontend) Custom block divider with export #698
 
 ## Changed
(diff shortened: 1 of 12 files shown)


virgile-dev (approved):
Sounds great to me. If the feature flag is enabled users matching the conditions will see the button, if not they won't see it right ? @AntoLC

AntoLC:
> Sounds great to me. If the feature flag is enabled users matching the conditions will see the button, if not they won't see it right ? @AntoLC

Yes but with nuances: 
- if the feature flag is disabled, the button will be **shown**
- if the feature flag is enabled and send true, the button will be **shown**
- if the feature flag is enabled and send false, the button will be **hidden**

PR #660, opened 2025-02-24
♻️(back) extract ancestor deleted_at directly from db in restore method

lunika:
## Purpose

In the restore method, all the ancestors with a deleted_at date set are extracted from the database and then the oldest value is extracted using the min python function. This usage of min can be removed by sorting directly the deleted_at at the database level and then fetching the first one. It's faster and easier to maintain.

## Proposal

- [x] extract ancestor deleted_at directly from db in restore method

### DIFF
--- a/src/backend/core/models.py
+++ b/src/backend/core/models.py
@@ -794,9 +794,11 @@ class Document(MP_Node, BaseModel):
         ancestors_deleted_at = (
             self.get_ancestors()
             .filter(deleted_at__isnull=False)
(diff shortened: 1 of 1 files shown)


qbey (approved):
👌should we add an index on deleted_at?

sampaccoud:
> 👌should we add an index on deleted_at?

ancestors__deleted_at is probably even more used than deleted_at for filtering.

sampaccoud (approved):

PR #663, opened 2025-02-25
🔒️Unsafe attachments

AntoLC:
## Purpose

We improve the security of unsafe files.

## Proposal

We decided to sanitize markup files.
We will prevent users to open files from a new tab, the files will be directly downloaded.
We alert the users when they are about to download unsafe files with a confirmation modal.
We could implement other security layer in future development inside this confirmation modal, like proposing [jecliqueoupas](https://jecliqueoupas.cyber.gouv.fr/documentation).

- [x] 🚸(frontend) block click on unsafe image
- [x] 🛂(frontend) secure download button
- [x] 🔒️(back) set ContentDisposition on media upload
- [x] 🔒️(nginx) manage Content-Security-Policy in nginx config
- [x] 🛂(backend) remove svg from unsafe

## Demo

image


src/backend/core/utils.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def sanitize_markup(markup: str, forbidden_tags=None) -> str:
+    """Remove forbidden tags and their contents from the given markup."""
+    if forbidden_tags is None:
+        forbidden_tags = FORBIDDEN_TAGS
+
+    soup = BeautifulSoup(markup, "xml")
+
+    forbidden_lower = {tag.lower() for tag in forbidden_tags}
```
lunika, line 15: Useless. You can remove it.

src/backend/core/utils.py
```diff
@@ -0,0 +1,19 @@
...
+    for tag in soup.find_all():
+        if tag.name.lower() in forbidden_lower:
+            tag.decompose()
+    return str(soup)
```
lunika, line 19:
```
for tag in soup.find_all(FORBIDDEN_TAGS):
    tag.decompose()

return str(soup)
```

src/backend/core/api/serializers.py
```diff
@@ -426,8 +428,29 @@ def validate(self, attrs):
...
+            "application/xml",
+        ]:
+            raw_content = file_obj.read()
+            decoded_str = raw_content.decode("utf-8", errors="replace")
+            sanitized_str = sanitize_markup(decoded_str)
+            sanitized_bytes = sanitized_str.encode("utf-8")
+            file_obj = ContentFile(sanitized_bytes, name=file_obj.name)
+
+        file_obj.seek(0)
+        return file_obj
```
lunika, line 454: As it I don't like it. It will become a bottleneck if the file is large. You should test if beautifulsoup accepts in its constructor the `file_obj` directly, if it has the same signature as a python file object then it should handle it in a better way than using `read`

src/backend/core/utils.py
AntoLC: As talked, commit removed.

src/backend/core/utils.py
AntoLC: As talked, commit removed.

src/backend/core/api/serializers.py
AntoLC: As talked, commit removed.

lunika (approved):

PR #671, opened 2025-03-01
🐛(backend) allow more extensions for media download

sampaccoud:
## Purpose

A user reported that mp4 video were not working. 
The request to download the video is getting a 403 whereas images work on the same document :thinking: 

## Proposal

The regex to validate media file extensions was too restrictive. Make it more permissive to allow any type of file extensions.


AntoLC (approved):
Surprising we didn't see that before.

PR #678, opened 2025-03-04
♻️(front) use sentry tags instead of extra scope

lunika:
## Purpose

To ease filtering issues on sentry, we want to use tags instead of extra scope. Tags are indexed and searchable, it's not the case with extra scope. Moreover using setExtra to add additional data is deprecated.

## Proposal

- [x] use sentry tags instead of extra scope

### DIFF
--- a/src/frontend/apps/impress/src/stores/useSentryStore.tsx
+++ b/src/frontend/apps/impress/src/stores/useSentryStore.tsx
@@ -12,21 +12,21 @@ interface SentryState {
 export const useSentryStore = create<SentryState>((set, get) => ({
   sentry: undefined,
   setSentry: (dsn, environment) => {
-    const sentry = get().sentry;
-    if (sentry) {
+    if (get().sentry) {
       return;
     }
 
-    set({
-      sentry: Sentry.init({
-        dsn,
-        environment,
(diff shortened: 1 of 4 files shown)


src/frontend/apps/impress/src/stores/useSentryStore.tsx
```diff
@@ -12,21 +12,21 @@ interface SentryState {
...
+    const sentry = Sentry.init({
+      dsn,
+      environment,
+      integrations: [Sentry.replayIntegration()],
+      release: packageJson.version,
+      replaysSessionSampleRate: 0.1,
+      replaysOnErrorSampleRate: 1.0,
+      tracesSampleRate: 0.1,
+    })
+    Sentry.setTag('service', 'frontend');
```
lunika, line 28: Haha fucking copilot replacing `application` by `service`

qbey (approved):
Thanks :)

qbey:
Note: While you are looking at it, did you consider adding

´´´
# Ignore the logs added by the DockerflowMiddleware
            ignore_logger("request.summary")
´´´

We added it on people because sentry errors on the backend were duplicated.

PR #684, opened 2025-03-06
🐛(frontend) Collaboration upgrade

AntoLC:
## Purpose

Last version of Blocknote (0.23.2) has a bug with the collaboration.
We upgrade Blocknote to `0.23.2-hotfix.0` that fix this problem.

We got some compatibility problems as well because of version not pinned in blocknote.
We pinned the problematic dependencies to fix these problems.

## Proposal

- [x] 🐛(frontend) fix collaboration error
- [x] 📌(frontend) pin yjs globally
- [x] 📌(frontend) pin blocknote globally
- [x] 📌(frontend) pin @react-pdf/renderer globally
- [x] 📌(frontend) pin docx globally
- [x] 📌(frontend) pin prosemirror-view globally


CHANGELOG.md
```diff
@@ -422,7 +425,8 @@ and this project adheres to
 - ✨(frontend) Coming Soon page (#67)
 - 🚀 Impress, project to manage your documents easily and collaboratively.
 
-[unreleased]: https://github.com/numerique-gouv/impress/compare/v2.3.0...main
+[unreleased]: https://github.com/numerique-gouv/impress/compare/v2.4.0...main
+[v2.4.0]: https://github.com/numerique-gouv/impress/releases/v2.4.0
```
lunika, line 429: You should revert this

AntoLC: Ah yes

src/frontend/package.json
```diff
@@ -40,6 +40,7 @@
     "@typescript-eslint/parser": "8.23.0",
     "docx": "9.1.1",
     "eslint": "8.57.0",
+    "prosemirror-view": "1.37.2",
```
lunika, line 43: Do we have to set a rule in renovate configuration ? Like ignoring it because we want to manage the version our self ?

AntoLC: We could but Blocknote made a fix for these 2 already:
- https://github.com/TypeCellOS/BlockNote/pull/1494
- https://github.com/TypeCellOS/BlockNote/pull/1431
Docx seems to work with newer version as well.

I have already a PR opened about bumping Blocknote, I think it will be ready before renovate tried to bump them.

lunika (approved):

PR #690, opened 2025-03-06
🔒️(back) restrict access to favorite_list endpoint

lunika:
## Purpose

favorite_list endpoint is accessible to anonymous user. This lead to an
error 500. This endpoint should be accessible only to authenticated
users.

## Proposal

- [x] 🔒️(back) restrict access to favorite_list endpoint
- [x]  🐛(back) fix ValidationError exception handler


sampaccoud (approved):
Nice catch! Clean fix 🙏🙏🙏

PR #695, opened 2025-03-10
♻️(frontend) replace cors proxy for export

AntoLC:
## Purpose

We were using the cors proxy of Blocknote.js to export the document.
Now we use our own proxy to avoid CORS issues.

## Proposal

- [x] ✨(back) create a cors proxy fetching docs external resources 
- [x] ♻️(frontend) replace cors proxy for export


src/backend/core/api/viewsets.py
```diff
@@ -1237,6 +1239,70 @@ def ai_translate(self, request, *args, **kwargs):
...
+            )
+
+        url = unquote(url)
+
+        decoded_bytes = base64.b64decode(doc.content)
+        if not url.encode("utf-8") in decoded_bytes:
+            return drf.response.Response(
+                {"detail": "URL not found in document content"},
+                status=drf.status.HTTP_404_NOT_FOUND,
+            )
```
AntoLC, line 1275: This part actually can create problems.
3 kind of problems:
- the front save the doc every minutes (if a diff), if the image is set and does not have time to be saved and the user want to export, the endpoint will fails
- we are able to add template, the template is not part of the document content, we can have images in the template
- soon we will have the header / footer feature, same problem, they are not part of the `doc.content` but can have images.

src/backend/core/tests/documents/test_api_documents_cors_proxy.py
```diff
@@ -0,0 +1,156 @@
...
+
+def test_api_docs_cors_proxy_valid_url(base64_doc_content):
+    """Test the CORS proxy API for documents with a valid URL."""
+    document = factories.DocumentFactory(link_reach="public")
+    document.content = base64_doc_content
+    document.save()
+
+    client = APIClient()
+    url_to_fetch = (
+        "https://upload.wikimedia.org/wikipedia/commons/3/31/Wiki_logo_Nupedia.jpg"
```
AntoLC, line 45: It seems to be flaky and send a 403, maybe we should use a more reliable one ? 
What about that: https://docs.numerique.gouv.fr/assets/logo-gouv.png

lunika (approved):

PR #699, opened 2025-03-11
⚡️(back) use redis as session backend in developement

lunika:
## Purpose

We want to persist the session during development. Otherwise the session is reset everytime the server is restart. This behavior make developing both a front and back feature a nigthmare, we spend our time login again and again

## Proposal

- [x] ⚡️(back) use redis as session backend in developement


qbey (approved):
Nice :) the default override is a good idea too.

@sdemagny for people
@lebaudantoine for meet

lunika:
The redis databaseis also set to `2` in order to keep the database `1` available for an other purpose and/or later use.

PR #706, opened 2025-03-12
🐛(frontend) SVG export

AntoLC:
## Purpose

The SVG were not displayed in the export (PDF / Docx).

## Proposal

We convert the SVG to PNG before exporting. 
Images could be out of the container, we added a max width.

- [x] 🐛(frontend) fix svg not rendering export pdf
- [x] 🐛(frontend) fix svg not rendering export dox


NathanVss (approved):
So sad I have nothing to say, well done ⛴️😁

PR #725, opened 2025-03-15
🐛(backend) compute ancestor_links in get_abilities if needed

lunika:
## Purpose

The refactor made in the tree view caching the ancestors_links to not compute them again in the document.get_abilities method lead to a bug. If the get_abilities method is called without ancestors_links, then they are computes on all the ancestors but not from the highest readable ancestor for the current user. We have to compute them with this constraints and to avoid to compute it again and again we save the result in a cache.

## Proposal

- [x] compute ancestor_links in get_abilities if needed

### DIFF
--- a/src/backend/core/models.py
+++ b/src/backend/core/models.py
@@ -730,6 +730,42 @@ class Document(MP_Node, BaseModel):
 
         return dict(links_definitions)  # Convert defaultdict back to a normal dict
 
+    def compute_ancestors_links(self, user):
+        """
+        Compute the ancestors links for the current document up to the highest readable ancestor.
+        """
+        cache_key = f"ancestors_links_{self.id!s}_{user.id!s}"
+        ancestors_links = cache.get(cache_key)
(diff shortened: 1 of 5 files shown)


src/backend/core/models.py
```diff
@@ -730,14 +730,50 @@ def get_links_definitions(self, ancestors_links):
...
+                continue
+
+            ancestors_links.append(
+                {"link_reach": ancestor.link_reach, "link_role": ancestor.link_role}
+            )
+            paths_links_mapping[ancestor.path] = ancestors_links.copy()
+
+        ancestors_links = paths_links_mapping.get(self.path[: -self.steplen], [])
+
+        cache.set(cache_key, ancestors_links)
```
sampaccoud, line 765: the problem with this cache is you didn't handle its expiration cases?

qbey (approved):
I think there is a way to reduce the amount of queries for the retrieve endpoint, but it would require much time.

PR #736, opened 2025-03-17
🚨(helm) fix helmfile lint

lunika:
## Purpose

Latest release of helmfile is applying the change related before as a warning. Environnements must be before releases but not in the same document of repositories.

## Proposal

- [x] fix helmfile lint


AntoLC (approved):

rouja (approved):
LGTM

PR #742, opened 2025-03-17
🌐(i18n) activate dutch language

lunika:
## Purpose

All the dutch translations are complete on crowdin. We activate it in the django settings and download all translations from crowdin

## Proposal

- [x] 🌐(i18n) activate dutch language


virgile-dev (approved):
Sounds good !

src/backend/core/migrations/0020_alter_user_language.py
```diff
@@ -0,0 +1,25 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("core", "0019_alter_user_language_default_to_null"),
+    ]
+
+    operations = [
+        migrations.AlterField(
```
qbey, line 12: This is noop, I would not make a migration for this

lunika: Thanks

AntoLC (approved):
🚀

rvveber:
You forgot to add a migration.
Everytime a model is modified, even if it's only adding an option, we need to create a migration.

Issue #731, opened 2025-03-17
docker-compose not working ??


changchichung:
running 
```
 make bootstrap FLUSH_ARGS='--no-input'
```
After the program starts, although I can open the URL, I can only see an icon spinning in circles continuously.

![](https://i.excen.one/u/rcVNGF.png)

and docker logs shows
```
kc_postgresql-1  | 2025-03-17 03:16:06.670 UTC [13343] FATAL:  role "root" does not exist
kc_postgresql-1  | 2025-03-17 03:16:07.738 UTC [13351] FATAL:  role "root" does not exist
kc_postgresql-1  | 2025-03-17 03:16:08.802 UTC [13359] FATAL:  role "root" does not exist
kc_postgresql-1  | 2025-03-17 03:16:09.868 UTC [13367] FATAL:  role "root" does not exist
kc_postgresql-1  | 2025-03-17 03:16:10.931 UTC [13375] FATAL:  role "root" does not exist
kc_postgresql-1  | 2025-03-17 03:16:11.996 UTC [13383] FATAL:  role "root" does not exist
kc_postgresql-1  | 2025-03-17 03:16:13.060 UTC [13392] FATAL:  role "root" does not exist
kc_postgresql-1  | 2025-03-17 03:16:14.127 UTC [13400] FATAL:  role "root" does not exist
kc_postgresql-1  | 2025-03-17 03:16:15.190 UTC [13408] FATAL:  role "root" does not exist
kc_postgresql-1  | 2025-03-17 03:16:16.253 UTC [13416] FATAL:  role "root" does not exist
kc_postgresql-1  | 2025-03-17 03:16:17.323 UTC [13424] FATAL:  role "root" does not exist
kc_postgresql-1  | 2025-03-17 03:16:18.387 UTC [13432] FATAL:  role "root" does not exist
kc_postgresql-1  | 2025-03-17 03:16:19.449 UTC [13440] FATAL:  role "root" does not exist
kc_postgresql-1  | 2025-03-17 03:16:20.516 UTC [13448] FATAL:  role "root" does not exist
kc_postgresql-1  | 2025-03-17 03:16:21.585 UTC [13456] FATAL:  role "root" does not exist
```

lunika:
Hi,

Thank you for your issue. Can you provide more information about the environement you are working on ?

Docker version ?
Docker compose version ?
What Operating System are you using ?

etc.

cameroncking:
I'm experiencing the same issue.

```
$ git status
[...]

PR #748, opened 2025-03-17
:bug:(backend) fix postgres healthcheck

cameroncking:
## Purpose

Today, the docker healthcheck does not include the postgreql user or database name.  This causes errors such as the following:

```
kc_postgresql-1  | 2025-03-17 03:16:17.323 UTC [13424] FATAL:  role "root" does not exist
```

## Proposal

Update the `pg_isready` command to specify the username and password such as `pg_isready -d $${POSTGRES_DB} -U $${POSTGRES_USER}`.

See also:
- https://github.com/suitenumerique/docs/issues/731 for an example of this log message being reported.  (however, it may not be the root cause of the issue described.)
- https://github.com/peter-evans/docker-compose-healthcheck/issues/16 for a similar report in another repo, fixed the same way

# Test

After applying this fix, the error message mentioned above no longer appears in `docker compose logs -f`.

```
$ make test
...
1124 passed, 9579 warnings in 18.60s
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,7 +4,7 @@ services:
   postgresql:
     image: postgres:16
     healthcheck:
-      test: ["CMD-SHELL", "pg_isready"]
+      test: ["CMD-SHELL", "pg_isready -d $${POSTGRES_DB} -U $${POSTGRES_USER}"]
       interval: 1s
       timeout: 2s
       retries: 300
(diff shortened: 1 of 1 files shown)


lunika:
Hi @cameroncking ,

Thank you very much for your contribution, we appreciate it. And it was something we planned to fix so your contribution is more than welcome.

I let you squash your commit and rename it following our convention https://suitenumerique.gitbook.io/handbook/git#type-scope-and-subject

cameroncking:
I believe I have updated the commit message correctly.  Thank you for the quick review!

lunika:
Almost done, The last line in your commit body is too long: 

> 4: B1 Line exceeds max length (81>80): "This resolves an error that appears in the logs, where 'root' is used by default."

Going to a new line after the word `by` will fix the problem.

lunika (approved):

PR #759, opened 2025-03-18
🚚(frontend) redirect to 401 page when 401 error

AntoLC:
## Purpose

 - Users could still be able to edit a document if the session was expired. It could give the feeling that the document was not saved.
 - `useAnalytics` was unmounting children because of function not correctly dispatched.

## Proposal

If during a mutation request (POST, PUT, DELETE), the server returns a 401 error, the user is redirected to the 401 page.

- [x] 🚚(frontend) redirect to 401 page when 401 error
- [x] 🐛(frontend) unmount components Analytics

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ and this project adheres to
     and descendants views #695
 - 🐛(action) fix notify-argocd workflow #713
 - 🚨(helm) fix helmfile lint #736
+- 🚚(frontend) redirect to 401 page when 401 error #759
 
 
 ## [2.4.0] - 2025-03-06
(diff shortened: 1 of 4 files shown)


src/frontend/apps/impress/src/features/auth/api/useAuthQuery.tsx
```diff
@@ -34,6 +34,7 @@ export function useAuthQuery(
     queryKey: [KEY_AUTH],
     queryFn: getMe,
     staleTime: 1000 * 60 * 15, // 15 minutes
+    retry: 1,
```
PanchoutNathan, line 37: why ?

PR #771, opened 2025-03-19
✨(frontend) add custom css style

AntoLC:
# Runtime Theming 🎨

### Introduction

We are excited to introduce a new feature that allows you to easily change the theme of our frontend application at **runtime** ⏱️. This feature provides a flexible way to customize the look and feel of our application without requiring any code changes.

### Overwriting the Theme with Custom CSS

Our application now supports a new setting, `FRONTEND_CSS_URL` 🔗, which allows you to add a custom CSS layer on top of our default theme. This setting can be used to override our default styles and apply your own custom theme.

### How to Use

To use this feature, simply set the `FRONTEND_CSS_URL` environment variable to the URL of your custom CSS file. For example:

```javascript
FRONTEND_CSS_URL=http://anything/custom-style.css
```

Once you've set this variable, our application will load your custom CSS file and apply the styles to our frontend application.

### Benefits

This feature provides several benefits, including:

*   **Easy customization** 🔄: With this feature, you can easily customize the look and feel of our application without requiring any code changes.
*   **Flexibility** 🌈: You can use any CSS styles you like to create a custom theme that meets your needs.
*   **Runtime theming** ⏱️: This feature allows you to change the theme of our application at runtime, without requiring a restart or recompilation.

### Example Use Case

Let's say you want to change the background color of our application to a custom color. You can create a custom CSS file with the following contents:

```css
body {
  background-color: #3498db;
}
```

Then, set the `FRONTEND_CSS_URL` environment variable to the URL of your custom CSS file. Once you've done this, our application will load your custom CSS file and apply the styles, changing the background color to the custom color you specified.

### Conclusion

Our new runtime theming feature provides a flexible and easy way to customize the look and feel of our frontend application 🎉.   
By setting the `FRONTEND_CSS_URL` environment variable, you can add a custom CSS layer on top of our default theme and apply your own custom styles.   
We hope this feature helps you to create a customized experience that meets your needs.

----

## Demo

Just by including a custom css, you can totally change the look and feel of your application:
https://antolc.github.io/docs-customized/docs-custom/public/custom-style.css

image

----

image

### DIFF
--- a/src/frontend/apps/impress/src/core/config/ConfigProvider.tsx
+++ b/src/frontend/apps/impress/src/core/config/ConfigProvider.tsx
@@ -1,4 +1,5 @@
 import { Loader } from '@openfun/cunningham-react';
+import Head from 'next/head';
 import { PropsWithChildren, useEffect } from 'react';
 
 import { Box } from '@/components';
@@ -54,10 +55,17 @@ export const ConfigProvider = ({ children }: PropsWithChildren) => {
   }
 
   return (
-    <AnalyticsProvider>
-      <CrispProvider websiteId={conf?.CRISP_WEBSITE_ID}>
-        {children}
-      </CrispProvider>
-    </AnalyticsProvider>
+    <>
+      {conf?.FRONTEND_CSS_URL && (
+        <Head>
+          <link rel="stylesheet" href={conf?.FRONTEND_CSS_URL} />
+        </Head>
+      )}
+      <AnalyticsProvider>
+        <CrispProvider websiteId={conf?.CRISP_WEBSITE_ID}>
+          {children}
+        </CrispProvider>
+      </AnalyticsProvider>
+    </>
   );
 };
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to
 ## Added
 
 - 📄(legal) Require contributors to sign a DCO #779
+- ✨(frontend) add custom css style #771
 
 ## Changed
 
--- a/docs/theming.md
+++ b/docs/theming.md
@@ -1,0 +1,33 @@
+# Runtime Theming 🎨
+
+### How to Use
+
+To use this feature, simply set the `FRONTEND_CSS_URL` environment variable to the URL of your custom CSS file. For example:
+
+```javascript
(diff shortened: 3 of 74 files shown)


lunika (approved):

src/frontend/apps/impress/src/core/config/ConfigProvider.tsx
```diff
@@ -54,10 +55,17 @@ export const ConfigProvider = ({ children }: PropsWithChildren) => {
...
+      {conf?.FRONTEND_CSS_URL && (
+        <Head>
+          <link rel="stylesheet" href={conf?.FRONTEND_CSS_URL} />
+        </Head>
+      )}
+      <AnalyticsProvider>
+        <CrispProvider websiteId={conf?.CRISP_WEBSITE_ID}>
+          {children}
+        </CrispProvider>
+      </AnalyticsProvider>
```
NathanVss, line 68: Maybe not in the approriate commit

NathanVss (approved):

src/frontend/apps/impress/src/core/config/ConfigProvider.tsx
AntoLC: The commit is made for this exact piece of code actually.

rvveber:
Could we determine classNames dynamically?
E.g. an Abstract Class and the Component would inherit from the Abstract Class.
Still with the ability to overwrite the className (eg. for backwards compatibility)

className: --docs-<component-name-snake-cased>

AntoLC:
> Could we determine classNames dynamically?
> 
> E.g. in an Abstract UI Component Class with the attribute: `className: --docs-<component-name-snake-cased>`
> 
> And every UI Component would inherit from this Abstract Base Class. Ensuring that there is always a className for every (UI) Component.
> 
> With the ability to overwrite the className (eg. for backwards compatibility on component-rename)

Seems a big refacto for something that aims to be easy. It means we will have to modify all our functional components to use Class components to get this inheritance. Class component is not really the way to do anymore, I don't think Class component supports easily `hooks`. Like you said, it could create backward issues that could be hard to find because the classes will change automatically, other developers not aware about this feature could easily create issues by renaming components.

PR #774, opened 2025-03-19
add: arm64 arch image for Docker Hub

PovilasID:
## Purpose

To let official docker image be available for ARM based CPUs without building it.

## Proposal

I edited a build script to push images for both amd64 (default) and arm64 (raspberry pis etc.) architecture CPUs. 
I built all of container on my arm64 machine without getting any build errors.


AntoLC:
Closed in favor of #1901.

PR #776, opened 2025-03-19
👷(changelog) shift changelog maintenance burden to release manager

sylvinus:
Most recent external contributions (including my own!) stumble on the changelog. IMHO, this represents an unnecessary friction for new contributors and a loss of focus for the reviewing team. 

I propose that we put the release manager in charge of maintaining the CHANGELOG.md file at each release, based on the commits being included in that specific release. To view unreleased changes, everyone can just browse the latest commit history.

### DIFF
--- a/.github/workflows/impress.yml
+++ b/.github/workflows/impress.yml
@@ -35,32 +35,6 @@ jobs:
       - name: Lint commit messages added to main
         run: ~/.local/bin/gitlint --commits origin/${{ github.event.pull_request.base.ref }}..HEAD
 
-  check-changelog:
-    runs-on: ubuntu-latest
-    if: |
-      contains(github.event.pull_request.labels.*.name, 'noChangeLog') == false &&
-      github.event_name == 'pull_request'
-    steps:
-      - name: Checkout repository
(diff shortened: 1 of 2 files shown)


sampaccoud (commented):
Agree but it's important that we keep our commit history as good as it is now.  So I propose we keep the git linters and make them informative. Then the person who merges the PR is responsible for squashing the commits and write a commit that complies whenever these tests are red.

lunika:
The checks you removed are not mandatory, you can submit a PR without modifying the changelog and if you modify it then I think it's important to keep the linter do its job.
So IMHO I will not change the CI part, we can decide to let new contributors to not modify the changelog. We can modify the contributing.md file to reflect this choice, acting that modifying it is now optional for external contribution and also indicates when it should be great to update it. Most of the recent contributions are about fixing the Readme, there is no need to update the changelog for this kind of modification. But if you refactor the way media are access then yes the changelog must be modified.

AntoLC:
As say @lunika, changelog are not mandatory. It is still a good reminder to have the linter to have easy track of the important feature or refacto.

PR #781, opened 2025-03-20
🐛(back) allow only images to be used with the cors-proxy

lunika:
## Purpose

The cors-proxy endpoint allowed to use every type of files and to execute it in the browser. We limit the scope only to images and Content-Security-Policy and Content-Disposition headers are also added to not allow script execution that can be present in a SVG file.

## Proposal

- [x] 🐛(back) allow only images to be used with the cors-proxy

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,13 @@ and this project adheres to
 
 ## [Unreleased]
 
+## Added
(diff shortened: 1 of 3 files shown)


AntoLC (approved):
Seems good, the front works as fine.

PR #797, opened 2025-03-24
♻️(documents) inherit manager from queryset

qbey:
## Purpose

During a code review, I saw we are overriding the MP_NodeManager and redefine the queryset filters:

- The MP_NodeManager sorts the queryset by `path` by default and it's not done on our side, is it on purpose?
- The fact we need to redefine `readable_per_se` as a boilerplate is surprising.

## Proposal

I suggest we use the Django mechanism to generate the manager from the queryset.

- [x] Use the `from_queryset` class method to generate the document manager.


src/backend/core/models.py
```diff
@@ -445,28 +445,15 @@ def readable_per_se(self, user):
...
-        """
-        Filters documents based on user permissions using the custom queryset.
-        :param user: The user for whom readable documents are to be fetched.
-        :return: A queryset of documents for which the user has direct access,
-            team access or link access. This will not return all the documents
-            that a user can read because it can be obtained via an ancestor.
-        """
-        return self.get_queryset().readable_per_se(user)
+        """Sets the custom queryset as the default."""
+        return self._queryset_class(self.model).order_by('path')
```
qbey, line 456: See 

``` python
class MP_NodeManager(models.Manager):
    """Custom manager for nodes in a Materialized Path tree."""

    def get_queryset(self):
        """Sets the custom queryset as the default."""
        return MP_NodeQuerySet(self.model).order_by('path')
```

lunika:
I suggest to inherit from `MP_NodeManager` directly as suggested in their documentation. Also in the library they inherit from the manager like we are doing : https://github.com/django-treebeard/django-treebeard/blob/master/treebeard/mp_tree.py#L133

If it's ok to mix the "django way" and what they made, I agree to merge this PR.

lunika (approved):

PR #801, opened 2025-03-24
🔒️(back) restrict accesss to document accesses

lunika:
## Purpose

Every user having an access to a document, no matter its role have access to the entire accesses list with all the user details. Only owner or admin should be able to have the entire list, for the other roles, they have access to the list containing only owner and administrator with less information on the username. The email and its id is removed

## Proposal

- [x] 🔒️(back) restrict accesss to document accesses

### DIFF
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -27,6 +27,26 @@ class UserSerializer(serializers.ModelSerializer):
         read_only_fields = ["id", "email", "full_name", "short_name"]
 
 
+class UserLightSerializer(UserSerializer):
+    """Serialize users with limited fields."""
+
+    id = serializers.SerializerMethodField(read_only=True)
+    email = serializers.SerializerMethodField(read_only=True)
+
+    def get_id(self, _user):
+        """Return always None. Here to have the same fields than in UserSerializer."""
+        return None
+
+    def get_email(self, _user):
+        """Return always None. Here to have the same fields than in UserSerializer."""
+        return None
+
+    class Meta:
+        model = models.User
+        fields = ["id", "email", "full_name", "short_name"]
+        read_only_fields = ["id", "email", "full_name", "short_name"]
+
+
 class BaseAccessSerializer(serializers.ModelSerializer):
     """Serialize template accesses."""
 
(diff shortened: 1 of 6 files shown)


src/backend/core/api/serializers.py
```diff
@@ -27,6 +27,26 @@ class Meta:
...
 
 
+class UserLightSerializer(serializers.ModelSerializer):
+    """Serialize users with limited fields."""
+
+    id = serializers.SerializerMethodField(read_only=True)
+    email = serializers.SerializerMethodField(read_only=True)
+
+    def get_id(self, _user):
+        """Return always None. Here to have the same fields than in UserSerializer."""
```
qbey, line 37: Don't you want to inherit from `UserSerializer` then? Or is this temporary while the frontend is not ready?

src/backend/core/api/viewsets.py
```diff
@@ -1457,6 +1452,32 @@ class DocumentAccessViewSet(
...
+
+    def get_queryset(self):
+        """Return the queryset according to the action."""
+        queryset = super().get_queryset()
+
+        if self.action == "list":
+            try:
+                document = models.Document.objects.get(pk=self.kwargs["resource_id"])
+            except models.Document.DoesNotExist:
+                return queryset.none()
```
qbey, line 1465: Isn't that weird the document is not always checked? Shouldn't we return 404 in those case (more global than just the list). 
No change required in this PR anyway (because I guess it would have too much impact).

src/backend/core/api/viewsets.py
```diff
@@ -1457,6 +1452,32 @@ class DocumentAccessViewSet(
...
+
+            roles = set(document.get_roles(self.request.user))
+            is_owner_or_admin = bool(roles.intersection(set(models.PRIVILEGED_ROLES)))
+            self.is_current_user_owner_or_admin = is_owner_or_admin
+            if not is_owner_or_admin:
+                # Return only the document owner access
+                queryset = queryset.filter(role__in=models.PRIVILEGED_ROLES)
```
qbey, line 1472: Do we really need to display those roles to readers? Should we display only the owner?

src/backend/core/api/viewsets.py
lunika: I totally agree with you. I discover this while working on this PR. IMO it has been done this way to avoid one more query to test if the Document exists or not.
I wrote a test proving that we never have a 404, we should probably do something on this subject.

src/backend/core/api/viewsets.py
lunika: I'm open on this suggestion. In a first time I limited only to owners. But as administrator have the same permissions I made the choice to also return bot owners and administrators.

src/backend/core/api/serializers.py
lunika: Done

Issue #769, opened 2025-03-19
Don't include BlockNote XL package in Docs


bzg:
BlockNote XL packages are features provided by code available either under the Affero GPL license or if the user is a sponsor. See [the note in BlockNote release v0.19.0](https://github.com/TypeCellOS/BlockNote/releases/tag/v0.19.0), where @YousefED wrote:

> To ensure we run a sustainable project, we are experimenting with XL packages. These are dual-licensed packages that are released under AGPL-3.0 licenses. You're free to use these in open source projects, but for usage in proprietary / commercial projects, a sponsorship is required.

DINUM uses two XL packages for [docs.numerique.gouv.fr](https://docs.numerique.gouv.fr), its own instance of Docs: this is made possible without forcing DINUM to publish its source code as AGPL because DINUM is a sponsor of BlockNote.

But DINUM is also including XL packages directly in Docs as an open source product *shipped for others*: see [these lines](https://github.com/suitenumerique/docs/blob/cab8ef51df79caac19f39965cecaa4c39062fe86/src/frontend/apps/impress/package.json#L22C7-L23C53).

This should be fixed by not including XL packages in Docs by default: it is the responsibility of the Docs reuser to include these packages and either become a BlockNote sponsor or adhere to the AGPL licence.

rvveber:
@bzg 
If i understand correctly, DINUM is allowed to use the APGL libraries for its own instance.
But is DINUM allowed to publish/distribute the frontend image with the mentioned APGL code to the public?
([The image is published as non-MIT](https://github.com/suitenumerique/docs/blob/main/.github/workflows/docker-hub.yml#L84))

bzg:
I have discussed this with @virgile-dev who will provide an update.

virgile-dev:
Hey @rvveber 
By default when you build and image from the code the env variable `PUBLISH_AS_MIT` is set to true.
In our CI we set it to true because that's how we use it in production so our docker image can be considered AGPL.
In the future we might publish a specific docker image that's only includes MIT code. But since BlockNote is working on licence exceptions we're waiting to see if this env variable is even needed anymore.
Cheers

bzg:
> In our CI we set it to true

You mean "to false"?

> because that's how we use it in production so our docker image can be considered AGPL.

In https://hub.docker.com/r/lasuite/impress-frontend there is no overview available: @virgile-dev do you want me to open another issue about adding information (and, in particular, licensing information) there, or shall I just reopen this one?

[...]

PR #803, opened 2025-03-25
chore(README) add a warning about legal compliance

bzg:
Until https://github.com/suitenumerique/docs/issues/769 is fixed, we need to advise reusers to either become BlockNote sponsors or not to use the XL packages.


bzg:
I update the warning wrt our discussion cc @lunika @virgile-dev. Let me know if this is acceptable for the team.

PR #804, opened 2025-03-25
🏗️(y-provider) manage auth in y-provider app

lunika:
## Purpose
The way to connect to the hocuspocus server needs to be proxified in
nginx to query a dedicated route in the django application and then
follow the request to the express server with the additionnal headers.
The auth can be done in the express server by querying the backend on
the document retrieve endpoint. If the response status code is 200, the
user has access to the document, otherwise it is not the case. Then we
can check the abilities to determine what the user can do or not.

## Proposal

 - [x] 🏗️(y-provider) manage auth in y-provider app
 - [x] 🔥(back) remove collaboration-auth endpoint

### DIFF
--- a/src/frontend/servers/y-provider/src/servers/hocusPocusServer.ts
+++ b/src/frontend/servers/y-provider/src/servers/hocusPocusServer.ts
@@ -1,4 +1,5 @@
 import { Server } from '@hocuspocus/server';
+import { validate as uuidValidate, version as uuidVersion } from 'uuid';
 
 import { fetchDocument } from '@/api/getDoc';
 import { getMe } from '@/api/getMe';
@@ -27,6 +28,12 @@ export const hocusPocusServer = Server.configure({
       return Promise.reject(new Error('Wrong room name: Unauthorized'));
     }
 
+    if (!uuidValidate(documentName) || uuidVersion(documentName) !== 4) {
+      console.error('Room name is not a valid uuid:', documentName);
+
+      return Promise.reject(new Error('Wrong room name: Unauthorized'));
+    }
+
     let can_edit = false;
 
     try {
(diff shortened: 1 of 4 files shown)


sylvinus:
I'm largely in favor of this change! it removes a hidden request that is difficult to debug, removes an API route and is overall I think the right design. It's best not to rely on headers received but let the app do the check itself actively.

src/frontend/servers/y-provider/src/servers/hocusPocusServer.ts
```diff
@@ -104,13 +40,25 @@ export const hocusPocusServer = Server.configure({
...
 
     connection.readOnly = !can_edit;
 
+    /*
+     * Unauthenticated users can be allowed to connect
+     * so we flag only authenticated users
+     */
+    try {
+      const user = await getMe(requestHeaders);
+      requestHeaders['x-user-id'] = user.id;
```
lunika, line 59: Is it possible to use the `context` instead instead of modifying the headers ?

AntoLC: Yes, done

lunika: Thanks

src/frontend/servers/y-provider/src/servers/hocusPocusServer.ts
```diff
@@ -27,6 +28,12 @@ export const hocusPocusServer = Server.configure({
       return Promise.reject(new Error('Wrong room name: Unauthorized'));
     }
 
+    if (!uuidValidate(documentName) || uuidVersion(documentName) !== 4) {
+      console.error('Romm name is not a valid uuid:', documentName);
```
AntoLC, line 32:
```
      console.error('Room name is not a valid uuid:', documentName);
```

AntoLC (approved):

PR #807, opened 2025-03-26
prevent breaking migration extracting attachements

lunika:
## Purpose

Extracting attachments can sometimes fail. The Ypy library can raise an
error when reading somewhere a null value. To avoid breaking the
migration, we catch the error.

## Proposal

- [x] ♻️(back) remove filtering on logging handler
- [x] 🐛(ci) use sha256 to sign argocd webhook call
- [x]  🐛(migration) prevent breaking migration extracting attachements


.github/workflows/docker-hub.yml
```diff
@@ -6,6 +6,7 @@ on:
   push:
     branches:
       - 'main'
+      - 'fix/migration-extract-attachments'
```
lunika, line 9: To be removed

lunika:
I think we have a problem with the y_py library. This library is unmaintained and we have trouble when loading documents. 
I will investigate time on the [pycrdt](https://github.com/jupyter-server/pycrdt) lib that seems actively maintained to see if we have the same error

PR #809, opened 2025-03-26
Replace Ypy by pycrdt

lunika:
## Purpose

Ypy is deprecated and unmaintained. We have problem with parsing
    existing documents. We replace it by pycrdt, library actively maintained
    and without the issues we have with Ypy.

## Proposal

- [x] ♻️(back) remove filtering on logging handler
- [x] 🐛(ci) use sha256 to sign argocd webhook call
- [x] ♻️(back) replace Ypy by pycrdt-


davidbrochart:
Hmm I don't see any reference to `pycrdt` in the code, am I missing something?

lunika:
> Hmm I don't see any reference to `pycrdt` in the code, am I missing something?

Yes, I just forget to push my code yesterday... It's ok now

davidbrochart:
LGTM, does it get rid of the error you had previously with Ypy?

lunika:
Yes we don't have the error on the same documents.

The error we had was: 

```
PanicException
called `Option::unwrap()` on a `None` value
```

For what I understand this is an internal rust error we can't really catch and I have no idea what is the real problem with the  yjs document.

davidbrochart:
You can set the environment variable [RUST_BACKTRACE=1](https://pyo3.rs/v0.23.3/debugging.html#getting-a-stacktrace) to get the stack trace on a `panic!`, but I wouldn't recommend using Ypy anyway, so probably not worth investigating.

.github/workflows/docker-hub.yml
```diff
@@ -6,6 +6,7 @@ on:
   push:
     branches:
       - 'main'
+      - 'fix/migration-extract-attachments'
```
lunika, line 9: To be removed

qbey (approved):

PR #829, opened 2025-03-31
Ci/trivy fails

lunika:
## Purpose

Description...

## Proposal

Description...

- [] item 1...
- [] item 2...


AntoLC (approved):
Nice, I couldn't nail where came from this error with `cross-spawn`

Issue #831, opened 2025-04-02
🧑‍💻 Consider adding a placeholder translation at the end of each language file to help prevent merge conflicts


dakshesh14:
Adding a placeholder translation value at the end of each language file could help mitigate merge conflicts. Since JSON doesn't allow trailing commas, this approach may make it easier to avoid conflicts when multiple contributors are working on translations.

PR #832, opened 2025-04-02
🧑‍💻 (frontend): add placeholder translation to mitigate merge conflicts

dakshesh14:
Added a placeholder translation entry at the end of each language file to reduce merge conflicts when multiple contributors are working on translations. This change also helps minimize unnecessary diff sizes.
Address: #831

## Purpose

Introduce a placeholder translation entry to prevent merge conflicts in translation files and reduce unnecessary diff size during collaborative work.

### DIFF
--- a/src/frontend/apps/impress/src/i18n/translations.json
+++ b/src/frontend/apps/impress/src/i18n/translations.json
@@ -210,14 +210,16 @@
       "You do not have permission to view users sharing this document or modify link settings.": "Ihnen fehlt die Berechtigung, die das Dokument teilenden Benutzer einzusehen oder die Linkeinstellungen zu ändern.",
       "Your current document will revert to this version.": "Ihr aktuelles Dokument wird auf diese Version zurückgesetzt.",
       "Your {{format}} was downloaded succesfully": "Ihr {{format}} wurde erfolgreich heruntergeladen",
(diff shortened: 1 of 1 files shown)


lunika:
Hi,
Thanks for your contribution.
All the translation process is automated and for now we didn't have any conflict.

Can you describe a bit how you work and how this lead to a conflict ?

Thank you

dakshesh14:
Hi, thanks for the clarification. I wasn’t aware that the translation process here is fully automated.

In some other projects I've contributed to, we encountered merge conflicts when multiple contributors added translations to the last line of a language file. Because JSON doesn’t allow trailing commas, adding a new entry requires modifying the previous line to append a comma, which results in unnecessary diffs and occasional merge conflicts. For reference, [this commit](https://github.com/suitenumerique/docs/pull/832/commits/845921793648c5c351fded3b76424c59964740f6#diff-cdf7a1fc1d5257f5167d81d876a05c06c6442f8a1a92f5bad5950c5bf0fd7360R213) shows a case where there's an unnecessary diff.

To prevent that, some teams add a placeholder translation at the end of the file—something I adopted in this PR.

That said, if translations here are handled exclusively by automation, this PR might not be necessary. I’m happy to close both the PR and the issue if that’s the case. Thanks again for taking the time to review!

lunika:
Hi,
Yes I can you can close it.
Thanks

PR #834, opened 2025-04-02
🔊(y-provider) improve and add logs

AntoLC:
## Purpose

We have somes entries with "No cookies". we add more logs to understand why we have this case. 
We add the datetime in front of each entries as well.


src/frontend/servers/y-provider/src/middlewares.ts
```diff
@@ -42,14 +44,17 @@ export const wsSecurity = (
...
     return;
   }
 
   const cookies = req.headers['cookie'];
   if (!cookies) {
     ws.close(4001, 'No cookies');
-    console.error('CORS policy violation: No cookies');
+    logger('CORS policy violation: No cookies');
+    logger('UA:', req.headers['user-agent']);
+    logger('IP:', req.ip);
```
lunika, line 56: Nop, don't log the user ip

lunika (approved):

PR #839, opened 2025-04-03
➕(backend) add `django-lasuite` dependency

qbey:
## Purpose

We work on a common library for our project (https://github.com/suitenumerique/django-lasuite), this is the first step on this.

## Proposal

Use the OIDC backends from the new library.

- [x] Override the `OIDCAuthenticationBackend`
- [x] Update code and tests
- [x] Set the proper version of `django-lasuite` in the `pyproject.toml` file


docs/examples/impress.values.yaml
```diff
@@ -33,8 +33,8 @@ backend:
     OIDC_RP_SIGN_ALGO: RS256
     OIDC_RP_SCOPES: "openid email"
     OIDC_VERIFY_SSL: False
-    USER_OIDC_FIELD_TO_SHORTNAME: "given_name"
-    USER_OIDC_FIELDS_TO_FULLNAME: "given_name,usual_name"
+    OIDC_USERINFO_SHORTNAME_FIELD: "given_name"
+    OIDC_USERINFO_FULLNAME_FIELDS: "given_name,usual_name"
```
lunika, line 37: I don't know how to deal with it. Renaming it directly is a breaking change. Having both makes non sense has the new settings will overwrite the legacy one.

lunika (approved):

PR #841, opened 2025-04-04
🔧(backend) configurable footer

AntoLC:
## Purpose

- We want to have different footer content per environment.
- We want to be able to not have footer as well

## Proposal

- We added `FRONTEND_FOOTER_FEATURE_ENABLED` environment variable, a feature flag to enable or not the footer.
- We added the `FRONTEND_URL_JSON_FOOTER` environment variable. It will give the possibility to generate your own footer content in the frontend.

## Structure content footer:

https://github.com/suitenumerique/docs/blob/1972ef4d07eb2be2b7b0ac901b00ac8ffc8b9d7f/src/frontend/apps/impress/public/contents/footer-demo.json

### DIFF
--- a/env.d/development/common.dist
+++ b/env.d/development/common.dist
@@ -64,3 +64,4 @@ COLLABORATION_WS_URL=ws://localhost:4444/collaboration/ws/
 
 # Frontend
 FRONTEND_THEME=default
+FRONTEND_URL_JSON_FOOTER=http://frontend:3000/contents/footer-demo.json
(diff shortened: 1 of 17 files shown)


env.d/development/common.dist
```diff
@@ -63,5 +63,6 @@ COLLABORATION_SERVER_SECRET=my-secret
 COLLABORATION_WS_URL=ws://localhost:4444/collaboration/ws/
 
 # Frontend
+FRONTEND_FOOTER_FEATURE_ENABLED=True
```
lunika, line 66: Add it line 67 to group with `FRONTEND_URL_JSON_FOOTER` ?

lunika (approved):

PR #845, opened 2025-04-06
⚗️(front) conditionaly install blocknote xl packages

lunika:
## Purpose

The blocknote xl packages are licensed under AGPL v3. We want to allow a reuser to build the project without them. For this, in a first step we don't list these packages in the dependencies but install them using a postinstall script.
This script is a node script looking in every package.json files under the apps directory, look if an `extraDependencies` key exists and if yes hten will install the listed dependencies after prompting the user if wants or not to install them.
Also an environment variable AUTO_INSTALL_EXTRA_DEPS can be used to explicitly install all the dependencies or to explicitly not install them. This will help to build an image with and without these dependencies. Moreover we also need thisin a CI context.

[Screencast From 2025-04-06 14-17-42.webm](https://github.com/user-attachments/assets/3108f1c0-8e7d-47fa-9f4b-533142764d2c)

## Proposal

- [x] ⚗️(front) conditionaly install blocknote xl packages
- [ ] rework the dockerfile to make it works
- [ ] rework the CI to make it works
- [ ] dynamically load these dependencies in the code and act if they are not present to not display the export feature.

### DIFF
--- a/src/frontend/apps/impress/package.json
+++ b/src/frontend/apps/impress/package.json
@@ -19,8 +19,6 @@
     "@blocknote/core": "0.23.2-hotfix.0",
     "@blocknote/mantine": "0.23.2-hotfix.0",
     "@blocknote/react": "0.23.2-hotfix.0",
-    "@blocknote/xl-docx-exporter": "0.23.2-hotfix.0",
-    "@blocknote/xl-pdf-exporter": "0.23.2-hotfix.0",
     "@fontsource/material-icons": "5.2.5",
     "@gouvfr-lasuite/integration": "1.0.2",
     "@gouvfr-lasuite/ui-kit": "0.1.3",
@@ -79,5 +77,9 @@
     "typescript": "*",
     "webpack": "5.98.0",
     "workbox-webpack-plugin": "7.1.0"
+  },
+  "extraDependencies": {
+    "@blocknote/xl-docx-exporter": "0.23.2-hotfix.0",
+    "@blocknote/xl-pdf-exporter": "0.23.2-hotfix.0"
   }
 }
--- a/src/frontend/bin/conditional-install.js
+++ b/src/frontend/bin/conditional-install.js
@@ -1,0 +1,139 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const path = require('path');
+const { execSync } = require('child_process');
+const readline = require('readline');
+
+// Check if AUTO_INSTALL_EXTRA_DEPS is explicitly set to false
+if (process.env.AUTO_INSTALL_EXTRA_DEPS === 'false') {
+  console.log('AUTO_INSTALL_EXTRA_DEPS is set to false, skipping script execution');
+  process.exit(0);
+}
+
+// Get the workspace root directory
+const workspaceRoot = process.cwd();
+
+// Check if AUTO_INSTALL_EXTRA_DEPS environment variable is set to bypass prompts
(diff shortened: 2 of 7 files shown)


src/frontend/bin/conditional-install.js
bzg, line 1: I suggest `install-extras.js` as a more explicit name.

src/frontend/bin/conditional-install.js
```diff
@@ -0,0 +1,139 @@
...
+        
+        // Process each dependency individually
+        for (const [pkg, version] of Object.entries(packageJson.extraDependencies)) {
+          const dependencyToInstall = `${pkg}@${version}`;
+          
+          // Prompt user if not in auto mode
+          let shouldInstall = autoMode;
+          if (!autoMode) {
+            shouldInstall = await promptUser(`
+Do you want to install ${dependencyToInstall}? (y/n):
```
bzg, line 99: I suggest this:

```
Do you want to install ${dependencyToInstall}? (y/n): 
These packages are dual-licensed by the author of BlockNote under
AGPL-3.0 or a proprietary license. If you install them, you need to
fulfill your licensing obligations.
```

src/frontend/apps/impress/package.json
```diff
@@ -79,5 +77,9 @@
     "typescript": "*",
     "webpack": "5.98.0",
     "workbox-webpack-plugin": "7.1.0"
+  },
+  "extraDependencies": {
```
bzg, line 81: I'd add this comment here:

```
// The XL packages are dual-licensed by the author of BlockNote under
// AGPL-3.0 or a proprietary license. If you install them, you need to
// fulfill your licensing obligations.
```

bzg (commented):
My comment are just suggestions, not requested changes.

src/frontend/apps/impress/package.json
```diff
@@ -79,5 +77,9 @@
...
+    "@blocknote/xl-docx-exporter": "0.23.2-hotfix.0",
+    "@blocknote/xl-pdf-exporter": "0.23.2-hotfix.0"
```
AntoLC, line 83: You don't need to do that, neither to have a `conditional-install` I think , just by setting a env var at build time and have a conditional import, with the tree shaking the packages will be excluded from the final build.

```ts
import dynamic from 'next/dynamic';

let Component = null;

if (process.env.WITH_AGPL_PACKAGES === 'true') {
  Component = dynamic(() => import('blocknote-xl-package'));
} else {
  Component = dynamic(() => import('blocknote-packag'));
}

export default function Whatever() {
  return <Component />;
}
```

Then here, before the `yarn build`:
https://github.com/suitenumerique/docs/blob/401913297e4393ab3ebd9ff086ca5ccbeee26ec5/src/frontend/Dockerfile#L42-L45
You have to add this env var:
```dockerfile
...
ARG SW_DEACTIVATED
ENV NEXT_PUBLIC_SW_DEACTIVATED=${SW_DEACTIVATED}

ARG WITH_AGPL_PACKAGES 
ENV WITH_AGPL_PACKAGES =${WITH_AGPL_PACKAGES }

RUN yarn build
...
```

----

The final code just take what is used by the app, we don't have `node_modules` anymore in the image:
https://github.com/suitenumerique/docs/blob/401913297e4393ab3ebd9ff086ca5ccbeee26ec5/src/frontend/Dockerfile#L54-L56

virgile-dev (approved):
Looks good to me! Thanks
Will it hide the UI associated with the feature when the person doing the yarn build says no ?

lunika:
Managed in #911

PR #701, opened 2025-03-11
[WIP] doc dnd + tree

PanchoutNathan:

PR #846, opened 2025-04-06
Fix link reach/role and add ancestors link access info

sampaccoud:
## Purpose
    
We are returning too many select options for the link access configuration.

## Proposal

- when the "restricted" reach is an option (key present in the returned dictionary), the possible values for link roles are now always None to make it clearer that they don't matter and no select box should be shown for roles.
- ~Never propose "restricted" as option for link reach when the ancestors already offer a public access. Indeed, restricted/editor was shown when the ancestors had public/read access. The logic was to propose editor role on more restricted reaches... but this does not make sense for restricted since the role does is not taken into account for this reach. Roles are set by each access line assign to users/teams.~
- ~Never propose "authenticated" as option for link reach when the ancestors already offer a public access whatever the role. The logic was to propose editor role for authenticated user when the public access from ancestors was only readonly... but this was juged too much complexity for not enough value created. It can still be brought back later.~ Restrict link reach and link role independently. Link reach can only be improved and link role can only be improved.~ This restricts a lot what is possible but we prefer to offer something simple and as easy as possible to understand. We can open more possibilities later if people request the use cases we have ignored.
- Add ancestors and computed links definitions to document API endpoint : the frontend needs to display inherited link accesses when it displays possible select options. We need to return this information to the client.
- Fix creating/updating document accesses for teams
- Stop requiring an owner for non-root documents (avoid useless access rights overrides that will complicated document management). If a child document is moved to the root, we will get the owners of the current root ancestor and automatically add them as owner to the child after moving it to the root.

We also realized that we needed to display specific access rights inherited from ancestors. This PR adds inherited accesses to the list of accesses returned for a document on `/api/v1.0/documents/{uuid}/accesses/`


lunika (approved):

AntoLC:
@PanchoutNathan The request types `documents/[DOC_ID]/accesses/` changed, it is not a pagination anymore.
See: https://github.com/suitenumerique/docs/blob/main/src/frontend/apps/impress/src/features/docs/doc-share/api/useDocAccesses.tsx

src/backend/core/models.py
```diff
@@ -301,23 +306,20 @@ def _get_roles(self, resource, user):
...
+        return RoleChoices.max(*roles)
 
     def _get_abilities(self, resource, user):
         """
         Compute and return abilities for a given user taking into account
         the current state of the object.
         """
-        roles = self._get_roles(resource, user)
+        role = self._get_role(resource, user)
+        is_owner_or_admin = role in (RoleChoices.OWNER, RoleChoices.ADMIN)
```
lunika:
```suggestion
        is_owner_or_admin = role in PRIVILEGED_ROLES
```

src/backend/core/api/serializers.py
```diff
@@ -32,21 +32,10 @@ class Meta:
...
-    def get_email(self, _user):
-        """Return always None. Here to have the same fields than in UserSerializer."""
-        return None
-
     class Meta:
         model = models.User
-        fields = ["id", "email", "full_name", "short_name"]
-        read_only_fields = ["id", "email", "full_name", "short_name"]
+        fields = ["full_name", "short_name"]
+        read_only_fields = ["full_name", "short_name"]
```
lunika, line 38: I did that to avoid the front application to have multiple models.

src/backend/core/models.py
```diff
@@ -1101,48 +1041,91 @@ def save(self, *args, **kwargs):
         super().save(*args, **kwargs)
         self.document.invalidate_nb_accesses_cache()
 
+    @property
+    def target_key(self):
+        """Get a unique for the actor targeted by the access, without possible conflict."""
```
lunika, line 1046: a unique key ?

src/backend/core/api/serializers.py
sampaccoud: The question it raised was: how do I differentiate between an model who has the value and the value is None and a model that doesn't have it? 
I removed it because I was introducing more differences between the 2 serializers and it would be confusing to have these other fields as well with null values :thinking:

src/backend/core/tests/test_models_documents.py
```diff
@@ -1168,184 +1168,135 @@ def test_models_documents_restore_complex_bis(django_assert_num_queries):
 
 
 @pytest.mark.parametrize(
-    "ancestors_links, select_options",
+    "reach, role, select_options",
     [
         # One ancestor
```
lunika, line 1173:
```suggestion
```

Or maybe `with ancestors` in opposition to `no ancestor` few lines after ?

lunika (approved):

src/backend/core/tests/test_models_documents.py
sampaccoud: I removed it as this comment does not make sense anymore after the refactorings.

src/backend/core/api/permissions.py
```diff
@@ -96,26 +97,27 @@ def has_permission(self, request, view):
         ).exists()
 
 
-class AccessPermission(permissions.BasePermission):
-    """Permission class for access objects."""
+class ResourceWithAccessPermission(permissions.BasePermission):
+    """A permission class for templates."""
```
lunika, line 101: This seems also used for Invitations.

src/backend/core/api/permissions.py
```diff
@@ -127,10 +129,74 @@ def has_object_permission(self, request, view, obj):
...
+    """Permission class for template access objects."""
+
+    def has_permission(self, request, view):
+        """check create permission for template accesses."""
+        if super().has_permission(request, view) is False:
+            return False
+
+        if view.action == "create":
+            roles = view.template.get_roles(request.user)
+            if not set(roles).intersection(set(choices.PRIVILEGED_ROLES)):
```
lunika, line 186: This is the only difference between `TemplateAccessPermission` and `DocumentAccessPermission`. Is there a way to factorize it ?

src/backend/core/api/serializers.py
```diff
@@ -32,133 +32,43 @@ class Meta:
...
-        read_only_fields = ["id", "team", "role", "abilities"]
+    def update(self, instance, validated_data):
+        """Make "user" field is readonly but only on update."""
+        validated_data.pop("user", None)
+        return super().update(instance, validated_data)
 
 
-class TemplateAccessSerializer(BaseAccessSerializer):
-    """Serialize template accesses."""
+class DocumentMinimalSerializer(serializers.ModelSerializer):
```
lunika, line 65: What about `DocumentLight` as we are already using this semantic.

src/backend/core/models.py
```diff
@@ -1102,48 +1041,94 @@ def save(self, *args, **kwargs):
...
+        Get a tuple of:
+        - the role equivalent to all ancestors of the
+          document related to the current access
+        - the role the user has on the current access
+        """
+        if not user.is_authenticated:
+            return None, None
+
+        try:
+            return self.user_roles_tuple
```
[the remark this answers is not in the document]
sampaccoud: Yeah that's the heavy lifting... Computing it in the viewset is a bit different because we compute it for the whole queryset without incurring tens of db queries. It should safely be equivalent as it is heavily tested.

We can think of a better way to pass this pre-computation from the viewset to the model, but I think it should be done in another PR because all roles computation have this pattern. See similar "user_roles" attribute on the Document model.

src/backend/core/api/permissions.py
sampaccoud: Good catch! Consider it done.

src/backend/core/models.py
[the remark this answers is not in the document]
sampaccoud: @lunika in fact it's not completely the same case as I have to pass the user here :thinking: 
I tried to propose a fixup to make it less magic: https://github.com/suitenumerique/docs/pull/846/commits/5e110d22ad7802bffeb93589423793f5d28b372a
What do you prefer?

AntoLC:
Merged here: https://github.com/suitenumerique/docs/pull/701

PR #851, opened 2025-04-07
🐛(docker-compose) unbind the y-provider service with frontend

AntoLC:
## Purpose

We cannot add new js dependency locally when we bind the frontend with the y-provider service. It results in
"EPERM: operation not permitted" when the `node_modules` has to be updated.
Better to remove the binding, we can add the binding locally during development on the y-provider.

image


lunika:
restarting it does not resolve the issue ?
Are you updating the `nodemon` dependency when you have this error ?

AntoLC:
> restarting it does not resolve the issue ? Are you updating the `nodemon` dependency when you have this error ?

No, I just add dependencies on the app "impress", not even on the "y-provider". But it is a monorepo, so the `node_modules` is shared.

lunika (approved):

Issue #561, opened 2025-01-16
Add docker compose up production configuration


virgile-dev:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
Not all have a k8s cluster at hand.
Currently we only offer helm chart for production use.

**Describe the solution you'd like**
I'd like to have a simple way to install Docs on a webserver. With Docker Compose up.

**Describe alternatives you've considered**

**Discovery, Documentation, Adoption, Migration Strategy**
If you can, explain how users will be able to use this and possibly write out a version the docs (if applicable).
Maybe a screenshot or design?

**Do you want to work on it through a Pull Request?**

alfureu:
Hi all, first of all, thanks for the great software. I would like to vouch for a unified (and simpler) production configuration. Unfortunately, the document under `docs/installation/compose.md` is still pretty complicated, including the corresponding docker-compose file with many-many containers. 

I would like to highlight the comparison software like Siyuan, which is dead simple to deploy, or Outline (also simpler, although a bit more complicated than Siyuan). I know many would disagree with me for various reason, I am not using Siyuan for various reasons, I just want to highlight that things can be a bit streamlined, simpler, user-friendly, you name it...

virgile-dev:
@unteem is working on this https://github.com/suitenumerique/docs/pull/855

PR #583, opened 2025-01-28
Configuration to deploy using compose

lunika:
## Purpose

To ease the deployment of the application on infrastructure not running
k8s we created a compose file ready for production.

## Proposal

- [x] rename compose file
- [x] remove usage of dockerize
- [x] create a config to deploy docs using compose in production
- [x] add documentation to install docs in production using compose   

## Improvement

- [] manage certbot in the compose configuration

Fix #561

PR #855, opened 2025-04-07
Documentation for self-hosting with docker compose

unteem:
## Purpose

Make self hosting of Docs easier with an example of a deployment procedure with docker compose and document how to configure Docs. 

While https://github.com/suitenumerique/docs/pull/583 propose an easy way to deploy Docs with docker and Make, here we describe more in details the various steps and requirements to deploy Docs. 

## Proposal

- [x] example to deploy and configure keycloak
- [x] example to deploy and configure minio
- [x] example to configure proxy and certs
- [x] example to deploy and configure Docs

## Improvements
- [x] Rephrase description of environment variables and categorize
- [x] Use template for nginx conf  

Fixes https://github.com/suitenumerique/docs/issues/561
Supersedes https://github.com/suitenumerique/docs/pull/583

 A one liner quick start could be a nice addition:
- [ ] merge all services in a single compose
- [ ] scripts to generate secrets


docs/installation/compose.md
```diff
@@ -0,0 +1,221 @@
...
+
+### Step 1: Edit the images tag with the desired version
+
+### Step 2: Pull the images
+
+```bash
+docker compose pull
+```
+
+### Step 3: Restart yout containers
```
sylvinus, line 211: yout

docker/files/production/etc/nginx/conf.d/default.conf.template
```diff
@@ -0,0 +1,120 @@
...
+        proxy_set_header Host $host;
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header X-Original-URL $request_uri;
+        
+        # Prevent the body from being passed
+        proxy_pass_request_body off;
+        proxy_set_header Content-Length "";
+        proxy_set_header X-Original-Method $request_method;
+    }
```
vnghia, line 79: I think we don't need this anymore as mentioned [here](https://github.com/suitenumerique/docs/commit/544dd00c16313197b22debb29ff5cdbd99452de1). I am not sure if we still need `/media-auth` or not. I've managed to spin up `docs` with `traefik` without these two `/collaboration-auth` and `/media-auth` with socket connection and real time collaboration.

lunika, line 79: You can remove this part, it has been removed in the main branch, it is not used anymore.

docs/examples/compose/nginx-proxy/README.md
```diff
@@ -0,0 +1,37 @@
...
+## Installation
+
+### Step 1: Prepare your working environment:
+
+```bash
+mkdir nginx-proxy
+curl -o compose.yaml https://raw.githubusercontent.com/suitenumerique/docs/refs/heads/main/docs/examples/compose/nginx-proxy/compose.yaml
+```
+
+### Step 2: Edit `DEFAULT_EMAIL` in the compose file.
```
lunika, line 19: There is no description for this step 2.

docs/examples/compose/nginx-proxy/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Create docker network
```
lunika, line 21: Missing `step 3`

docs/installation/compose.md
```diff
@@ -0,0 +1,221 @@
...
```
lunika, line 211: Before restarting the containers you should suggest to read the `UPGRADE.md` file.

env.d/production.dist/common
```diff
@@ -0,0 +1,8 @@
+DOCS_HOST=docs.domain.tld
+KEYCLOAK_HOST=id.domain.tld
+S3_HOST=storage.domain.tld
+MINIO_HOST=storage.domain.tld
```
z3ntu, line 4: Is MINIO_HOST used?

env.d/production.dist/kc_postgresql
```diff
@@ -0,0 +1,13 @@
+# Postgresql db container configuration
+POSTGRES_DB=keycloak
+POSTGRES_USER=keycloak
+POSTGRES_PASSWORD=<generate postgres password>
+PGDATA=/var/lib/postgresql/data/pgdata
+
+# Keyclaok postgresql configuration
```
z3ntu, line 7: Keyclaok

docs/installation/compose.md
```diff
@@ -0,0 +1,221 @@
...
+- **Docker Compose:** Install it as per the [official documentation](https://docs.docker.com/compose/install/).
+
+> [!NOTE]
+> `docker-compose` may not be supported. You are advised to use `docker compose` instead.
+
+## Step 1: Prepare your working environment:
+
+```bash
+mkdir -p docs/env.d
+curl -o compose.yaml https://raw.githubusercontent.com/suitenumerique/docs/refs/heads/main/docs/examples/compose.yaml
```
z3ntu, line 34: you put the compose.yaml in a different directory than docs/examples/compose.yaml

docs/installation/compose.md
```diff
@@ -0,0 +1,221 @@
...
+curl -o env.d/common https://raw.githubusercontent.com/suitenumerique/docs/refs/heads/main/docs/env.d/production.dist/common
+curl -o env.d/backend https://raw.githubusercontent.com/suitenumerique/docs/refs/heads/main/docs/env.d/production.dist/backend
+curl -o env.d/yprovider https://raw.githubusercontent.com/suitenumerique/docs/refs/heads/main/docs/env.d/production.dist/yprovider
```
z3ntu, line 37: missing env.d/postgresql?

docs/examples/compose/keycloak/README.md
```diff
@@ -0,0 +1,88 @@
...
+2. Navigate to the realm tab and click on the "Create realm" button.
+3. Enter the name of the realm  - `docs`.
+4. Click "Create".
+
+#### Step 2: Create a New Client
+
+1. Navigate to the "Clients" tab.
+2. Click on the "Create client" button.
+3. Enter the client ID - e.g. `docs`.
+4. Enable "Client authentification" option.
```
z3ntu, line 80: Client authentication

docs/examples/compose/compose.yaml
```diff
@@ -0,0 +1,73 @@
...
+  backend:
+    image: lasuite/impress-backend:latest
+    user: ${DOCKER_USER:-1000}
+    restart: always
+    environment:
+    - DJANGO_CONFIGURATION=Production
+    env_file:
+    - env.d/backend
+    - env.d/postgresql
+    - env.d/common
```
z3ntu, line 29: I believe common needs to come before backend, otherwise you get errors like this
```
WARN[0000] The "DOCS_HOST" variable is not set. Defaulting to a blank string. 
WARN[0000] The "DOCS_HOST" variable is not set. Defaulting to a blank string. 
WARN[0000] The "S3_HOST" variable is not set. Defaulting to a blank string. 
```

docs/installation/compose.md
```diff
@@ -0,0 +1,221 @@
...
+docker compose up -d
+```
+> [!NOTE]
+> Version of the images are set to latest, you should pin it to the desired version to avoid unwanted upgrades when pulling latest image.
+
+## Step 5: Run the database migration and create Django admin user
+
+```bash
+docker compose run --rm app python manage.py migrate
+docker compose run --rm app python manage.py createsuperuser --email <admin email> --password <admin password>
```
z3ntu, line 192: "app" should be "backend"? also at the bottom of this file

docs/installation/compose.md
```diff
@@ -0,0 +1,221 @@
...
+```
+
+Replace `<admin email>` with the email of your admin user and generate a secure password. 
+
+Your docs instance is now avalailable on the domain you defined, https://docs.yourdomain.tld.
```
z3ntu, line 197: avalailable

docs/examples/compose/compose.yaml
```diff
@@ -0,0 +1,73 @@
...
+    user: "101"
+    env_file:
+    - env.d/common
+    # Uncomment and set your values if using our nginx proxy example
+    #environment:
+    # - VIRTUAL_HOST=${DOCS_HOST} # used by nginx proxy 
+    # - VIRTUAL_PORT=8083 # used by nginx proxy
+    # - LETSENCRYPT_HOST=${DOCS_HOST} # used by lets encrypt to generate TLS certificate
+    volumes:
+    - ./default.conf.template:/etc/nginx/templates/default.conf.template
```
z3ntu, line 61: I think `/docker-entrypoint.d/20-envsubst-on-templates.sh` is never run so the template never becomes a proper config.

The upstream nginx container uses `ENTRYPOINT ["/docker-entrypoint.sh"]` which should call this but this is overwritten in the `lasuite/impress-frontend` image to `/usr/local/bin/entrypoint`

Am I missing something?

docker/files/production/etc/nginx/conf.d/default.conf.template
virgile-dev: @unteem can you confirm?

mustafa0x:
This commit will make docs more accessible to a huge number of teams. Eagerly waiting to see it committed!

docker/files/production/etc/nginx/conf.d/default.conf.template
unteem: I need to keep `location /collaboration/ws/` to expose the websocket endpoint. 
I removed `/collaboration-auth` and `/collaboration/api` which are not needed anymore

docs/examples/compose/nginx-proxy/README.md
unteem: fixed

env.d/production.dist/common
unteem: yep not needed

docker/files/production/etc/nginx/conf.d/default.conf.template
```diff
@@ -0,0 +1,88 @@
...
+        # Preserve original host and additional headers
+        proxy_set_header X-Forwarded-Proto https;
+        proxy_set_header Origin $http_origin;
+        proxy_set_header Host $host;
+    }
+
+    # Proxy auth for media
+    location /media/ {
+        # Auth request configuration
+        auth_request /media-auth;
```
lunika, line 70: Are you sure this is working ? The backend endpoint is `/api/v1.0/documents/media-auth/`. How nginx queries it ?

docs/examples/compose/compose.yaml
```diff
@@ -0,0 +1,75 @@
+services:
+  postgresql:
+    image: postgres:16
+    healthcheck:
+      test: ["CMD", "pg_isready", "-q", "-U", "docs", "-d", "docs"]
```
lunika, line 5:
```suggestion
      test: ["CMD-SHELL", "pg_isready -d $${POSTGRES_DB} -U $${POSTGRES_USER}"]
```

lunika: I don't know the `-q` option, you can decide to keep it or not

docs/installation/compose.md
```diff
@@ -0,0 +1,226 @@
...
+We provide a sample configuration for running Docs using Docker Compose. Please note that this configuration is experimental, and the official way to deploy Docs in production is to use [k8s](../installation/k8s.md)
+
+## Requirements
+
+- A modern version of Docker and its Compose plugin.
+- A domain name and DNS configured to your server.
+- An Identity Provider that supports OpenID Connect protocol - we provide [an example to deploy Keycloak](../examples/compose/keycloak/README.md).
+- An Object Storage that implements S3 API - we provide [an example to deploy Minio](../examples/compose/minio/README.md).
+- A Postgresql database
+- A Redis database
```
lunika, line 12: Are they provided as an example or we should use it ?

docs/installation/compose.md
```diff
@@ -0,0 +1,226 @@
...
+### Step 3: Restart your containers
+
+```bash
+docker compose restart
+```
+
+### Step 4: Run the database migration
+Your database schema may need to be updated, run:
+```bash
+docker compose run --rm app python manage.py migrate
```
lunika, line 225:
```suggestion
docker compose run --rm backend python manage.py migrate
```

docs/examples/compose/keycloak/compose.yaml
```diff
@@ -0,0 +1,36 @@
+services:
+  kc_postgresql:
+    image: postgres:16
+    healthcheck:
+      test: ["CMD", "pg_isready", "-q", "-U", "keycloak", "-d", "keycloak"]
```
lunika, line 5: Same here, you should use the environment variables provided in the `env.d/kc_postgresql` otherwise once changed this check will not work anymore.

docker/files/production/etc/nginx/conf.d/default.conf.template
unteem: yep my bad, got confused and removed `/media-auth` part while removing the `collaboration-auth`

docs/installation/compose.md
unteem: I added :
> we provide [an example in the compose file](../examples/compose/compose.yaml).

Does it make it more clear ? 

In Redis and Postgres sections it is also mentioned :
> Docs uses Redis for caching. While an external Redis can be used, our example provides a deployment method.

docs/examples/compose/compose.yaml
unteem: `-q` is for quiet, I just removed it for consistency

docker/files/production/etc/nginx/conf.d/default.conf.template
unteem: `/collaboration-auth` can be removed but we still need to keep `media-auth`

docs/installation/compose.md
```diff
@@ -0,0 +1,226 @@
...
+```bash
+mkdir -p docs/env.d
+curl -o compose.yaml https://raw.githubusercontent.com/suitenumerique/docs/refs/heads/main/docs/examples/compose/compose.yaml
+curl -o env.d/common https://raw.githubusercontent.com/suitenumerique/docs/refs/heads/main/docs/env.d/production.dist/common
+curl -o env.d/backend https://raw.githubusercontent.com/suitenumerique/docs/refs/heads/main/docs/env.d/production.dist/backend
+curl -o env.d/yprovider https://raw.githubusercontent.com/suitenumerique/docs/refs/heads/main/docs/env.d/production.dist/yprovider
+curl -o env.d/common https://raw.githubusercontent.com/suitenumerique/docs/refs/heads/main/docs/env.d/production.dist/postgresql
+```
+
+## Step 2: Configuration
```
virgile-dev, line 41: Hey @unteem ! This section is not super specific to compose. 
Should we document in a generic file like a configuration.md ?

virgile-dev (changes requested):
How about we keep an installation.md  at the root of the /docs folder where we can list the different deployment option (K8S, Compose and NIX) ?

docs/examples/compose/compose.yaml
z3ntu: I don't think this is working either for me. A few seconds after launching the containers I get `frontend-1 exited with code 0` , no other logs are visible. Not sure how to debug this...

virgile-dev (approved):
Let's merge this and make documentation for different deployment solution more homogeneous in a next PR.

docs/examples/compose/compose.yaml
z3ntu: That's it now with the updated config (https://github.com/suitenumerique/docs/compare/b5cd3e783ee5919512aa3e05596bc3e6ced31a03..a8f1684cb1df1e367f14b3e27be01712e395b277)
```
frontend-1    | /docker-entrypoint.sh: /docker-entrypoint.d/ is not empty, will attempt to perform configuration
frontend-1    | /docker-entrypoint.sh: Looking for shell scripts in /docker-entrypoint.d/
frontend-1    | /docker-entrypoint.sh: Launching /docker-entrypoint.d/10-listen-on-ipv6-by-default.sh
frontend-1    | 10-listen-on-ipv6-by-default.sh: info: can not modify /etc/nginx/conf.d/default.conf (read-only file system?)
frontend-1    | /docker-entrypoint.sh: Sourcing /docker-entrypoint.d/15-local-resolvers.envsh
frontend-1    | /docker-entrypoint.sh: Launching /docker-entrypoint.d/20-envsubst-on-templates.sh
frontend-1    | 20-envsubst-on-templates.sh: Running envsubst on /etc/nginx/templates/default.conf.template to /etc/nginx/conf.d/default.conf
frontend-1    | /docker-entrypoint.d/20-envsubst-on-templates.sh: line 53: can't create /etc/nginx/conf.d/default.conf: Permission denied
frontend-1 exited with code 1
```
Could you please share the log from the container on your side? I feel like we're trying two different things.

unteem: yep my bad, I forgot to add the `command` entry otherwise it does not work. I double checked, now I'm ISO with my working test environment.

FYI I made a couple changes, mainly adding `env.d/yprovider` to the backend container which was missing

unteem: arf sorry again...  I missed to many changes made in my test env while going back on forth 

```
    volumes:
    - ./default.conf.template:/etc/nginx/templates/docs.conf.template
```

default.conf is readonly, it needs to be mounted in another file.

virgile-dev:
@unteem can you make the last test pass so we can merge this please ?

env.d/production.dist/backend
```diff
@@ -0,0 +1,65 @@
...
+OIDC_OP_AUTHORIZATION_ENDPOINT=https://${KEYCLOAK_HOST}/realms/${REALM_NAME}/protocol/openid-connect/auth
+OIDC_OP_TOKEN_ENDPOINT=https://${KEYCLOAK_HOST}/realms/${REALM_NAME}/protocol/openid-connect/token
+OIDC_OP_USER_ENDPOINT=https://${KEYCLOAK_HOST}/realms/${REALM_NAME}/protocol/openid-connect/userinfo
+OIDC_OP_LOGOUT_ENDPOINT=https://${KEYCLOAK_HOST}/realms/${REALM_NAME}/protocol/openid-connect/logout
+OIDC_RP_CLIENT_ID=<client_id>
+OIDC_RP_CLIENT_SECRET=<client secret>
+OIDC_RP_SIGN_ALGO=RS256
+OIDC_RP_SCOPES="openid email"
+#USER_OIDC_FIELD_TO_SHORTNAME
+#USER_OIDC_FIELDS_TO_FULLNAME
```
z3ntu, line 47: These two variables seem to have been renamed? https://github.com/suitenumerique/docs/blob/d952815/src/backend/core/authentication/backends.py#L17-L28

env.d/production.dist/yprovider
```diff
@@ -0,0 +1,7 @@
+Y_PROVIDER_API_BASE_URL=http://${YPROVIDER_HOST}:4444/api
+Y_PROVIDER_API_KEY=<generate a random key>
+COLLABORATION_SERVER_SECRET=<generate a random key>
+COLLABORATION_SERVER_ORIGIN=https://${DOCS_HOST}
+COLLABORATION_API_URL=https://${DOCS_HOST}/collaboration/api/
```
z3ntu, line 5: Something's a bit wrong with this value I think? I'm getting this error e.g. when removing a user from a document.
```
backend-1  |   File "/usr/local/lib/python3.12/site-packages/rest_framework/mixins.py", line 91, in destroy
backend-1  |     self.perform_destroy(instance)
backend-1  |   File "/app/core/api/viewsets.py", line 1466, in perform_destroy
backend-1  |     CollaborationService().reset_connections(
backend-1  |   File "/app/core/services/collaboration_services.py", line 40, in reset_connections
backend-1  |     raise requests.HTTPError(
backend-1  | requests.exceptions.HTTPError: Failed to notify WebSocket server. Status code: 404, Response: <snip>
```

Does `/collaboration/api/` not needed to be proxied in the `default.conf.template` file to `YPROVIDER_HOST`?

edit: related comment https://github.com/suitenumerique/docs/pull/855#discussion_r2097948468

docker/files/production/etc/nginx/conf.d/default.conf.template
```diff
@@ -0,0 +1,102 @@
...
+        try_files $uri @proxy_to_docs_frontend;
+    }
+
+    location /api {
+        try_files $uri @proxy_to_docs_backend;
+    }
+
+    location /admin {
+        try_files $uri @proxy_to_docs_backend;
+    }
```
z3ntu, line 45: `/admin` also requires `/static` proxied to the backend from what I can tell, otherwise all resources (mostly visible is CSS) are missing when accessing the admin console.
```nginx
    location /static {
        try_files $uri @proxy_to_docs_backend;
    }
```

docs/examples/compose/minio/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+```shellscript
+mc mb --with-versioning minio/<your-bucket-name>
+```
+
+Replace `your-bucket-name` with the desired name for your bucket e.g. `docs-media-storage`
+
+### Additional notes:
+
+For increased security you should create a dedicated user with `readwrite` access to the Bucket. In the following example we will use MinIO root user.
```
z3ntu, line 103: This should be clarified imo, it took me a while until I made the connection of currently root user & password credentials being the ones for
```ini
AWS_S3_ACCESS_KEY_ID=<s3 access key>
AWS_S3_SECRET_ACCESS_KEY=<s3 secret key>
```
Also maybe at least link to outside docs that explain how to set up a user? I imagine most (including me) are not familiar with how one usually sets up minio.

I'm currently trying to set that up, from my understanding I need to create a new user, https://min.io/docs/minio/linux/administration/identity-access-management/minio-user-management.html#id1 and then it sounds like I shouldn't use the user credentials directly but create an accesskey to put into the config? https://min.io/docs/minio/linux/reference/minio-mc-admin/mc-admin-accesskey-create.html#command-mc.admin.accesskey.create

docker/files/production/etc/nginx/conf.d/default.conf.template
unteem: yep you are right. Fixed

env.d/production.dist/yprovider
unteem: I thought I could drop it, my bad it seems to be used to reset connections. Fixed

docs/examples/compose/minio/README.md
unteem: Initially I added an example but I eventually dropped it. The intention here is more to provide a quick start example than a thorough process to go to production. I don't know if we want to go that deep in details, like creating a user with the correct permissions. 

Docs requires some "complex" parts like an IDP and Object Storage, I think its important for those who want to self host those parts to read the upstream doc and understand by themselves all the implications. 

In this context I think skipping user and policy creation and using the minio root user is good enough as an example. 

In this case to create the user it would look like this:
1. Create policy for the bucket
```
cat > readwritedocs.json << EOF
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:*"
            ],
            "Resource": [
                "arn:aws:s3:::<your-bucket-name>/*"
            ]
        }
    ]
}
EOF
```
```
mc admin policy create myminio readwritedocs readwritedocs.json
```
2. Create user
 ```
mc admin user add myminio <username> <password>
```
4. Assign policy to user
```
mc admin policy attach myminio getonly --user=<your-user>
```

Comment (unteem) [reply docs/examples/compose/minio/README.md:103]: curious to have your thoughts on this @virgile-dev
should I add this part in the example ?

**Comment by AbdelRahmanAlTamimi** [comment]
@unteem 

Hi, thank you for the update!  
Just a quick note — the `curl` commands in this section return a 404 error:  
https://github.com/unteem/docs/blob/documentation-docker/docs/installation/compose.md#step-1-prepare-your-working-environment

It seems the URLs might be outdated or incorrect. Could you please double-check them?

Thanks again!

File: docs/examples/compose/minio/README.md (hunk 24 above)
Comment (virgile-dev) [reply docs/examples/compose/minio/README.md:103]: @lunika What do you think ?

Comment (lunika) [reply docs/examples/compose/minio/README.md:103]: A warning can be added on the OIDC and Minio part to make sure people read the related project documentation on how to go in production with these tools.

Comment (virgile-dev) [reply docs/examples/compose/minio/README.md:103]: @unteem can you update accordingly please ?

Comment (z3ntu) [reply docs/examples/compose/minio/README.md:103]: @unteem Thanks for sharing the policy stuff!

And I guess the step after is to run `mc admin accesskey create myalias <username> --access-key $AWS_S3_ACCESS_KEY_ID --secret-key $AWS_S3_SECRET_ACCESS_KEY`, right? It's a bit difficult to figure out what Minio recommends for such as setup, e.g. you can also use the Minio root user & password for `AWS_S3_ACCESS_KEY_ID` and `AWS_S3_SECRET_ACCESS_KEY` and that works totally fine. I assume you can also use the user credentials we just created and that'll work. But given the variable naming in docs, I assume creating an accesskey is the recommended thing to do.

From my perspective, I'm setting this up as a small instance for at most 20 people that use it occasionally, so we don't need any fancy big setup with sharding, load balancing, etc. We also only have everything running on one server, so there's zero high-availability anyways.

So having in these docs here how to get to a usable and secure production (while ignoring all HA and extra things for larger instances) would be good imo.

Thanks again for working on this!

File: docs/examples/compose/compose.yaml [hunk 25]
```diff
@@ -0,0 +1,78 @@
...
+    env_file:
+    - env.d/postgresql
+    - env.d/common
+    environment:
+    - PGDATA=/var/lib/postgresql/data/pgdata
+    volumes:
+    - ./data/databases/backend:/var/lib/postgresql/data/pgdata
+
+  redis:
+    image: redis:5
```
z3ntu: Am I mistaken or is Redis 5 EOL since a long time?

https://redis.io/docs/latest/operate/rs/installing-upgrading/product-lifecycle/

Given the new license of Redis it might also make sense to use Valkey by default instead?

docs/examples/compose/minio/README.md
unteem: > A warning can be added on the OIDC and Minio part to make sure people read the related project documentation on how to go in production with these tools.

Its already the case, there are 3 warnings in the beginning pointing to upstream documentation

```
> \[!CAUTION\]
> We provide those instructions as an example, it should not be run in production. For production environments, deploy MinIO [in a Multi-Node Multi-Drive (Distributed)](https://min.io/docs/minio/linux/operations/install-deploy-manage/deploy-minio-multi-node-multi-drive.html#minio-mnmd) topology
````
```

> \[!CAUTION\]
> We provide those instructions as an example, for production environments, you should follow the [official documentation](https://www.keycloak.org/documentation).
``` 

On minio I also point out that in prod, minio root user should not be used but a user with readwrite access

I'm not sure I can do more ?

unteem: > @unteem Thanks for sharing the policy stuff!
> 
> And I guess the step after is to run `mc admin accesskey create myalias <username> --access-key $AWS_S3_ACCESS_KEY_ID --secret-key $AWS_S3_SECRET_ACCESS_KEY`, right? It's a bit difficult to figure out what Minio recommends for such as setup, e.g. you can also use the Minio root user & password for `AWS_S3_ACCESS_KEY_ID` and `AWS_S3_SECRET_ACCESS_KEY` and that works totally fine. I assume you can also use the user credentials we just created and that'll work. But given the variable naming in docs, I assume creating an accesskey is the recommended thing to do.
> 
> From my perspective, I'm setting this up as a small instance for at most 20 people that use it occasionally, so we don't need any fancy big setup with sharding, load balancing, etc. We also only have everything running on one server, so there's zero high-availability anyways.
> 
> So having in these docs here how to get to a usable and secure production (while ignoring all HA and extra things for larger instances) would be good imo.
> 
> Thanks again for working on this!

Yep if you want to use `mc` to do operations like creating a user, CRUD on buckets/objects etc. you will need to [set an alias](https://min.io/docs/minio/linux/reference/minio-mc/mc-alias-set.html#command-mc.alias.set)

I would recommend to use a specific user and not the root/admin even if in your setup it could be fine to use the admin user, its always better to have least privileges access. It makes even more sense if at one point you use this minio instance for other apps, better to have one "user" per app.

unteem:
> @unteem
> 
> Hi, thank you for the update! Just a quick note — the `curl` commands in this section return a 404 error: https://github.com/unteem/docs/blob/documentation-docker/docs/installation/compose.md#step-1-prepare-your-working-environment
> 
> It seems the URLs might be outdated or incorrect. Could you please double-check them?
> 
> Thanks again!

URL are broken because they are pointing to the final URL, as the PR is not merged yet you need to replace https://raw.githubusercontent.com/suitenumerique/ by https://raw.githubusercontent.com/unteem/

unteem:
> > @unteem
> > Hi, thank you for the update! Just a quick note — the `curl` commands in this section return a 404 error: https://github.com/unteem/docs/blob/documentation-docker/docs/installation/compose.md#step-1-prepare-your-working-environment
> > It seems the URLs might be outdated or incorrect. Could you please double-check them?
> > Thanks again!
> 
> URL are broken because they are pointing to the final URL, as the PR is not merged yet you need to replace https://raw.githubusercontent.com/suitenumerique/ by https://raw.githubusercontent.com/unteem/

Ok some were broken :see_no_evil: and you need to replace https://raw.githubusercontent.com/suitenumerique/docs/refs/heads/main/ by https://raw.githubusercontent.com/unteem/docs/refs/heads/documentation-docker/ (with my branch). 

Its fixed now

docs/examples/compose/compose.yaml
unteem: good catch update to v8, I don't use redis in prod... :)
I think as an example it should be fine to leave redis, Valkey is indeed a good option.

env.d/production.dist/backend
z3ntu: I've fixed that in https://github.com/suitenumerique/docs/pull/1100

AbdelRahmanAlTamimi:
> > > @unteem
> > > Hi, thank you for the update! Just a quick note — the `curl` commands in this section return a 404 error: https://github.com/unteem/docs/blob/documentation-docker/docs/installation/compose.md#step-1-prepare-your-working-environment
> > > It seems the URLs might be outdated or incorrect. Could you please double-check them?
> > > Thanks again!
> > 
> > 
> > URL are broken because they are pointing to the final URL, as the PR is not merged yet you need to replace https://raw.githubusercontent.com/suitenumerique/ by https://raw.githubusercontent.com/unteem/
> 
> Ok some were broken 🙈 and you need to replace https://raw.githubusercontent.com/suitenumerique/docs/refs/heads/main/ by https://raw.githubusercontent.com/unteem/docs/refs/heads/documentation-docker/ (with my branch).
> 
> Its fixed now

Kind note: The sixth command in first step should be named postgresql instead of common.

PR #857, opened 2025-04-08
Feature/custom translations frontend

rvveber:
Adds the possibilty to inject/override custom translations at runtime.


src/frontend/apps/impress/src/features/language/hooks/useTranslationsCustomizer.ts
```diff
@@ -0,0 +1,138 @@
...
+    }
+    return localStorage.getItem(key);
+  },
+  setItem: (key: string, value: string): void => {
+    if (typeof window === 'undefined') {
+      return;
+    }
+    localStorage.setItem(key, value);
+  },
+};
```
AntoLC, line 23: Quite usefull, could maybe be available for the all Impress app. What about adding this code in `/src/utils/storages.ts` ?

src/frontend/apps/impress/src/features/language/hooks/useTranslationsCustomizer.ts
```diff
@@ -0,0 +1,138 @@
...
+
+export const useTranslationsCustomizer = () => {
+  const { data: conf } = useConfig();
```
AntoLC, line 26:
```suggestion
export const useTranslationsCustomizer = (translationsUrl:string) => {
```

docs/customization.md
```diff
@@ -0,0 +1,75 @@
...
+{
+  "en": {
+    "Docs": "Notes",
+    "Create New Document": "+"
+  },
+  "de": {
+    "Docs": "Notizen",
+    "Create New Document": "+"
+  }
+}
```
AntoLC, line 41:
```suggestion
{
  "en": {
    "translation": {
      "Docs": "Notes",
      "Create New Document": "+"
    }
  },
  "de": {
    "translation": {
      "Docs": "Notizen",
      "Create New Document": "+"
    }
  }
}
```

docs/customization.md
```diff
@@ -0,0 +1,75 @@
...
+```
+
+Then set the `FRONTEND_CUSTOM_TRANSLATIONS_URL` environment variable to the URL of this JSON file. The application will load these translations and override the default ones where specified.
+
+## Custom CSS
```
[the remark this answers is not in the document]
rvveber: Hm, but currently it is just custom CSS right?

- For _Theming_ it would need to include an example of overwriting a css variable that is generated by the theme.
- All customizations are/should be and usually are expected to be at runtime, from a customizers point of view atleast, so explicitly adding runtime, suggest that translations are not customized at runtime. It is confusing
- Target user group for documentation is one that knows what css is, it's the requirement to theme, i don't think we need to name it something else to explain what is possible to non-technicals. For that maybe the README would be a better place.
- We want to avoid build-time customizations entirely, and for now all content of this documentation is at run-time, why make it extra?

src/frontend/apps/impress/src/features/language/hooks/useTranslationsCustomizer.ts
```diff
@@ -0,0 +1,138 @@
...
+    if (isCustomized !== null || !conf) {
+      return;
+    }
+
+    const customUrl = conf.FRONTEND_CUSTOM_TRANSLATIONS_URL;
+    const cachedUrl = safeLocalStorage.getItem(CACHE_URL_KEY);
+
+    // Fast path: If we have cached translations for the same URL
+    if (cachedUrl === customUrl) {
+      const cached = safeLocalStorage.getItem(CACHE_KEY);
```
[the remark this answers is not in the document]
rvveber: I cached it, because we are waiting on two async requests here
1. fetch conf to receive the FRONTEND_CUSTOM_TRANSLATIONS_URL
2. fetch FRONTEND_CUSTOM_TRANSLATIONS_URL to receive the json

I think it is irresponsible to not cache it, since it would define a new critical-path/bottleneck for render-blocking loading times.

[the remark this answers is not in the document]
rvveber: But i'm also seeing now that i should not need to wait for conf.
1. If we have translations in cache, we should already apply them.
2. If the parallelly fetched translations are different we should apply them again.

docs/customization.md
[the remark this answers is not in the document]
AntoLC: We want to offer 2 different ways, to override the theme, 1 runtime, 1 buildtime, so it makes sense to say that this one is the runtime way. It depends the case, runtime does not every time fit the infrastructure you are working with, in our case we need the build time way for the moment concerning the theme.

src/frontend/apps/impress/src/features/language/hooks/useTranslationsCustomizer.ts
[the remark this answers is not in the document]
AntoLC: I think that is a good point (`critical-path/bottleneck for render-blocking loading times`). I want to apply it to the config endpoint.

What about this pattern, more react-querish ?

```ts
import { useQuery } from '@tanstack/react-query';
import { Resource } from 'i18next';

import { APIError, errorCauses } from '@/api';

const LOCAL_STORAGE_KEY = 'CUSTOM_TRANSLATIONS';

function getCachedTranslation() {
  try {
    const jsonString = localStorage.getItem(LOCAL_STORAGE_KEY);
    return jsonString ? (JSON.parse(jsonString) as Resource) : undefined;
  } catch {
    return undefined;
  }
}

function setCachedTranslation(translations: Resource) {
  localStorage.setItem(LOCAL_STORAGE_KEY, JSON.stringify(translations));
}

export interface FetchCustomTranslationeParams {
  url?: string;
}

export const fetchCustomTranslation = async ({
  url,
}: FetchCustomTranslationeParams) => {
  if (!url) {
    return null;
  }

  const response = await fetch(url);
  if (!response.ok) {
    throw new APIError(
      'Failed to get the translation',
      await errorCauses(response),
    );
  }

  return response.json() as Promise<Resource>;
};

export const KEY_TRANSLATION = 'custom-translation';

export function useCustomTranslation(url?: string) {
  const cachedData = getCachedTranslation();
  const response = useQuery<Resource | null, APIError>({
    queryKey: [KEY_TRANSLATION, url || ''],
    queryFn: () => fetchCustomTranslation({ url }),
    initialData: cachedData,
    staleTime: 0,
    refetchOnWindowFocus: false,
    refetchOnReconnect: false,
  });

  if (response.data && response.isSuccess) {
    setCachedTranslation(response.data);
  }

  return response;
}

```

Then to call it, it would be quite easy:
```ts
export const useTranslationsCustomizer = (translationUrl?: string) => {
  const { i18n } = useTranslation();
  const { data: translations } = useCustomTranslation(translationUrl);

  useEffect(() => {
    if (!translations) {
      return;
    }

    Object.entries(translations).forEach(([lng, namespaces]) => {
      Object.entries(namespaces).forEach(([ns, value]) => {
        i18next.addResourceBundle(lng, ns, value, true, true);
      });
    });

    const currentLanguage = i18n.language;
    void i18next.changeLanguage(currentLanguage);

    // Emit added event to make sure components re-render
    i18next.emit('added', currentLanguage, 'translation');
  }, [i18n.language, translations]);
};
```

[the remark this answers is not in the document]
AntoLC: I made this PR about the config cache optimization: https://github.com/suitenumerique/docs/pull/867

src/frontend/apps/impress/src/core/config/ConfigProvider.tsx
```diff
@@ -35,8 +39,10 @@ export const ConfigProvider = ({ children }: PropsWithChildren) => {
   }, [conf?.FRONTEND_THEME, setTheme]);
 
   useEffect(() => {
-    void synchronizeLanguage();
-  }, [synchronizeLanguage]);
+    void synchronizeLanguage().finally(() => {
+      void customizeTranslations();
```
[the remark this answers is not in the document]
rvveber: I improved it. No more async or void calls and no dependency chains.
Check it out.

docs/customization.md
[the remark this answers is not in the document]
rvveber: ✅

docs/customization.md
```diff
@@ -0,0 +1,75 @@
+# Application Customization 🛠️
+
+This document outlines the various ways you can customize our application to suit your specific needs without modifying the core codebase.
+
+## Available Customization Options
+
+1. [Translations](#custom-translations)
+2. [CSS/Theme](#custom-css)
+
```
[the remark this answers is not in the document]
rvveber: ✅

src/frontend/apps/impress/src/features/language/hooks/useTranslationsCustomizer.ts
rvveber: ✅

src/frontend/apps/impress/src/features/language/hooks/useTranslationsCustomizer.ts
rvveber: I decouple all required vars as parameters, for both TranslationCustomizer and LanguageSynchronizer
✅

docs/customization.md
rvveber: ✅

src/frontend/apps/e2e/__tests__/app-impress/config.spec.ts
```diff
@@ -170,6 +170,67 @@ test.describe('Config', () => {
...
+    const mockTranslationsUrl =
+      'http://dummyhost.example.com/translations/custom.json';
+
+    // Mock the config endpoint to include the custom translations URL
+    await page.route('**/api/v1.0/config/', async (route) => {
+      const request = route.request();
+      if (request.method().includes('GET')) {
+        await route.fulfill({
+          json: {
+            ...config,
```
AntoLC, line 187:
```suggestion
            ...CONFIG,
```

src/frontend/apps/impress/src/features/language/hooks/useTranslationsCustomizer.ts
AntoLC, line 1: The problem that I see is that we change the pattern of the application, all the features used `/api/` to have a clear separation of concern, and we use `react-query`. 
By leveraging `react-query`, no need to do tricks with `useRef` as `react-query` has an inner cache, it will be more readable by reducing `conditions` and `try catch`.

rvveber: Yes, i agree! 

I've spent some time, to experiment with this pattern, and i noticed that there are more inconcistencies across the code.
I wanted to set an example for best practice separation of concerns for Queries and Mutations, making sense of the CRUD principle.

I have added 5 commits that explain it in detail.
And i added a little extra (1. assert, 2. debug) for Developer Experience.

Let me know how it looks to you 👍🏼

AntoLC, line 1: For me we still have a pattern problem not leveraging React Query, let's wait for #961 so.

src/frontend/apps/impress/src/features/language/hooks/useSynchronizedLanguage.ts
```diff
@@ -0,0 +1,105 @@
...
+      if (i18n.resolvedLanguage === closestFrontendLanguage) {
+        return;
+      }
+
+      log('Updating frontend language (%O)', {
+        requested: language,
+        from: i18n.resolvedLanguage,
+        to: closestFrontendLanguage,
+      });
+      void i18n.changeLanguage(closestFrontendLanguage);
```
AntoLC, line 70: Same it is part of the development part, should be removed.

src/frontend/apps/impress/src/features/auth/api/useAuthMutation.ts
```diff
@@ -0,0 +1,80 @@
...
+import { User } from './types';
+import { KEY_AUTH } from './useAuthQuery';
+
+// Mutations are separated from queries to allow for better separation of concerns.
+// Mutations are responsible for C (CREATE), U (UPDATE), and D (DELETE) in CRUD.
+
+// --- Create ---
+function createUser(): Promise<User> {
+  throw new Error('Not yet implemented.');
+}
```
AntoLC, line 14: We are creating things that we don't need yet (maybe never ?), same for the deleteUser.

After this feature is made to manage the authentication, I don't know if it is the good place to add the user mutation.

src/frontend/apps/impress/src/features/language/hooks/useSynchronizedLanguage.ts
```diff
@@ -0,0 +1,105 @@
...
+    if (currentBackendLanguage) {
+      changeLanguageSynchronized(currentBackendLanguage);
+    } else if (currentFrontendLanguage) {
+      changeLanguageSynchronized(currentFrontendLanguage);
+    }
+  }, [
+    currentBackendLanguage,
+    currentFrontendLanguage,
+    changeLanguageSynchronized,
+    currentUser,
```
AntoLC, line 93: Is `currentUser` necessary ?

src/frontend/apps/impress/src/features/language/hooks/useSynchronizedLanguage.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  const changeLanguageSynchronized = useCallback(
+    (language: string, user?: User) => {
+      changeFrontendLanguage(language);
+      changeBackendLanguage(language, user ?? currentUser);
+    },
+    [changeBackendLanguage, changeFrontendLanguage, currentUser],
+  );
+
+  useEffect(() => {
```
AntoLC, line 83: I have the feeling this useEffect is trigger often.

src/frontend/apps/impress/src/features/auth/api/types.ts
```diff
@@ -11,5 +11,5 @@ export interface User {
   email: string;
   full_name: string;
   short_name: string;
-  language: string;
+  language?: string;
```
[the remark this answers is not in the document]
rvveber: Yes, language is nullable.
It was introduced with the language synchronization feature and should've been changed back then.

It allows us to update the users preferred language automatically from the i18n browser detected language.

src/frontend/apps/impress/src/features/auth/api/useAuthMutation.ts
rvveber: Thanks for your valuable feedback on the `useAuthMutation` pull request. I appreciate you highlighting the concerns around the scope and placement of the user-related mutations.

I agree with your observations, and I believe a couple of changes will clarify the intent and improve the structure:

1.  **Renaming for Clarity:**
    * You're right, the current names are a bit misleading. I propose renaming:
        * `useAuthQuery` to `useUserQuery`
        * `getMe` to `readUser` (which will default to fetching `/users/me/` but could accept an ID).
        * `useAuthMutation` to `useUserMutation`.
    * This better reflects that we are fundamentally querying and mutating `User` data.

2.  **Relocation to Entity-Specific Path:**
    * Following the renaming, I'll move these hooks to a more appropriate location, such as `api/entities/user/`. This aligns with the idea that these are general-purpose tools for managing the `User` entity, not strictly tied to the authentication feature's codebase.

**Regarding `createUser` and `deleteUser`:**

My intention in including `createUser` and `deleteUser` (even as unimplemented stubs for now) within `useUserMutation` was to:

* **Establish a Complete CRUD Pattern:** Provide a full, standard interface (Create, Read, Update, Delete) for the `User` entity from the outset. This is a common and robust pattern, similar to what one might find in ORMs, and it makes the API for `User` management predictable.
* **Serve as a Clear Blueprint:** This structure acts as a boilerplate, clearly defining how all entity mutations should be handled. It sets an example for consistency when we add mutations for other entities in the future.
* **Enhance Readability & Single Responsibility:** Grouping all mutation logic for a `User` (`create`, `update`, `delete`) within `useUserMutation` maintains a clear single responsibility for modifying user data. I believe this makes the code more self-documenting and easier to navigate.
* **Future Preparedness with Minimal Overhead:** While these functions aren't used by the current feature iteration, defining them now prepares us for future requirements. The stubs themselves add negligible overhead but offer significant structural benefits and clarity for ongoing development.

By renaming and relocating these hooks to focus on the `User` entity, I believe the inclusion of the full CRUD set (even with stubs) becomes more logical as it defines the comprehensive API for `User` management.

I've spent considerable time researching best practices for data fetching and mutations in React, and this pattern of co-locating entity-specific hooks with a complete operational set seemed the most explicit and maintainable.

Please let me know if these proposed changes and this explanation address your concerns. 
I'm keen to ensure we're building a clean and scalable foundation.

src/frontend/apps/impress/src/features/language/hooks/useCustomTranslations.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export const useCustomTranslations = () => {
+  const { i18n } = useTranslation();
+  const { data: currentConfig } = useConfig();
+
+  const currentCustomTranslations: Resource = useMemo(
+    () => currentConfig?.theme_customization?.translations || {},
+    [currentConfig],
+  );
+
```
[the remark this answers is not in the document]
rvveber: Thanks for the suggestion on simplifying `currentCustomTranslations`!

You're right that it would functionally update correctly. The main reason for keeping `useMemo` here is to optimize the `useEffect` that depends on `currentCustomTranslations`.

Specifically, `useMemo` ensures **referential stability**. If `translations` are empty (`{}`), without `useMemo`, a new `{}` object instance would be created on every render. This would cause the `useEffect` to run unnecessarily, even if the actual `currentConfig` hasn't changed. `useMemo` prevents this by returning the same object reference if the dependencies haven't changed.

It's a subtle point, but helps avoid those extra effect runs!

src/frontend/apps/impress/src/features/header/components/Header.tsx
```diff
@@ -5,7 +5,7 @@ import IconDocs from '@/assets/icons/icon-docs.svg';
 import { Box, StyledLink } from '@/components/';
 import { useCunninghamTheme } from '@/cunningham';
 import { ButtonLogin } from '@/features/auth';
-import { LanguagePicker } from '@/features/language';
+import { LanguagePicker } from '@/features/language/components';
```
AntoLC, line 8:
```suggestion
import { LanguagePicker } from '@/features/language';
```

src/frontend/apps/impress/src/features/home/components/HomeHeader.tsx
```diff
@@ -6,7 +6,7 @@ import { Box } from '@/components';
 import { useCunninghamTheme } from '@/cunningham';
 import { ButtonTogglePanel, Title } from '@/features/header/';
 import { LaGaufre } from '@/features/header/components/LaGaufre';
-import { LanguagePicker } from '@/features/language';
+import { LanguagePicker } from '@/features/language/components';
```
AntoLC, line 9:
```suggestion
import { LanguagePicker } from '@/features/language';
```

src/frontend/apps/impress/src/features/left-panel/components/LeftPanel.tsx
```diff
@@ -6,7 +6,7 @@ import { Box, SeparatedSection } from '@/components';
 import { useCunninghamTheme } from '@/cunningham';
 import { ButtonLogin } from '@/features/auth';
 import { HEADER_HEIGHT } from '@/features/header/conf';
-import { LanguagePicker } from '@/features/language';
+import { LanguagePicker } from '@/features/language/components';
```
AntoLC, line 9:
```suggestion
import { LanguagePicker } from '@/features/language';
```

AntoLC (approved):
Well done, very useful feature. 🚀

PR #866, opened 2025-04-10
🐛(nginx) fix 404 when accessing a doc

AntoLC:
## Purpose

We improve the nginx way to access to a specific doc, to not have a 404 before accessing the resource.

## Proposal

We stop to wait for a initial attempt that give a 404. 
If we see a UUID in the url we will redirect to the doc/[id] page. 
`Next` will then manage the 404.

## Easy way to test

- go to `src/frontend/apps/impress`
- `yarn build`
- Access the app from nginx, in `docker-compose.yml`, add this service:
```yaml
  nginx-ci:
    image: nginx:1.25
    ports:
      - "3000:3000"
    volumes:
      - ./src/frontend/apps/impress/conf/default.conf:/etc/nginx/conf.d/default.conf
      - ./src/frontend/apps/impress/out:/usr/share/nginx/html
    depends_on:
      - app-dev
      - keycloak
```
- go to `http://localhost:3000`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@ and this project adheres to
 
 ⚡️(frontend) reduce unblocking time for config #867
 
+## Fixed
+
+- 🐛(nginx) fix 404 when accessing a doc #866
+
 ## [3.1.0] - 2025-04-07
 
 ## Added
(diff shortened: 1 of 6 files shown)


lunika:
I think it can also tested using tilt.

lunika (approved):

PR #867, opened 2025-04-11
⚡️(frontend) reduce unblocking time for config

AntoLC:
## Purpose

We will serve the config from the cache if available in waiting for the config to be loaded.
It will remove the loading time for the config except when the config is not available in the cache.


rvveber (approved):
LGTM

PanchoutNathan (approved):

PR #869, opened 2025-04-14
🔒️(back) don't allow an owner to change or delete other owner accesses

lunika:
## Purpose

Owner accesses can not be modified or deleted from other owners. Only current owner can modify and delete its own access.

## Proposal

- [x] 🔒️(back) don't allow an owner to change or delete other owner accesses


lunika:
As discuss with @rl-83 the issue is not in the back application. The business logic implemented in the back application is the one expected. The front does not reflect the abilities return by the access API.

PR #870, opened 2025-04-14
📝(contributing) add signed commits paragraph

lunika:
## Purpose

Add a paragraph to inform that only signed commits are accepted now.

## Proposal

- [x] 📝(contributing) add signed commits paragraph


lunika:
Managed in #871

PR #874, opened 2025-04-15
🚸(backend) make document search on title accent-insensitive

sampaccoud:
## Purpose

We want the document search on title to be accent-insensitive.

## Proposal

This should work in both cases:
- search for "vélo" when the document title contains "velo"
- search for "velo" when the document title contains "vélo"

### DIFF
--- a/src/backend/core/api/filters.py
+++ b/src/backend/core/api/filters.py
@@ -1,5 +1,7 @@
 """API filters for Impress' core application."""
 
+import unicodedata
+
 from django.utils.translation import gettext_lazy as _
 
 import django_filters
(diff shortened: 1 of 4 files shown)


src/backend/core/api/filters.py
```diff
@@ -1,18 +1,73 @@
...
+# pylint: disable=abstract-method
+class Unaccent(Func):
+    """
+    PostgreSQL unaccent function wrapper for use in Django ORM queries.
+
+    This allows you to annotate a field using the unaccented version of a
+    text column, enabling accent-insensitive filtering.
+    """
+
+    function = "unaccent"
```
lunika, line 31: This function already exists in django https://docs.djangoproject.com/en/5.1/ref/contrib/postgres/lookups/#unaccent

src/backend/core/api/filters.py
```diff
@@ -1,18 +1,73 @@
...
+
 class DocumentFilter(django_filters.FilterSet):
     """
-    Custom filter for filtering documents.
+    Custom filter for filtering documents on title (accent and case insensitive).
     """
 
-    title = django_filters.CharFilter(
+    title = AccentInsensitiveCharFilter(
         field_name="title", lookup_expr="icontains", label=_("Title")
```
lunika, line 71: As I understand with the unaccent documentation you can do:
```suggestion
        field_name="title", lookup_expr="unaccent__icontains", label=_("Title")
```

did you try it ?

src/backend/core/api/filters.py
sampaccoud: True! Good catch!

src/backend/core/api/filters.py
sampaccoud: Good idea! 
And it works :pray:

lunika (approved):

PR #889, opened 2025-04-18
🔒(frontend) enhance file download security

PanchoutNathan:
## Purpose

Added a safety check for URLs in the FileDownloadButton component. Now, before opening a URL, it verifies if the URL is safe using the isSafeUrl function.
This prevents potentially unsafe URLs from being opened in a new tab.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+## Fixed 
+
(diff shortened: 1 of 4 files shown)


NathanVss (approved):
I approve now as it must pass as a hotfix. But imo we should add all the tests cases that clear the OWASP recommendations

lunika (approved):
Can you update the changelog please

PR #893, opened 2025-04-21
🔧(backend) Allow overwriting the data directory

mweinelt:
## Purpose

Deployments that don't rely on Docker should be given the option to use a different data directory.

## Proposal

Allow customization of the `DATA_DIR` through an environment variable of the same name.

If the environment variable is not set the behaviour remains the same as before.


mweinelt:
I don't think it makes sense to overwrite the data directory within the docker container. After all it is just a mountpoint and the storage volume gets mounted on top of it.

lunika (approved):
Hi @mweinelt 

Thank you for your contribution.
Can you rebase your branch on the main one please ?

PR #900, opened 2025-04-22
♻️(frontend) bind ui with ability access

AntoLC:
## Purpose

Some actions were not available in the frontend but allowed in the backend.

## Proposal

We bind the UI with the ability access coming from the backend.

### DIFF
--- a/src/frontend/apps/e2e/__tests__/app-impress/doc-export.spec.ts
+++ b/src/frontend/apps/e2e/__tests__/app-impress/doc-export.spec.ts
@@ -79,7 +79,7 @@ test.describe('Doc Export', () => {
       })
       .click();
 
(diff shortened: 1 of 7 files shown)


src/frontend/apps/e2e/__tests__/app-impress/doc-export.spec.ts
```diff
@@ -79,7 +79,7 @@ test.describe('Doc Export', () => {
       })
       .click();
 
-    await page
+    void page
```
lunika, line 82: I didn't kno it is possible to do that. It is not the same to remove `await` without `void` ? What means the usage of `void` here ?

lunika (approved):

PR #911, opened 2025-04-28
🚩(frontend) version MIT only

AntoLC:
## Purpose

We have some packages that are not MIT compatible, so if the env var `PUBLISH_AS_MIT` is set to true, we don't build the application with features that are not MIT compatible.
For the moment, it concerns only the export packages.

## Proposal

Set the environement variable:
`PUBLISH_AS_MIT=false` -> will include the export feature

- [x] 🚩(frontend) version MIT only

## Demo

Thanks to the treeshacking, the packages of the features not included will not be included as well.
`toReactPDFDocument` is a method from `@blocknote/xl-pdf-exporter`, we will search the method in the js files of the app:

With `PUBLISH_AS_MIT=false`
image

With `PUBLISH_AS_MIT=true`
image


virgile-dev:
Hey! Keep in mind that this has to work for all XL packages that have AGPL licences that we will want to include.

bzg:
The name `MIT_only` is ambiguous: does it mean you only want MIT licensed libs? or you want to publish under MIT?

The "only" seems to indicate you only want MIT-licensed lib, which I guess is not what you really want.

I suggest `PUBLISH_AS_MIT` to capture the question "Do you want to publish (your build) as MIT or not?", leaving the scenarios open in case the answer is "No" (the user can then publish as AGPL or under proprietary terms.)

AntoLC:
> The name `MIT_only` is ambiguous: does it mean you only want MIT licensed libs? or you want to publish under MIT?
> 
> The "only" seems to indicate you only want MIT-licensed lib, which I guess is not what you really want.
> 
> I suggest `PUBLISH_AS_MIT` to capture the question "Do you want to publish (your build) as MIT or not?", leaving the scenarios open in case the answer is "No" (the user can then publish as AGPL or under proprietary terms.)

I changed it to `PUBLISH_AS_MIT`.

By default `PUBLISH_AS_MIT` is at true, meaning without the export feature. https://github.com/suitenumerique/docs/blob/3693f0c6382a087944f2ce31f14e60bedfbcaaee/src/frontend/apps/impress/.env#L3
It will have to be set to false explicitly before building the image in order to have the export feature.
https://github.com/suitenumerique/docs/blob/3693f0c6382a087944f2ce31f14e60bedfbcaaee/.github/workflows/docker-hub.yml#L84

Could you give a hand you think about how to document this env variable in our documentation ?

bzg:
While attempting to find my way through documenting `PUBLISH_AS_MIT`, I've opened [this PR](https://github.com/suitenumerique/docs/pull/941) for minor fixes against `docs/env.md` and [this issue](https://github.com/suitenumerique/docs/issues/942) for some other env variables.

bzg (approved):

lunika (approved):

Issue #909, opened 2025-04-28
Add support of `_FILE` environment variables for secrets


soyouzpanda:
## Feature Request

Add support of `_FILE` environment variables for secrets.

As discussed here: https://github.com/NixOS/nixpkgs/pull/401798#discussion_r2061087805, we need to pass the secrets, like `DJANGO_SECRET_KEY`, `COLLABORATION_SERVER_SECRET`, etc. in files.
The current solution is to wrap the gunicorn/celery/collaboration server in a script that export environment variables after `cat`-ing the files.

But we think that the support of secrets in files should be instead supported by docs, through environment variables like `DJANGO_SECRET_KEY_FILE`, `COLLABORATION_SERVER_SECRET_FILE`, etc.

mweinelt:
Passing them around as files is more compatible with the [systemd credential API](https://systemd.io/CREDENTIALS/) and can rely on the filesystem discretionary access control, which is what we want to use in NixOS.

PR #912, opened 2025-04-28
✨(backend) support `_FILE` environment variables for secrets

soyouzpanda:
Related to https://github.com/suitenumerique/docs/issues/909

### DIFF
--- a/src/backend/impress/settings.py
+++ b/src/backend/impress/settings.py
@@ -19,6 +19,7 @@ from django.utils.translation import gettext_lazy as _
 import sentry_sdk
 from configurations import Configuration, values
 from csp.constants import NONE
+from lasuite.configuration.values import SecretFileValue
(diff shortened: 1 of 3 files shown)


src/backend/impress/settings.py
```diff
@@ -37,6 +37,34 @@ def get_release():
     except (FileNotFoundError, KeyError):
         return "NA"  # Default: not available
 
+class SecretFileValue(values.Value):
```
lunika, line 40: tests are missing to validate this new class.
There is no docstring explaining how to use it
There is no documentation at all.

lunika:
I suggest to add this in https://github.com/suitenumerique/django-lasuite/

As you are doing nix packages for all projects, you will need it everywhere. Instead copy/pasting it, do it in the `django-lasuite` package.

Thanks

PR #927, opened 2025-05-02
🚧(resource-server) Allow documents API via RS

qbey:
## Purpose

This provides a base configuration to allow to access all "/documents" API via OIDC resource server authentication.

## Proposal

- [x] add new "resource server" API endpoints (instead of allowing the whole API to be authenticated via resource server)
- [ ] requires https://github.com/suitenumerique/django-lasuite/pull/16 to be merged and release before being able to use with this version of keycloak


virgile-dev:
Implemented here #1923

PR #932, opened 2025-05-04
📝(doc)  Update env.md add AI_FEATURE_ENABLED

rielzzapps:
📝(doc) added AI_FEATURE_ENABLED

This is false by default. Without this env setting on true AI will not be available in the docs application. The setting was missing in the env options.

## Purpose

Complete documentation


AntoLC (approved):
Thank you for your contribution !

PR #936, opened 2025-05-05
django-lasuite malware detection integration

lunika:
## Purpose

The django lasuite library has a dedicated module managing malware detection.
We have to configure it using django settings and then put the logic we want to apply in a callback.
While the analyse is not made, we set a temporary status to the file metadata and this status is check in the media-auth endpoint to determine is the s3 signature should be return or not.
If a file is glas as unsafe, it is deleted.

## Proposal

- [x] ✨(backend) force loading celery shared task in libraries
- [x] ✨(backend) configure lasuite.malware_detection module
- [x] ✨(backend) manage uploaded file status and call to malware detection

### DIFF
--- a/src/backend/core/malware_detection.py
+++ b/src/backend/core/malware_detection.py
@@ -1,0 +1,51 @@
+"""Malware detection callbacks"""
+
+import logging
+
+from django.core.files.storage import default_storage
+
+from lasuite.malware_detection.enums import ReportStatus
+
+from core.enums import DocumentAttachmentStatus
+from core.models import Document
+
+logger = logging.getLogger(__name__)
+
+
+def malware_detection_callback(file_path, status, error_info, **kwargs):
+    """Malware detection callback"""
+
+    if status == ReportStatus.SAFE:
+        logger.info("File %s is safe", file_path)
+        # Get existing metadata
+        s3_client = default_storage.connection.meta.client
+        bucket_name = default_storage.bucket_name
+        head_resp = s3_client.head_object(Bucket=bucket_name, Key=file_path)
+        metadata = head_resp.get("Metadata", {})
+        metadata.update({"status": DocumentAttachmentStatus.READY})
+        # Update status in metadata
+        s3_client.copy_object(
+            Bucket=bucket_name,
+            CopySource={"Bucket": bucket_name, "Key": file_path},
+            Key=file_path,
+            ContentType=head_resp.get("ContentType"),
+            Metadata=metadata,
+            MetadataDirective="REPLACE",
+        )
+        return
+
+    document_id = kwargs.get("document_id")
+    logger.error(
+        "File %s for document %s is infected with malware. Error info: %s",
+        file_path,
+        document_id,
+        error_info,
+    )
+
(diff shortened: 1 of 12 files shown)


securitykernel:
IMHO the best way to add malware scanning would be to implement an [ICAP](https://blogs.oracle.com/cloudsecurity/post/using-icap-to-scale-anti-malware-protection-for-saas-services) client that forwards files to an ICAP server that the admin can configure. This is the way most services, e.g., Nextcloud, do it.

lunika:
This part of the library can have many implementation. Maybe it is possible to add an ICAP backend alongside the other ones ?

lunika:
See this documentation: https://github.com/suitenumerique/django-lasuite/blob/3538e64546c769682ea8315ca95cb0da2a966efc/documentation/how-to-use-malware-detection-backend.md

src/backend/core/malware_detection.py
```diff
@@ -0,0 +1,51 @@
...
+        )
+        return
+
+    document_id = kwargs.get("document_id")
+    logger.error(
+        "File %s for document %s is infected with malware. Error info: %s",
+        file_path,
+        document_id,
+        error_info,
+    )
```
qbey, line 43: I wonder if this is the proper logger level ? The dev team should act on something ? 
Also, should we use a dedicated logger name for this ? Like `django.security` so we can use it to send logs to another destination ?

src/backend/impress/settings.py
```diff
@@ -680,6 +681,21 @@ class Base(Configuration):
         },
     }
 
+    MALWARE_DETECTION = {
+        "BACKEND": values.Value(
+            "lasuite.malware_detection.backends.dummy.DummyBackend",
```
qbey, line 686: I'm not sure I like this, because by default, file will be marked as "safe" which is not the case.

lunika: We can only provide the dummy backend by default. JCOP needs credentials.

src/backend/core/malware_detection.py
lunika: No the dev team can't act on something. You suggest to change it to info ?
We can use a dedicated name yes. `docs.security` ?

qbey:
```suggestion
    security_logger = logging.getLogger(f"docs.security.{__name__}")  # in case we want to add precise logging handler

    security_logger.warning(
        "File %s for document %s is infected with malware. Error info: %s",
        file_path,
        document_id,
        error_info,
    )
```

Warning because it is not a bug, but we might keep an eye on it. 
What do you think?

src/backend/impress/settings.py
qbey: Ok, since the "analyze status" is not stored anywhere this is not an issue: the status is only used by the callback to upload the file. I guess you can keep it like this then :)

src/backend/core/malware_detection.py
qbey: I'm not sure about the logger name... maybe you can propose something better ^^ or just remove the `__name__` which might not be that helpful actually.

qbey (approved):

PR #943, opened 2025-05-06
🔖(patch) release 3.2.1

AntoLC:
## Fixed

- 🐛(frontend) fix list copy paste #943

### DIFF
--- a/src/frontend/package.json
+++ b/src/frontend/package.json
@@ -34,6 +34,7 @@
     "@typescript-eslint/eslint-plugin": "8.32.0",
     "@typescript-eslint/parser": "8.32.0",
     "eslint": "8.57.0",
+    "prosemirror-model": "1.25.0",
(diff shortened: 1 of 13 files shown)


src/frontend/package.json
```diff
@@ -34,6 +34,7 @@
     "@typescript-eslint/eslint-plugin": "8.32.0",
     "@typescript-eslint/parser": "8.32.0",
     "eslint": "8.57.0",
+    "prosemirror-model": "1.25.0",
```
lunika, line 37: It will be updated next week. Is it what you want ?

AntoLC: Do you mean we should add it in `renovate` ? Yes you're right I think better to do that.

lunika: Yes I mean, should we add it to renovate indeed

lunika (approved):

PR #948, opened 2025-05-07
✨(back) allow theme customnization using a configuration file

lunika:
## Purpose

We want to customize the theme by using a configuration file. This configuration file path can be defined using the settings THEME_CUSTOMIZATION_FILE_PATH. If this file does not exists or is an invalid json, an empty json object will be added in the config endpoint.

## Proposal

- [x] ✨(back) allow theme customnization using a configuration file
- [x] remove footer endpoint
- [x] cache config endpoint
- [ ] documentation


configuration/theme/demo.json
AntoLC, line 1: This one is just for the POC with helm if I follow correctly, we are using `/src/backend/impress/configuration/theme/default.json` by default.

Should we move `/configuration/theme/demo.json` to `/src/helm/env.d/configuration/theme/demo.json` so ? 
Is there a reason this folder is at the root of the project ?

src/backend/impress/settings.py
```diff
@@ -891,6 +891,11 @@ class Production(Base):
             "OPTIONS": {
                 "CLIENT_CLASS": "django_redis.client.DefaultClient",
             },
+            "KEY_PREFIX": values.Value(
```
AntoLC, line 900:
```suggestion
            "CACHES_KEY_PREFIX": values.Value(
```

configuration/theme/demo.json
lunika: No you are right we can move it. It is here because in my first implementation I used this file in the backend too.

Dockerfile
```diff
@@ -87,6 +87,9 @@ RUN wget https://svn.apache.org/repos/asf/httpd/httpd/trunk/docs/conf/mime.types
 # Copy entrypoint
 COPY ./docker/files/usr/local/bin/entrypoint /usr/local/bin/entrypoint
 
+# Copy configuration
+VOLUME [ "/configuration" ]
```
AntoLC, line 91: Is it correct, the configuration is not to the root anymore ?

docker-compose.yml
```diff
@@ -72,6 +72,7 @@ services:
     volumes:
       - ./src/backend:/app
       - ./data/static:/data/static
+      - ./configuration:/configuration
```
AntoLC, line 75: Same, the configuration is not to the root anymore.

AntoLC (approved):
To have the test e2e to pass, here:
https://github.com/suitenumerique/docs/blob/0a4ec404a3cbda12168358d389e7c762e53cc5a4/src/frontend/apps/e2e/__tests__/app-impress/common.ts#L10

You can replace it with:
```diff
-  FRONTEND_FOOTER_FEATURE_ENABLED: true,
+  theme_customization: {},
```

Dockerfile
lunika: Yes we can remove it indeed

PR #949, opened 2025-05-07
♻️ (frontend) do multiple refacto

Zorin95670:
## Purpose

- Simplify if statement 
- Simplify redundant conditional expression
- Remove unnecessary local variable
- Remove var usage instead of let/const
- Add explicit React import to ensure JSX compatibility
- Add test file to improve coverage
- Add documentation

### DIFF
--- a/src/frontend/servers/y-provider/jest.config.js
+++ b/src/frontend/servers/y-provider/jest.config.js
@@ -1,4 +1,4 @@
-var config = {
+const config = {
   rootDir: './__tests__',
   testEnvironment: 'node',
   transform: {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to
 ## Changed
 
 - 📝(frontend) Update documentation
+- ✅(frontend) Improve tests coverage
 
 ## [3.2.1] - 2025-05-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to
 - ✨ Add a custom callout block to the editor #892
 - 🚩(frontend) version MIT only #911
 
+## Changed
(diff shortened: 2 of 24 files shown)


src/frontend/servers/y-provider/jest.config.js
```diff
@@ -1,4 +1,4 @@
-var config = {
+const config = {
```
fflorent, line 1: This can be enforced by the use of this eslint rule BTW:
https://eslint.org/docs/latest/rules/no-var

Zorin95670: Add a commit that modify configuration to include root file.

src/frontend/apps/impress/src/components/BoxButton.tsx
```diff
@@ -1,4 +1,4 @@
-import { forwardRef } from 'react';
+import React, { forwardRef } from 'react';
```
AntoLC, line 1: We pin to "react19.1.0", so adding "React" is not necessary, same for the others.

https://github.com/suitenumerique/docs/blob/bd79f84e070ebb42212859be1bbd7229c0ecbba2/src/frontend/package.json#L38

AntoLC (commented):
Hi, thank you for your contribution !

We try to keep meaningful commits in the history, could you squash them to have I think 3 commits ?
- one about minor refactoring (condition etc..)
- one about adding documentation
- one about adding tests

Same about the changelog, only meaningful entries, maybe 1 about documentation and 1 about coverage improvement should be enough, wdyt ?

Thank you.

src/frontend/apps/impress/src/components/BoxButton.tsx
Zorin95670: Remove commit that make this change!

PR #959, opened 2025-05-12
🏗️(frontend) footer configurable

AntoLC:
## Purpose

We need to have different footer per instance. 
The footer has to be displayable or not.

## Proposal

Bind the frontend footer with the theme configuration file: 
https://github.com/suitenumerique/docs/blob/main/src/backend/impress/configuration/theme/default.json

- The footer is configured from a json to be easily overwritable
- The legal pages are removed from the codebase, for our instances they will be stored directly in Docs, as a doc.

----

- [x] 🏗️(frontend) Footer configurable
- [x] 🔥(frontend) remove legal pages

## Demo

footer-configurable


docs/theming.md
```diff
@@ -30,4 +30,27 @@ body {
...
+
+### Settings 🔧
+
+```shellscript
+THEME_CUSTOMIZATION_FILE_PATH=<path>
+```
+
+### Example of JSON
+
+The json must follow some rules: https://github.com/suitenumerique/docs/blob/main/src/backend/impress/configuration/theme/default.json
```
rvveber, line 47: To illustrate the rules, can we point it to the demo.json in helm directory instead of the default.json?

Some customizations, like the translations do not make sense to be in default.json (yet).
Instead i've put them in demo.json to illustrate how it is used. 
If the user views this documentation, maybe it is better to link to the same files

AntoLC: Yes make sense.

src/backend/impress/configuration/theme/default.json
```diff
@@ -1,124 +1,129 @@
...
-            "label": "licence etalab-2.0",
-            "href": "https://github.com/etalab/licence-ouverte/blob/master/LO.md"
-          }
-        }
+  "footer": {
+    "default": {
+      "logo": {
+        "src": "/assets/icon-docs.svg",
+        "width": "54px",
+        "alt": "Docs Logo",
```
rvveber, line 7: How about we consider a more direct approach for translating strings? 

If we used translation slugs instead of plan text:

1. It could simplify things by removing the need for `default`, `fr`, `en` and similar sub-keys. 

2. It would also alleviate the burden of updating the full theme configuration when a new language is added, a process currently more involved than maintaining a flat JSON for translations. 

3. This shift would also enhance modularity and better separate responsibilities.

4. It would allow translators of crowdin to properly translate the default values.

5. It would reduce the size of the theming file significantly - potentially improving payload transfer speed and overall reactivity of the app

6. It would allow nested translations to be integrated in the future, further reducing complexity and improving flexibility

rvveber (changes requested):

lunika (approved):

src/frontend/apps/impress/src/features/footer/Footer.tsx
```diff
@@ -16,9 +20,42 @@ const BlueStripe = styled.div`
...
+      langData && 'externalLinks' in langData
+        ? langData?.externalLinks
+        : footerJson?.default?.externalLinks;
+    setExternalLinks(externalLinks);
+
+    const bottomInformation =
+      langData && 'bottomInformation' in langData
+        ? langData?.bottomInformation
+        : footerJson?.default?.bottomInformation;
+    setBottomInformation(bottomInformation);
```
PanchoutNathan, line 53: Why not make an object of type "Content Type" directly instead of making 3/4 states

src/frontend/apps/impress/src/features/footer/Footer.tsx
```diff
@@ -102,66 +138,59 @@ export const Footer = () => {
...
+            $margin={{ top: 'big' }}
+            $variation="600"
+            $display="inline"
+            className="--docs--footer-licence"
           >
-            <Text $variation="600">licence etalab-2.0</Text>
-            <IconLink width={18} />
-          </StyledLink>
-        </Text>
+            {bottomInformation.label}{' '}
```
PanchoutNathan, line 177: not a big to add a space like this

src/frontend/apps/impress/src/features/footer/Footer.tsx
```diff
@@ -102,66 +138,59 @@ export const Footer = () => {
...
               >
-                {label}
-              </Text>
-            </StyledLink>
-          ))}
+                <Text
+                  $variation="600"
+                  $size="m"
+                  $transition="box-shadow 0.3s"
+                  $css={`
```
PanchoutNathan, line 157: use 
```suggestion
$css={css`...`}
````

File: src/frontend/apps/impress/src/features/footer/Footer.tsx [hunk 6]
```diff
@@ -102,66 +138,59 @@ export const Footer = () => {
...
-                $transition="box-shadow 0.3s"
+          {legalLinks &&
+            legalLinks.map(({ label, href }) => (
+              <StyledLink
+                key={label}
+                href={href}
                 $css={`
-                  &:hover {
-                    box-shadow: 0px 2px 0 0 var(--c--theme--colors--greyscale-text);
+                  padding-right: 1rem;
```
PanchoutNathan, line 147: ditto

src/frontend/apps/impress/src/features/footer/Footer.tsx
```diff
@@ -102,66 +138,59 @@ export const Footer = () => {
...
-            <Text $variation="600">licence etalab-2.0</Text>
-            <IconLink width={18} />
-          </StyledLink>
-        </Text>
+            {bottomInformation.label}{' '}
+            {bottomInformation.link && (
+              <StyledLink
+                href={bottomInformation.link.href}
+                target="__blank"
+                $css={`
```
PanchoutNathan, line 182: ditto

PanchoutNathan (approved):

src/backend/impress/configuration/theme/default.json
AntoLC: Until we cannot override translations per instance we cannot do this way. 
But I think you're right, we should refacto this part as soon we can override translations, I will open an issue about it.

src/frontend/apps/impress/src/features/footer/Footer.tsx
AntoLC: It is rewrote like that by prettier.

src/backend/impress/configuration/theme/default.json
rvveber: Saved for later: #994

PR #961, opened 2025-05-12
Feature/query caching and debug flags

rvveber:
Is based on https://github.com/suitenumerique/docs/pull/857 - (merge first)

Perfomance Improvement: Adds query caching that is persisted across browser restarts
Developer Exprience Improvement: Adds and documents usage of debug flags


src/backend/impress/settings.py
```diff
@@ -58,7 +58,14 @@ class Base(Configuration):
...
     """
 
-    DEBUG = False
+    DEBUG_RAW = values.Value(
+        default=False,
+        environ_name="DEBUG",
+        environ_prefix=None,
+    )
+    DEBUG_NAMESPACES = DEBUG_RAW.split(",") if isinstance(DEBUG_RAW, str) else []
+    DEBUG = bool(DEBUG_RAW)
```
lunika, line 67: I'm not really confident with this and It's I think error prone.

src/frontend/apps/impress/src/core/AppProvider.tsx
```diff
@@ -13,12 +17,16 @@ import { ConfigProvider } from './config/';
...
+ *        - instant if debug flag "no-cache" active - 3 minutes otherwise
+ *    - gcTime:
+ *      - global time until cache is purged from the persister and needs to be renewed
+ *        - since its cached in localStorage, we can set it to a long time (48h)
  */
 const defaultOptions = {
   queries: {
-    staleTime: 1000 * 60 * 3,
+    staleTime: debug.enabled('no-cache') ? 0 : 1000 * 60 * 3, // 3 minutes
+    gcTime: debug.enabled('no-cache') ? 0 : 1000 * 60 * 60 * 48, // 48 hours
```
AntoLC, line 29: About the debug, I am not sure it is the best way to do. 
As a developer I like to be near the production setting, it helps me in this case to see which query I have to invalidate. If I have `no-cache`, I will not know that when I change the **title of a document** I need to invalidate the cache of the **document list** request.

After if you don't want cache, we could set these variables from settings:
```suggestion
    staleTime: process.env.NEXT_PUBLIC_CACHE_STALE_TIME,
    gcTime: process.env.NEXT_PUBLIC_CACHE_GC_TIME
```
By default they will be in `.env`:
```
NEXT_PUBLIC_CACHE_STALE_TIME=1000 * 60 * 3
NEXT_PUBLIC_CACHE_GC_TIME=1000 * 60 * 60 * 48
```
but in `.env.local` you could set them as you want, `.env.local` is not versioned:
```
NEXT_PUBLIC_CACHE_STALE_TIME=0
NEXT_PUBLIC_CACHE_GC_TIME=0
```

src/backend/impress/settings.py
```diff
@@ -761,6 +768,9 @@ def post_setup(cls):
                 "OIDC_ALLOW_DUPLICATE_EMAILS cannot be set to True simultaneously. "
             )
 
+        if cls.DEBUG and "no-cache" in cls.DEBUG_NAMESPACES:
+            cls.THEME_CUSTOMIZATION_CACHE_TIMEOUT = 0
```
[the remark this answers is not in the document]
AntoLC: Yes in `env.d\development\common`.

AntoLC (commented):
It is not working as expected, it keeps all the requests in cache, by doing like that:
- the logout is not working anymore
- the doc list is in cache so if someone invite you or send you a public link, it will not appear anymore in the doc list (until the cache is invalidate)
- I guess all the requests that change regularly between multiple users will be in cache, giving lot of side effect during collaboration

What could be interesting is to have by default all the requests not in cache (as we have already), and to decide the ones that we want to have with a persistant cache, like `useConfig`. 
⚠️ After we have a subtility with `useConfig`, it has a persistant cache, but we still do the request in background, if the response has some changes, it will rehydrate the components with the new changes.

src/backend/impress/settings.py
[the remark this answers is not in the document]
rvveber: As long as it is only theme customization we are caching.
But what if there will be more in the future?

This was more to illustrate how the debug namespaces would work.

src/backend/impress/settings.py
rvveber: Agree, it is.

Arguably we can remove the extended DEBUG handling for the backend.

I see too options:
A: Remove the debug handling for the backend
B: Write a dedicated function to handle it solidly

Which to choose?

src/frontend/apps/impress/src/pages/_app.tsx
```diff
@@ -18,6 +18,28 @@ export default function App({ Component, pageProps }: AppPropsWithLayout) {
...
+     *    via Local storage
+     *      window.localStorage.debug = "features:language,no-cache,...";
+     *
+     *    via code (uses Local storage)
+     *      import debug from 'debug';
+     *      debug.enable('no-cache,features:language,...');
+     */
+    //debug.enable('no-cache,features:language');
+  }
+
```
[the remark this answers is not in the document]
rvveber: It is just here to illustrate usage for future development and enable quick toggle from code.

src/frontend/apps/impress/src/core/AppProvider.tsx
rvveber: Thanks for the feedback. 

1.  **Debug Mode and Cache:** The `debug` flag for cache control is off by default, aligning with production behavior. Developers who enable it for local debugging are intentionally altering cache settings for temporary diagnostic purposes.

2.  **Data Updates and Cache Invalidation:** For data modifications (e.g., updating a document title), we should adhere to the mutation pattern ([as demonstrated here](https://github.com/suitenumerique/docs/pull/857#discussion_r2098098188)). This ensures that related queries are correctly invalidated and fresh data is fetched.

3.  **Configurability of `staleTime` and `gcTime`:**
    * Making the default `staleTime` and `gcTime` configurable via environment variables could introduce complexity. Because each `useQuery` can override defaults - we would need to add env configuration for each `useQuery` that has a different `staleTime` and `gcTime`.
    * Generally, we, as developers designing the queries, are best positioned to determine appropriate `staleTime` and `gcTime` values based on the nature of the data. Allowing arbitrary instance-level customization or overly flexible environment configurations might lead to inconsistent caching behavior and unexpected issues. It's preferable to establish sensible, well-reasoned defaults within the codebase for different types of queries.

rvveber:
> It is not working as expected, it keeps all the requests in cache, by doing like that:
> 
> * the logout is not working anymore
> * the doc list is in cache so if someone invite you or send you a public link, it will not appear anymore in the doc list (until the cache is invalidate)
> * I guess all the requests that change regularly between multiple users will be in cache, giving lot of side effect during collaboration
> 
> What could be interesting is to have by default all the requests not in cache (as we have already), and to decide the ones that we want to have with a persistant cache, like `useConfig`. ⚠️ After we have a subtility with `useConfig`, it has a persistant cache, but we still do the request in background, if the response has some changes, it will rehydrate the components with the new changes.

Thanks for the catch!
It was naive of me thinking it would be a drop-in replacement.
There are lot of cases that need extensive testing and adaption.

To the points you found i will address the following additional points in the next days:
* add cache busting with `process.env.NEXT_PUBLIC_BUILD_ID` - to avoid using cache from old builds after updates.
* exempt the authQuery from beeing persisted
* test all existing queries 
* ensure all mutations that access queries properly invalidate the query

AntoLC:
I close it, I added tags to find it easily again. 
Feel free to open it if you want to work on it again.

PR #977, opened 2025-05-16
✨(backend) allow setting session cookie age via env var

sampaccoud:
## Purpose

We want to be able to increase the duration of the cookie session.

## Proposal

Add django-configuration IntegerValue to expose the `SESSION_COOKIE_AGE` as an environment variable.

### DIFF
--- a/src/backend/impress/settings.py
+++ b/src/backend/impress/settings.py
@@ -462,7 +462,9 @@ class Base(Configuration):
     # Session
     SESSION_ENGINE = "django.contrib.sessions.backends.cache"
     SESSION_CACHE_ALIAS = "default"
-    SESSION_COOKIE_AGE = 60 * 60 * 12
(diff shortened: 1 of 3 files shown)


src/backend/impress/settings.py
```diff
@@ -462,7 +462,9 @@ class Base(Configuration):
     # Session
     SESSION_ENGINE = "django.contrib.sessions.backends.cache"
     SESSION_CACHE_ALIAS = "default"
-    SESSION_COOKIE_AGE = 60 * 60 * 12
+    SESSION_COOKIE_AGE = values.IntegerValue(
```
lunika, line 465:
```suggestion
    SESSION_COOKIE_AGE = values.PositiveIntegerValue(
```

This is a nitpick, you can decide to leave it as it.

lunika (approved):

PR #991, opened 2025-05-21
✨(back) add endpoint cheking media status

lunika:
## Purpose

With the usage of a malware detection system, we need a way to know the
file status. The front will use it to display a loader while the analyse
is not ended.

Also, we want to have the media-check url returned on the attachment-upload
response instead of the media url directly. The front will know the
endpoint to use to check the media status.

## Proposal

- [x] ✨(back) add endpoint cheking media status
- [x] ♻️(back) return the media-check url on the attachment_upload response


lunika:
Commits are managed in #984

Issue #928, opened 2025-05-02
Keep session cookie during 5 days


virgile-dev:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
Users complain they have to login everyday.
Right now we keep the session for only 12 hours : https://github.com/suitenumerique/docs/blob/main/src/backend/impress/settings.py#L466

**Describe the solution you'd like**
Let's keep the session cookie for 5 days.

lunika:
Managed in https://github.com/suitenumerique/docs/pull/977

PR #993, opened 2025-05-21
✨ Keep session cookie during 5 days

ZouicheOmar:
This PR adresses issue #928.
## Purpose
Extend the user session to last 5 days

## Proposal
Add `django-lasuite` session refresh, which already implements
OIDC token refresh and default `SESSION_COOKIE_AGE` to 5 days.

- [x] ✨(backend) extend user session to 5 days
- [ ] 📝(doc) update session refresh configuration

### DIFF
--- a/src/backend/core/authentication/backends.py
+++ b/src/backend/core/authentication/backends.py
@@ -2,6 +2,7 @@
 
 import logging
 import os
+import time
 
 from django.conf import settings
 from django.core.exceptions import SuspiciousOperation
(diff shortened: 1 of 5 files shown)


ZouicheOmar:
As `django-lasuite` already implements a token refresh middleware, thought it would be a good fit, but with a couple changes; the middleware decides to do a refresh request by checking againts a `oidc_token_expiration` field that is supposed to be in the request's associated session, so I've added it in `OIDCAuthenticationBackend`. Second change is by updating the `oidc_token_expiration` if a refresh is successful, and that is to prevent a refresh request to be sent on each consequent user request, which could be overkill, maybe and unecessary roundtrip and could probably cause bugs.

PR #997, opened 2025-05-23
💄Fine tuning 3.3.0

AntoLC:
## Purpose

- [x] 🚩(frontend) feature flag on blocking edition
- [x] ✏️(frontend) change antivirus fail sentence

## New env var
```
COLLABORATION_ONLY_CONNECTED_CAN_EDIT=[boolean]
```
If users were not connected to the collaboration server, they were not be able to edit documents.
We decided to add a feature flag on this feature as it can be quite restrictive.
We can now enable or disable this feature at runtime thanks to the env variable `COLLABORATION_ONLY_CONNECTED_CAN_EDIT`.


src/frontend/apps/e2e/__tests__/app-impress/common.ts
```diff
@@ -4,6 +4,7 @@ export const CONFIG = {
   AI_FEATURE_ENABLED: true,
   CRISP_WEBSITE_ID: null,
   COLLABORATION_WS_URL: 'ws://localhost:4444/collaboration/ws/',
+  COLLABORATION_ONLY_CONNECTED_CAN_EDIT: false,
```
lunika, line 7: `ONLY_CONNECTED` is confusing. I understood first only user being connected to the application. I would suggestion `COLLABORATION_ACTIVATED_WITHOUT_WS_CONNECTION` ? Is it more accurate ?

lunika (approved):

PR #1008, opened 2025-05-24
Track actions using .capture()

lebaudantoine:
## Purpose

Avoid relying on the autocapture, isntead track specific events.

## Proposal

Track : 
- AI action and performance / error rate data
- Doc creation from the left panel 
- Open search modal

Feel free to suggest any idea.


virgile-dev:
Here are things I'd like to track. Ideally I'd want to have the % of users / docs who use this features
- Add pinned document
- Export a document
- Check version history
- Use of filtered list of documents (My Docs and Shared Docs)
- Click on the Summary 
- Invite a user to a doc
- Use of link share (public or connected)

PR #1033, opened 2025-06-02
🔧(docker) add a service in compose to frontend development

lunika:
## Purpose

We want a serice in compose starting the frontend application in development mode. We want to take the advantage of the hot reload module, so the sources are mounted inside the container.

## Proposal

- [x] 🔧(docker) add a service in compose to frontend development

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,7 +128,7 @@ run-backend: ## Start only the backend application and all needed services
 run: ## start the wsgi (production) and development server
 run: 
 	@$(MAKE) run-backend
-	@$(COMPOSE) up --force-recreate -d frontend
+	@$(COMPOSE) up --force-recreate -d frontend-development
 .PHONY: run
 
 status: ## an alias for "docker compose ps"
@@ -315,7 +315,7 @@ frontend-lint: ## run the frontend linter
 .PHONY: frontend-lint
 
 run-frontend-development: ## Run the frontend in development mode
-	@$(COMPOSE) stop frontend
+	@$(COMPOSE) stop frontend frontend-development
 	cd $(PATH_FRONT_IMPRESS) && yarn dev
 .PHONY: run-frontend-development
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -128,7 +128,7 @@ run-backend: ## Start only the backend application and all needed services
 run: ## start the wsgi (production) and development server
 run: 
 	@$(MAKE) run-backend
-	@$(COMPOSE) up --force-recreate -d frontend
+	@$(COMPOSE) up --force-recreate -d frontend-development
```
Kharonus, line 156: Great work, can you amend the `build-frontend` make target, too, so that it builds both services?

AntoLC: Only the port 3000 is available, so starting both containers will cause problems.

docker-compose.yml
```diff
@@ -157,7 +157,23 @@ services:
...
+      context: .
+      dockerfile: ./src/frontend/Dockerfile
+      target: impress-dev
+      args:
+        API_ORIGIN: "http://localhost:8071"
+        PUBLISH_AS_MIT: "false"
+        SW_DEACTIVATED: "true"
     image: impress:frontend-development
+    volumes:
+      - ./src/frontend:/home/frontend
```
AntoLC, line 126: There is lot of side effects by keeping the `node_modules` sync, I think it is common pattern to exclude it:

```suggestion
      - ./src/frontend:/home/frontend
      - /home/frontend/node_modules
      - /home/frontend/apps/impress/node_modules
```

lunika:
@AntoLC @Kharonus I made other modification, the y-provider is run in dev mode and I removed all the services running application in production mode, we don't use them.

Makefile
Kharonus: it is not about running, but building. might be handy to have both images ready to go and they share a lot of the docker context, so no big overhead.

But I saw, that lunika now removed the previous frontend service completely and replaced it, so this is now obsolete :D

AntoLC (approved):
Looks good !

Issue #879, opened 2025-04-15
Structure of the <nav>


cyberbaloo:
**Problematic behavior**
The left column looks like a navigation, the items in this column should be in a `<nav>` element

The buttons in the left column should be structured in a `<ul><li>` list to allow screen reader users to know the number of elements present in the navigation.

**Possible Solution**
- Add a `role="navigation"` to the element `<nav>`
- Add a list `<ul><li>`

Example :

```
<ul>
<li><button>Tous les documents</button></li>
<li><button>Mes documents</button></li>
<li><button>Partagés avec moi</button></li>
</ul>
```

PR #1047, opened 2025-06-03
add role navigation and semantic list to left panel nav

Appryll:
Enhance accessibility: add role="navigation" and semantic list to left panel <nav>
Issue associée :  [879](https://github.com/suitenumerique/docs/issues/879)

### DIFF
--- a/src/frontend/apps/impress/src/features/left-panel/components/LefPanelTargetFilters.tsx
+++ b/src/frontend/apps/impress/src/features/left-panel/components/LefPanelTargetFilters.tsx
@@ -2,7 +2,7 @@ import { usePathname, useRouter, useSearchParams } from 'next/navigation';
 import { useTranslation } from 'react-i18next';
 import { css } from 'styled-components';
 
-import { Box, BoxButton, Icon, Text } from '@/components';
+import { BoxButton, Icon, Text } from '@/components';
 import { useCunninghamTheme } from '@/cunningham';
 import { DocDefaultFilter } from '@/features/docs';
 import { useLeftPanelStore } from '@/features/left-panel';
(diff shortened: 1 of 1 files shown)


src/frontend/apps/impress/src/features/left-panel/components/LefPanelTargetFilters.tsx
```diff
@@ -47,48 +47,50 @@ export const LeftPanelTargetFilters = () => {
...
   return (
-    <Box
-      $justify="center"
-      $padding={{ horizontal: 'sm' }}
-      $gap={spacingsTokens['2xs']}
-      className="--docs--left-panel-target-filters"
-    >
-      {defaultQueries.map((query) => {
-        const isActive = target === query.targetQuery;
+    <nav role="navigation" aria-label="Docs filters">
```
Ovgodd, line 50: Hello Thanks for the semantic update !
 You can remove role="navigation", as it’s implicit on the nav element. It’s only necessary if you're using a non-semantic container like a div.

and Also, consider updating the aria-label something like "Document sections" would be more meaningful for screen reader users than "Docs filters".

```suggestion
    <nav aria-label="Document sections">
```

src/frontend/apps/impress/src/features/left-panel/components/LefPanelTargetFilters.tsx
```diff
@@ -47,48 +47,50 @@ export const LeftPanelTargetFilters = () => {
...
+                $css={css`
+                  width: 100%;
+                  cursor: pointer;
+                  background-color: ${isActive
+                    ? colorsTokens['greyscale-100']
+                    : undefined};
+                  font-weight: ${isActive ? 700 : undefined};
+                  &:hover {
+                    background-color: ${colorsTokens['greyscale-100']};
+                  }
```
Ovgodd, line 79: There’s no visible focus outline, so it's hard to know where we are.

👉 You may need to explicitly add a :focus-visible style on the BoxButton to restore proper keyboard visibility.

```suggestion
                  }
&:focus-visible {
outline: 2px solid ${colorsTokens['primary-500']};
outline-offset: 2px;
border-radius: ${spacingsTokens['3xs']};
}
```
focus

src/frontend/apps/impress/src/features/left-panel/components/LefPanelTargetFilters.tsx
Ovgodd, line 50: The visual layout seems slightly broken after this change probably some missing spacing or margin on the filter items.  

sidebar

Ovgodd:
a new PR has been created : [1262](https://github.com/suitenumerique/docs/pull/1262)

PR #1050, opened 2025-06-03
Add Notion import

clarani:
:warning: THIS PR IS STILL A DRAFT :warning: 

:star2:  Contributors: @emersion @Tguisnet @castavo @NicolasRitouet @clarani
:computer:  Project of the Hack Days Hackathon

## Purpose
The goal of this feature is to allow users to import notion workspace into Docs in 1 clic.

The user can connect to Notion by clicking on a button, select the pages he/she wants to import and the pages are directly imported into Docs.
A progress bar displays the import progression.

A lot of Notion block types are handled:
- column
- heading
- text and text formatting
- checkbox
- bullet and numbered list
- child page
- callout
- images

The unsupported blocks are replaced by a textual warning. The database pages from Notion are ignored since Docs does not handle databases for the moment.

## For later
- The import is quite long since the Notion API is slow. An improvement would be to make the import run in the background.
- We don't not handle the import errors for the moment.


emersion:
Ref https://github.com/suitenumerique/docs/pull/1049 which adds custom blocks support to y-provider.

emersion:
Note, this depends on #701 for nested documents.

emersion:
Why has this been closed?

emersion:
Hm, seems like this is a side-effect of deleting the branch this was based on.

Will rebase and reopen later.

emersion:
Opened https://github.com/suitenumerique/docs/pull/1141 (GitHub won't let me re-open this PR.)

PR #1065, opened 2025-06-10
🩹(backend) default CORS_ALLOW_ALL_ORIGINS to False

lunika:
## Purpose

The settings CORS_ALLOW_ALL_ORIGINS was set to True by default.

This error is inherited from a old mistake made back in the days while working on the initial impress demo.

This is not something we want, this should be only allowed in development. We change the value in all the manifests in order to have the desired behavior in non development environments.

## Proposal

- [x] 🩹(backend) default CORS_ALLOW_ALL_ORIGINS to False


AntoLC (commented):
Better late than never ^^

What about this endpoint ?
https://github.com/suitenumerique/docs/blob/1edac41bf861428b79adb33673deacf6945c92d5/src/backend/core/api/viewsets.py#L671-L679

lunika:
> Better late than never ^^
> 
> What about this endpoint ?
> 
> https://github.com/suitenumerique/docs/blob/1edac41bf861428b79adb33673deacf6945c92d5/src/backend/core/api/viewsets.py#L671-L679

This endpoint is used in a server to server context. CORS is not used in that context

PR #1078, opened 2025-06-16
✨(frontend) Duplicate a doc

AntoLC:
## Purpose

Gives the possibility to duplicate a document.

## Proposal

- [x] ♻️(frontend) Simplify AGPL export pattern 
- [x] ✨(frontend) Duplicate a doc

## Demo

[Demo duplicate doc ! - Docs.webm](https://github.com/user-attachments/assets/924901ab-3c0b-437d-a3fe-96fe417a4080)

## Edge case

We save before duplication to ensure the latest state is duplicated.

⚠️ Possible data lost:
- if current user is reader, it will not save, so 1mn could be lost with duplication if editor are writing in the same moment.
- if duplication is trigger from option document list, we cannot access the provider, so 1mn could be lost with duplication if editor are writing in the same moment.

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -954,6 +954,8 @@ class DocumentViewSet(
         )
         serializer.is_valid(raise_exception=True)
         with_accesses = serializer.validated_data.get("with_accesses", False)
+        roles = set(document.get_roles(request.user))
+        is_owner_or_admin = bool(roles.intersection(set(models.PRIVILEGED_ROLES)))
 
         base64_yjs_content = document.content
 
@@ -985,7 +987,7 @@ class DocumentViewSet(
         ]
 
         # If accesses should be duplicated, add other users' accesses as per original document
-        if with_accesses:
+        if with_accesses and is_owner_or_admin:
             original_accesses = models.DocumentAccess.objects.filter(
                 document=document
             ).exclude(user=request.user)
--- a/src/backend/core/tests/documents/test_api_documents_duplicate.py
+++ b/src/backend/core/tests/documents/test_api_documents_duplicate.py
@@ -168,14 +168,17 @@ def test_api_documents_duplicate_success(index):
         assert response.status_code == 403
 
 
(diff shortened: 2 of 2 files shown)


virgile-dev:
hey @AntoLC good job ! 
can you also duplicate from the document list ?
@rl-83 do you think we should have an option to duplicate without the rights ?

AntoLC:
> can you also duplicate from the document list

Yes we can

virgile-dev:
Hey @AntoLC  @rl-83 talked with @sampaccoud about this.
It told me the back provides an option to duplicate with or without the rights. 
The UI should should give the option to the user.

virgile-dev:
> Hey @AntoLC @rl-83 talked with @sampaccoud about this. It told me the back provides an option to duplicate with or without the rights. The UI should should give the option to the user.

By default we shouldn't copy the associated members and rights. Let's remove this from this PR. I've opened an [issue](https://github.com/suitenumerique/docs/issues/1094) to add the mentioned above

src/frontend/apps/e2e/__tests__/app-impress/doc-header.spec.ts
```diff
@@ -456,6 +456,37 @@ test.describe('Doc Header', () => {
...
   });
+
+  test('it duplicates a document', async ({ page, browserName }) => {
+    const [docTitle] = await createDoc(page, `Duplicate doc`, browserName);
+
+    const editor = page.locator('.ProseMirror');
+    await editor.click();
+    await editor.fill('Hello Duplicated World');
+
+    await page.reload();
```
lunika: This does not reflect what really happen when you work on a document. This reload I think trigger the save of the document content, this is why you have the same content at the end of this test

AntoLC: It will now save if possible when duplication.
- if reader, it will not save, so 1mn could be lost with duplication if editor are writing in the same moment.
- if duplication is trigger from option document list, we cannot access the provider, so 1mn could be lost with duplication if editor are writing in the same moment.

PR #1079, opened 2025-06-16
⬆️(back) upgrade django to version 5.2

lunika:
## Purpose

Django 5.2 is now mature enough and we can use it in production. In some tests the number of sql queries is increasing. This is because the `full_clean` method called in the `save` method on all our models is creating a transaction, so a savepoint and release is added. We also fix deprecated warning in this commit.

## Proposal

- [x] ⬆️(back) upgrade django to version 5.2


src/backend/impress/settings.py
```diff
@@ -849,6 +849,7 @@ class Test(Base):
         "django.contrib.auth.hashers.MD5PasswordHasher",
     ]
     USE_SWAGGER = True
+    STATIC_ROOT = None
```
qbey, line 854: Why do you need this?

lunika: We have warning in the tests saying that the `/data/static` directory does not exists. We can create it or set to None the `STATIC_ROOT` variable as static are not used in the tests.

lunika:
```
core/tests/test_api_config.py::test_api_config_with_invalid_json_theme_customization_file[False]
core/tests/test_api_config.py::test_api_config_with_invalid_json_theme_customization_file[True]
core/tests/test_api_config.py::test_api_config_with_theme_customization[False]
core/tests/test_api_config.py::test_api_config_with_theme_customization[True]
  /home/runner/.local/lib/python3.13/site-packages/django/core/handlers/base.py:61: UserWarning: No directory at: /data/static/
    mw_instance = middleware(adapted_handler)
```

AntoLC (commented):
Some tests fails, `e2e` must be replaced in some places:
- https://github.com/suitenumerique/docs/blob/django-52/src/frontend/apps/e2e/__tests__/app-impress/doc-create.spec.ts#L47-L48
- https://github.com/suitenumerique/docs/blob/django-52/src/frontend/apps/e2e/__tests__/app-impress/doc-member-list.spec.ts#L134
- https://github.com/suitenumerique/docs/blob/django-52/src/frontend/apps/e2e/__tests__/app-impress/doc-member-list.spec.ts#L178
- https://github.com/suitenumerique/docs/blob/django-52/src/frontend/apps/e2e/__tests__/app-impress/doc-visibility.spec.ts#L154

PR #1085, opened 2025-06-19
➕(back) install and configure django csp

### LINKED ISSUE #1000: Install and configure library django-csp
## Feature Request

We want to install and configure the library [django-csp](https://django-csp.readthedocs.io/en/latest/). 

The default should be a strict rule, this rule must be configurable using django-configuration.

lunika:
## Purpose

We want to protect all requests from django with c content security policy header. We use the djang-csp library and configure it with default values.

## Proposal

- [x] ➕(back) install and configure django csp

Fixes #1000

### DIFF
--- a/src/backend/impress/settings.py
+++ b/src/backend/impress/settings.py
@@ -285,6 +285,7 @@ class Base(Configuration):
         "django.contrib.auth.middleware.AuthenticationMiddleware",
         "django.contrib.messages.middleware.MessageMiddleware",
         "dockerflow.django.middleware.DockerflowMiddleware",
+        "csp.middleware.CSPMiddleware",
     ]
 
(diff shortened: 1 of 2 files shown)


AntoLC (commented):
Do we still need that so ?
https://github.com/suitenumerique/docs/blob/c7261cf507f7d0a96947ffa20017b7a22bbd4b5b/src/backend/core/api/viewsets.py#L1455
https://github.com/suitenumerique/docs/blob/c7261cf507f7d0a96947ffa20017b7a22bbd4b5b/docker/files/etc/nginx/conf.d/default.conf#L24

AntoLC (approved):

lunika:
In NGINX yes because the response does not come from django but minio. 
For the other I will take a look.

PR #1093, opened 2025-06-25
Give priority to users connected to collaboration server (aka no websocket feature)

lunika:
## Purpose

Some users have no access to the websocket protocol and can't collaborate to a docs. This lead to inconsistent document content because they will override the content on every save.

With this PR we implement a behavior to prevent the loss of data.
Priority is given to users connected to the collaboration server, while users are connected to it. users without websocket can't save their content, no matter if a user without websocket came first on the document.
Also there is a priority between user without websocket. The first user saving a document acquire a lock and this lock remains active while the user is working on the document. After 2 minutes of inactivity, the user release the lock and an other user without a websocket can save its work.

## Proposal

- [x] ✨(y-provider) add endpoint returning document connection state
- [x] ✨(back) check on document update if user can save it


src/backend/impress/settings.py
```diff
@@ -470,6 +471,7 @@ class Base(Configuration):
     SESSION_COOKIE_AGE = values.PositiveIntegerValue(
         default=60 * 60 * 12, environ_name="SESSION_COOKIE_AGE", environ_prefix=None
     )
+    SESSION_COOKIE_NAME = "docs_sessionid"
```
qbey, line 484: note: every user will be disconnected on next deployment

src/backend/core/services/collaboration_services.py
```diff
@@ -41,3 +41,31 @@ def reset_connections(self, room, user_id=None):
...
+        except requests.RequestException as e:
+            raise requests.HTTPError("Failed to get document connection info.") from e
+
+        if response.status_code != 200:
+            raise requests.HTTPError(
+                f"Failed to get document connection info. Status code: {response.status_code}, "
+                f"Response: {response.text}"
+            )
+
+        return response.json()
```
qbey: Please add a comment (or a dataclass) to easily know the json is something like `{"count": int, "exists": bool}`?
I'm even wondering if the return of this method should not be directly `return count, exists`

src/backend/core/api/viewsets.py
```diff
@@ -631,6 +632,82 @@ def perform_destroy(self, instance):
         """Override to implement a soft delete instead of dumping the record in database."""
         instance.soft_delete()
 
+    def _compute_no_websocket_cache_key(self, document_id):
+        """Compute the cache key for the no websocket cache."""
+        return f"docs:no-websocket:{document_id}"
```
qbey: This is called only once, are you sure you need a dedicated method instead of an f-string where used?
(also this method should be static)

src/backend/core/api/viewsets.py
```diff
@@ -631,6 +632,82 @@ def perform_destroy(self, instance):
...
+
+    def _can_user_edit_document(self, document_id, set_cache=False):
+        """Check if the user can edit the document."""
+        try:
+            connection_info = CollaborationService().get_document_connection_info(
+                document_id,
+                self.request.session.session_key,
+            )
+        except requests.HTTPError as e:
+            capture_exception(e)
```
qbey:
```suggestion
            logger.exception("Failed to call collaboration server: ", e)
```
No?

src/backend/core/api/viewsets.py
```diff
@@ -631,6 +632,82 @@ def perform_destroy(self, instance):
...
+
+        return False
+
+    def perform_update(self, serializer):
+        """Check rules about collaboration."""
+        if serializer.validated_data.get("websocket"):
+            return super().perform_update(serializer)
+
+        if self._can_user_edit_document(serializer.instance.id, set_cache=True):
+            return super().perform_update(serializer)
```
qbey, line 691: You may want to raise a specific error content to display a proper message to the user

src/backend/core/api/viewsets.py
```diff
@@ -631,6 +632,82 @@ def perform_destroy(self, instance):
...
+
+        logger.debug(
+            "Users connected to the websocket but current editor not connected to it. Can not edit."
+        )
+
+        return False
+
+    def perform_update(self, serializer):
+        """Check rules about collaboration."""
+        if serializer.validated_data.get("websocket"):
```
qbey:
```suggestion
        if serializer.validated_data.get("websocket", False):
```
Just to be explicit with the default behavior (I know `None` is falsy, but...)

src/backend/core/middleware.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def __init__(self, get_response):
+        """Initialize the middleware."""
+        self.get_response = get_response
+
+    def __call__(self, request):
+        """Force session creation for unauthenticated users."""
+
+        if not request.user.is_authenticated and request.session.session_key is None:
+            request.session.save()
```
qbey: Why `save` instead of `create`?

Is there a test somewhere to check this middleware behavior?

src/backend/core/tests/documents/test_api_documents_can_edit.py
```diff
@@ -0,0 +1,219 @@
...
+
+pytestmark = pytest.mark.django_db
+
+
+def test_api_documents_can_edit_anonymous():
+    """Anonymous users can not edit documents."""
+    document = factories.DocumentFactory()
+    client = APIClient()
+    response = client.get(f"/api/v1.0/documents/{document.id!s}/can-edit/")
+    assert response.status_code == 401
```
AntoLC: This test is misleading, I got the feeling that this feature could not manage anonymous users.

src/backend/core/api/viewsets.py
AntoLC: There is a feature flag variable that we use frontside, should we use it ?

```suggestion
        if serializer.validated_data.get("websocket") || setting.COLLABORATION_WS_NOT_CONNECTED_READY_ONLY === false:
```

src/frontend/servers/y-provider/__tests__/getDocumentConnectionInfoHandler.test.ts
```diff
@@ -0,0 +1,174 @@
+import request from 'supertest';
+import { v4 as uuid } from 'uuid';
+
+const port = 5555;
```
AntoLC: I can see the test fails, better to use another port, this one is used here:
https://github.com/suitenumerique/docs/blob/2e1b1121337af2ac305f04e6d0b2fc0f933956d9/src/frontend/servers/y-provider/__tests__/collaborationResetConnections.test.ts#L3

I guess the port of the previous test does not have the time to close.

AntoLC (approved):
I tried the feature frontside to test if can-edit toggle correctly, seems to works well!

src/backend/impress/settings.py
```diff
@@ -649,6 +651,12 @@ class Base(Configuration):
         environ_prefix=None,
     )
 
+    NO_WEBSOCKET_CACHE_TIMEOUT = values.Value(
```
AntoLC, line 664: Should be added here `docs/env.md`

src/backend/core/services/collaboration_services.py
lunika: Done, I choose to return `count` and `exists`

src/backend/core/api/viewsets.py
lunika: Not sure to understand what you mean. I raise nothing here.

src/backend/core/middleware.py
lunika: It is tested with the integration test (update and can_edit endpoint)

src/backend/core/api/viewsets.py
lunika: We should remove it with this PR probably ?

src/backend/impress/settings.py
lunika: Yes I will add it

src/backend/core/api/viewsets.py
AntoLC: Yes why not, the blocking will be less impacting, we maybe not need a feature flag anymore.

src/backend/core/tests/documents/test_api_documents_can_edit.py
```diff
@@ -11,16 +11,38 @@
...
+    )
+    ws_resp = responses.get(endpoint_url, json={"count": 0, "exists": False})
+
     response = client.get(f"/api/v1.0/documents/{document.id!s}/can-edit/")
-    assert response.status_code == 401
+
+    if role == "reader":
+        assert response.status_code == 401
+    else:
+        assert response.status_code == 200
```
AntoLC, line 36: We should update the docstring as well.

AntoLC (approved):

virgile-dev:
:partying_face:

PR #1152, opened 2025-07-10
🐛(back) manage can-edit endpoint without created room in the ws

lunika:
## Purpose

In a scenario where the first user is editing a docs without websocket and nobody has reached the websocket server first, the y-provider service will return a 404 and we don't handle this case in the can-edit endpoint leading to a server error.

## Proposal

- [x] 🐛(back) manage can-edit endpoint without created room in the ws


AntoLC (approved):
Should we still manage the error here as well ?
https://github.com/suitenumerique/docs/blob/2e6cb109ef7cc29716181846aae7576478f33458/src/backend/core/api/viewsets.py#L523-L526

lunika:
> Should we still manage the error here as well ?
> 
> https://github.com/suitenumerique/docs/blob/2e6cb109ef7cc29716181846aae7576478f33458/src/backend/core/api/viewsets.py#L523-L526

I think we should. We want to know if we have an error 500 for example.

PR #892, opened 2025-04-18
✨ Add a custom callout block to the editor

ZouicheOmar:
## Purpose
This PR addresses issue #305. Add a new custom block to the editor, the callout block.

## Proposal
- Introduced a custom emoji picker component, using the `emoji-mart` library, which is already used by blocknote. That's to have custom triggers and use of the component.
- Introduced a callout component in the editor that uses the custom emoji picker.
- Updated BlockNoteEditor to include this new block.
- Updated BlockNoteSuggestionMenu and  BlockNoteToolbar to include easy insertion of the callout block.
- Adapted ModalExport to include both PDF and Docx exports for the document containing the new block. 
- Updated translations to include the name and the sentences displayed in the slash menus for the callout.

## Demo
[callout.webm](https://github.com/user-attachments/assets/b92f4294-f17c-43ab-b002-506aeea025cc)
 
## PDF and Docx Exports
callout_pdf_docx

Would love any feedback if this needs improvements, corrections or any changes.

PR #1161, opened 2025-07-10
(frontend) fix callout block navigation issues & styles

ZouicheOmar:
## Purpose
This PR addresses issue #892.
Using the callout block can provoque cursor issues when using arrow for text navigation.
The cursor gets unexpected positions. This happens when the callout block is the first of 
the document and the users tries to go up, and on all directions when the cursor is 
between consecutive callout blocks.
This PR also updates minor styles for the callout block.

## Proposal
- [x] 💄(frontend) update callout styles
- [x]  🐛(frontend) fix callout block arrow navigation

### DIFF
--- a/src/frontend/apps/impress/src/features/docs/doc-editor/components/BlockNoteEditor.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-editor/components/BlockNoteEditor.tsx
@@ -18,7 +18,12 @@ import { Box, TextErrors } from '@/components';
 import { Doc, useIsCollaborativeEditable } from '@/docs/doc-management';
 import { useAuth } from '@/features/auth';
 
-import { useHeadings, useUploadFile, useUploadStatus } from '../hook/';
+import {
+  useCalloutBlock,
+  useHeadings,
+  useUploadFile,
+  useUploadStatus,
+} from '../hook/';
 import useSaveDoc from '../hook/useSaveDoc';
 import { useEditorStore } from '../stores';
 import { cssEditor } from '../styles';
@@ -131,6 +136,7 @@ export const BlockNoteEditor = ({ doc, provider }: BlockNoteEditorProps) => {
 
   useHeadings(editor);
   useUploadStatus(editor);
+  useCalloutBlock(editor);
 
   useEffect(() => {
     setEditor(editor);
--- a/src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-blocks/CalloutBlock.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-blocks/CalloutBlock.tsx
@@ -70,8 +70,10 @@ export const CalloutBlock = createReactBlockSpec(
               }
             `}
             $align="center"
+            $justify="center"
             $height="28px"
             $width="28px"
+            $minWidth="28px"
             $radius="4px"
           >
             {block.props.emoji}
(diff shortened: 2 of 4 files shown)


virgile-dev:
Hey ! Thanks for the follow up @ZouicheOmar 
When you're ready please ask @AntoLC for a review :)

ZouicheOmar:
Also added tests.
@AntoLC would appreciate your review on this

src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-blocks/CalloutBlock.tsx
```diff
@@ -70,8 +70,10 @@ export const CalloutBlock = createReactBlockSpec(
               }
             `}
             $align="center"
+            $justify="center"
             $height="28px"
             $width="28px"
+            $minWidth="28px"
```
AntoLC: If `$width` is set `$minWidth` is not necessary.
```suggestion
```

src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-blocks/CalloutBlock.tsx
```diff
@@ -85,7 +87,16 @@ export const CalloutBlock = createReactBlockSpec(
...
+            as="p"
+            className="inline-content"
+            ref={contentRef}
+            $css={css`
+              & > div {
+                padding-top: 2px;
+              }
+            `}
+          />
         </Box>
```
AntoLC: Ah yes well seen, I think `$justify="center"` will be more appropriate.
I can see as well in the dom that it is a "p" tag that is wrapping a "div" tag, I don't think it is correct, better to let it as a `div` so (default tag).
 

```suggestion
          <Box
            className="inline-content"
            ref={contentRef}
            $justify="center"
          />
        </Box>
```

src/frontend/apps/impress/src/features/docs/doc-editor/components/BlockNoteEditor.tsx
```diff
@@ -125,6 +125,7 @@ export const BlockNoteEditor = ({ doc, provider }: BlockNoteEditorProps) => {
       },
       uploadFile,
       schema: blockNoteSchema,
+      disableExtensions: ['gapCursor'],
```
AntoLC, line 128: I don't see the bug actually, could you provide a image or a small video of it ?

AntoLC (commented):
Thank you for the contribution. 🎉 

Some returns.

src/frontend/apps/impress/src/features/docs/doc-editor/components/BlockNoteEditor.tsx
ZouicheOmar: ### Demo
https://github.com/user-attachments/assets/3e677453-5d3c-47ed-a6d0-cd3dde2ebe65

Here's what it looks like with the extension enabled. When the cursor goes above or between callouts, a [GapCursor](https://prosemirror.net/docs/ref/#gapcursor) instance is added. 
The extension comes with tiptap, yet it's not used by any of the Docs blocks.

### On the blocknotejs docs
https://github.com/user-attachments/assets/ff2cfbce-c1ed-43ea-ae19-6e7aebb11b3f

AntoLC: Ok, I am scared of the side effects that we do not foresee.
I can see in the doc that it is useful for table as well.
https://prosemirror.net/docs/ref/#gapcursor

@YousefED Do you have a opinion about it ? 
Do you think we could remove this extension without too much problems ?

AntoLC:
Let's close it, I am not confident to deactivate a extension that can be potentially usefull for other blocks.
Thank you for the contribution !

virgile-dev:
@ZouicheOmar sorry ! I should have check with Antho before asking you this.
Let me know if there something else you want to do on Docs and will have call with Antho so that we scope it properly.

PR #1172, opened 2025-07-13
Allow masking a document

sampaccoud:
## Purpose

Once users have visited a document to which they have access, they can't remove it from their list view anymore.

Several users reported that this is annoying because a document that gets a lot of updates, keeps popping up at the top of their list view. During our latest inter ministerial steering committee, I promised to fix it fast.

They want to be able to mask the document in a click. 

## Proposal

- [x] Add a field "is_masked" on the LinkTrace model
- [x] Add an API action on documents detail endpoint through which logged-in users can mask/unmask a document for themselves
- [x] Add a filter allowing to see only masked/unmasked documents (by default, masked documents are hidden from list views)
- [x] Update failing tests and add tests for the new "mask" API action 

Once this is merged, we will be able to add a "masked documents" section in the left side bar where the masked documents can still be found. Progressive disclosure design commands that the "masked documents" section is only shown if there is at least one document in it. @rl-83 I've seen your mockups once but could not find them anymore in Docs' Figma. Can you add them here for reference?


src/backend/core/migrations/0024_add_is_masked_field_to_link_trace.py
```diff
@@ -0,0 +1,36 @@
...
+                    ("nl-nl", "Nederlands"),
+                    ("es-es", "Español"),
+                ],
+                default=None,
+                help_text="The language in which the user wants to see the interface.",
+                max_length=10,
+                null=True,
+                verbose_name="language",
+            ),
+        ),
```
AntoLC, line 35: It keeps going this "Alter"; no idea where it could come from ?

src/backend/core/tests/documents/test_api_documents_favorite_list.py
```diff
@@ -41,8 +41,8 @@ def test_api_document_favorite_list_authenticated_with_favorite():
     client = APIClient()
     client.force_login(user)
 
-    # User don't have access to this document, let say it had access and this access has been
-    # removed. It should not be in the favorite list anymore.
+    # User don't have access to this document (e.g the user had access and this
+    # access was removed. It should not be in the favorite list anymore.
```
AntoLC: It is lacking a closing parenthesis somewhere.

src/backend/core/api/viewsets.py
```diff
@@ -1118,6 +1117,45 @@ def favorite(self, request, *args, **kwargs):
...
+                {"detail": "Document was masked"},
+                status=drf.status.HTTP_201_CREATED,
+            )
+
+        # Handle DELETE method to unmask document
+        try:
+            link_trace = models.LinkTrace.objects.get(document=document, user=user)
+        except models.LinkTrace.DoesNotExist:
+            return drf.response.Response(
+                {"detail": "Document was already not masked"},
```
AntoLC, line 1148: It is not already "not masked" I think.

```suggestion
                {"detail": "Document does not exist"},
```

src/backend/core/api/viewsets.py
```diff
@@ -1118,6 +1117,45 @@ def favorite(self, request, *args, **kwargs):
             status=drf.status.HTTP_200_OK,
         )
 
+    @drf.decorators.action(detail=True, methods=["post", "delete"], url_path="mask")
```
AntoLC, line 1117: It should not be a `patch` instead of a `post` ? From a front perspective, you can only mask something that already exist I suppose.


```suggestion
    @drf.decorators.action(detail=True, methods=["patch", "delete"], url_path="mask")
```

src/backend/core/api/viewsets.py
```diff
@@ -1118,6 +1117,45 @@ def favorite(self, request, *args, **kwargs):
...
+                return drf.response.Response(
+                    {"detail": "Document was already masked"},
+                    status=drf.status.HTTP_200_OK,
+                )
+            link_trace.is_masked = True
+            link_trace.save()
+            return drf.response.Response(
+                {"detail": "Document was masked"},
+                status=drf.status.HTTP_201_CREATED,
+            )
```
AntoLC: Same as before, I am not sure that it should manage the link_trace creation, if you arrive here, the link_trace should be existing.

src/backend/core/migrations/0024_add_is_masked_field_to_link_trace.py
rvveber: Iirc it was [introduced in this PR](https://github.com/suitenumerique/docs/pull/742#issuecomment-2772398602)
@AntoLC

AntoLC: Ah yes! I didn't see your previous comment. Seems good to add it so.

src/backend/core/api/viewsets.py
```diff
@@ -1118,6 +1117,45 @@ def favorite(self, request, *args, **kwargs):
...
+        except models.LinkTrace.DoesNotExist:
+            return drf.response.Response(
+                {"detail": "Document was already not masked"},
+                status=drf.status.HTTP_200_OK,
+            )
+
+        link_trace.is_masked = False
+        link_trace.save()
+        return drf.response.Response(
+            {"detail": "Document was unmasked"},
```
```suggestion
```

A `204` response has no content.

src/backend/core/api/viewsets.py
```diff
@@ -1118,6 +1117,45 @@ def favorite(self, request, *args, **kwargs):
...
+            link_trace, created = models.LinkTrace.objects.get_or_create(
+                document=document, user=user, defaults={"is_masked": True}
+            )
+            if not created and link_trace.is_masked:
+                return drf.response.Response(
+                    {"detail": "Document was already masked"},
+                    status=drf.status.HTTP_200_OK,
+                )
+            link_trace.is_masked = True
+            link_trace.save()
```
```suggestion
            link_trace.save(update_fields=["is_masked"])
```

src/backend/core/api/viewsets.py
```diff
@@ -1118,6 +1117,45 @@ def favorite(self, request, *args, **kwargs):
...
+        try:
+            link_trace = models.LinkTrace.objects.get(document=document, user=user)
+        except models.LinkTrace.DoesNotExist:
+            return drf.response.Response(
+                {"detail": "Document was already not masked"},
+                status=drf.status.HTTP_200_OK,
+            )
+
+        link_trace.is_masked = False
+        link_trace.save()
```
```suggestion
        link_trace.save(update_fields=["is_masked"])
```

src/backend/core/api/viewsets.py
sampaccoud: actually I think it should indeed be a POST because it is not on the detail endpoint but on the "mask" endpoint of the document. We are creating/deleting masks. As you said, we know the document already exists since it is a detail endpoint.

src/backend/core/api/viewsets.py
sampaccoud: learnt something :pray:

src/backend/core/api/viewsets.py
sampaccoud: ✅ When to use 400 Bad Request
Return 400 if:

The client is expected to have a LinkTrace, and not having one means the request doesn’t make sense.

For example, if LinkTrace is required as a precondition and its absence means the client is misusing the API.

src/backend/core/api/viewsets.py
sampaccoud: ditto. I'm now returning a 400.

AntoLC (approved):

PR #1184, opened 2025-07-16
♻️(frontend) search on all docs if no children

AntoLC:
## Purpose

When searching for documents, if no children are found, the search will now include all documents instead of just those with children.

## Proposal

- [x] ♻️(frontend) search on all docs if no children

### DIFF
--- a/src/frontend/apps/impress/src/features/left-panel/components/LeftPanelHeader.tsx
+++ b/src/frontend/apps/impress/src/features/left-panel/components/LeftPanelHeader.tsx
@@ -3,7 +3,8 @@ import { useRouter } from 'next/router';
 import { PropsWithChildren, useCallback, useState } from 'react';
 
 import { Box, Icon, SeparatedSection } from '@/components';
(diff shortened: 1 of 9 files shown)


src/frontend/apps/impress/src/features/left-panel/components/LeftPanelHeader.tsx
```diff
@@ -12,10 +13,10 @@ import { useLeftPanelStore } from '../stores';
 import { LeftPanelHeaderButton } from './LeftPanelHeaderButton';
 
 export const LeftPanelHeader = ({ children }: PropsWithChildren) => {
+  const { currentDoc } = useDocStore();
+  const isDoc = !!currentDoc;
```
lunika: `isDoc` is probably not describing enough what is the purpose of this variable.

AntoLC: I refactorized this part in the fixup, moving logic in `DocSearchModal` it would be cleaner I think.

lunika (approved):

PR #1190, opened 2025-07-16
✨(frontend) subdocs can manage link reach

AntoLC:
## Purpose

The subdocs can now have their own link reach properties, dissociated from the parent document.

## Proposal

- [x] ✨(frontend) subdocs can manage link reach

## Demo 

[Test 8989 - Docs.webm](https://github.com/user-attachments/assets/94c6b3dc-2dca-4318-bcbb-8b52b91a7684)

### DIFF
--- a/src/frontend/apps/impress/src/features/docs/doc-tree/hooks/useTreeUtils.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-tree/hooks/useTreeUtils.tsx
@@ -9,9 +9,8 @@ export const useTreeUtils = (doc: Doc) => {
     isParent: doc.nb_accesses_ancestors <= 1, // it is a parent
     isChild: doc.nb_accesses_ancestors > 1, // it is a child
     isCurrentParent: treeContext?.root?.id === doc.id || doc.depth === 1, // it can be a child but not for the current user
-    isDesyncronized: !!(
+    isDesynchronized: !!(
       doc.ancestors_link_reach &&
-      doc.ancestors_link_role &&
       (doc.computed_link_reach !== doc.ancestors_link_reach ||
         doc.computed_link_role !== doc.ancestors_link_role)
     ),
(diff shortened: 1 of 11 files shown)


virgile-dev (approved):

src/frontend/apps/impress/src/features/docs/doc-tree/hooks/useTreeUtils.tsx
```diff
@@ -1,14 +1,10 @@
...
 export const useTreeUtils = (doc: Doc) => {
-  const treeContext = useTreeContext<Doc>();
-
   return {
-    isParent: doc.nb_accesses_ancestors <= 1, // it is a parent
-    isChild: doc.nb_accesses_ancestors > 1, // it is a child
-    isCurrentParent: treeContext?.root?.id === doc.id || doc.depth === 1, // it can be a child but not for the current user
+    isTopParent: doc.depth === 1,
+    isChild: doc.depth > 1,
+    isCurrentTopParent: typeof doc.children === 'undefined', // it can be a child but not for the current user
```
lunika: if a doc has no children, it can't be a parent. Not sure to understand what you are looking for.

AntoLC: I need to know if a doc is the current root, the problem is, a child can be a root for another user, so `doc.depth === 1` cannot work in some cases:
```
Parent -> restricted
├─Child -> Public
```
If I share the `Child` link to a anonymous user, `Child` will be a root for this anonymous user, with `doc.depth === 2`.
The only variable available to know if a doc is the current tree root, is with the property `children` apparently.

If you can make me a clean backend property it will be nice though (ex: `doc.is_root`).

AntoLC: The doc response came from the endpoint `documents/[DOC_ID]/`, it is why it was without `child` array.

There is not a reliable way to know if a doc is the top root without context, I use another way to get the information that I need so:
https://github.com/suitenumerique/docs/blob/e8147d02c1bb9eeea86882a6425ab39be7c0929e/src/frontend/apps/impress/src/features/docs/doc-tree/components/DocTree.tsx#L210

lunika (approved):

PR #1197, opened 2025-07-17
🐛(frontend) fix empty left panel after deleting doc

AntoLC:
## Purpose

When we were deleting a parent document, the left panel was getting empty. It was because the panel thought that it was a child document and was trying clear dynamically the panel.

## Proposal

Now, we are checking if the document is a parent or not, if it is a parent we just redirect to the homepage.


CHANGELOG.md
```diff
@@ -8,8 +8,9 @@ and this project adheres to
```
lunika: In the commit body, mentionning `root document` instead of `parent document` is more accurate IMO. Every doc can be a parent but the only one concerned here is the root document.

CHANGELOG.md
```diff
@@ -8,8 +8,9 @@ and this project adheres to
 
 ## [Unreleased]
 
-## Fixed
+### Fixed
 
+- 🐛(frontend) fix empty left panel after deleting doc #1197
```
lunika:
```suggestion
- 🐛(frontend) fix empty left panel after deleting root doc #1197
```

lunika (approved):

PR #1200, opened 2025-07-18
🔧(project) change env.d system by using local files

AntoLC:
## Purpose

We had lot of problems with the previous `env.d` system. Users were often confused by the need to change the `env.d` files manually, leading to issues when using the project locally.

## Proposal

This PR introduces a new system that uses `.env.local` files, which are automatically created and can be modified by users without affecting the original `env.d` files. This should simplify the development process and reduce confusion by removing the need to manually edit `env.d` files.


sylvinus:
LGTM! We did the same on messages

AntoLC:
> LGTM! We did the same on messages

Ah yes, quite nice this part ^^

https://github.com/suitenumerique/messages/blob/9bef38c53a30e353c238abc87bf39ff5d26a05cc/Makefile#L79-L110

PR #1213, opened 2025-07-24
✨(api) add API route to fetch document content

### LINKED ISSUE #1206: Html and markdown endpoint to retrieve content of public docs
## Feature Request

Some third party apps and users want to reuse documents to be displayed somewhere else than Docs

Expose the content of public documents through an API endpoint. 
For simplicity let's start only with public docs so we don't have to deal with authentication
We want to expose the markdown and html formats of a document to third parties

Currently users can copy content as Markdown and HTML.

Blocknote has an API for this maybe we could expose it.
This is a request from @sylvinus

sylvinus:
This allows API users to fetch document content, enabling the use of Docs as a headless CMS for instance, or any kind of document processing, with `/api/v1.0/documents/xyz/content/?content_format=html` URLs.

Fixes #1206.


StephanMeijer:
## Overhead

Instead of using the `"content"` and `"format"` fields, you can use the `Content-Type`-header and the body of the request. This will have as effect:

- No overhead in JSON-encoding
- Aligning with the HTTP spec more
- No overhead in base64: CPU/memory but also network traffic (base64 also increases traffic with ~33% and JSON also increases payload size with a tiny bit)

This is in line with the existing `convert`-endpoint.

## Utilizing the existing `convert`-endpoint

Instead of writing a new endpoint, you can rewrite the `convert`-endpoint to accept a `Accept` and `Content-Type` header.

e.g.

```
POST /convert

Content-Type: text/markdown
Accept: application/vnd.yjs.doc

# Heading

Some text
```

Would transform the markdown document into a YJs document (raw non-base64 encoded content).

e.g.

```
POST /convert

Content-Type: application/vnd.yjs.doc
Accept: text/html

<binary contents>
```

Would transform the YJs document (raw non-base64 encoded content) to HTML.

It'd then respond by:

```
Content-Type: text/html

<h1>Heading</h1><p>Some text</p>
```

## Last but not least.

If you have any questions about this approach, feel free to schedule a call with me and/or ask @lunika to be in our secret Matrix room.

I also left some code style comments.

src/frontend/servers/y-provider/src/handlers/contentHandler.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+interface ContentRequest {
+  content: string;
+  format: string;
+}
+
+const editor = ServerBlockNoteEditor.create();
+
+export const contentHandler = async (
```
StephanMeijer: See earlier comment about utilizing existing endpoint.

src/frontend/servers/y-provider/src/handlers/contentHandler.ts
```diff
@@ -0,0 +1,71 @@
...
+      result = await editor.blocksToHTMLLossy(blocks as PartialBlock[]);
+    } else {
+      res.status(400).json({ error: 'Unsupported format' });
+      return;
+    }
+
+    res.status(200).json({
+      content: result,
+      format: format,
+    });
```
StephanMeijer: `result` is effectively a useless assignment.

```suggestion
    if (!blocks || blocks.length === 0) {
      res.status(200).json({ content: null, format });
    } else if (format === 'json') {
      res.status(200).json({ content: blocks, format });
    } else if (format === 'markdown') {
      res.status(200).json({ content: await editor.blocksToMarkdownLossy(blocks as PartialBlock[]), format });
    } else if (format === 'html') {
      res.status(200).json({ content: await editor.blocksToHTMLLossy(blocks as PartialBlock[]), format });
    } else {
      res.status(400).json({ error: 'Unsupported format' });
    }
```

StephanMeijer: Also: i'd extract this to a seperate piece of logic instead of having a huge if-else if-else-construction.

src/frontend/servers/y-provider/__tests__/content.test.ts
```diff
@@ -0,0 +1,207 @@
...
+        expect(parsedContent[1].content[0].text).toBe('This is test content.');
+      } else if (format === 'markdown') {
+        expect(typeof response.body.content).toBe('string');
+        expect(response.body.content.trim()).toBe(markdownContent);
+      } else if (format === 'html') {
+        expect(typeof response.body.content).toBe('string');
+        expect(response.body.content).toBe(
+          '<h1>Test Document</h1><p>This is test content.</p>',
+        );
+      }
```
StephanMeijer: Having if-else if's based on test parameters is generally considered bad practice. Better to split into separate tests.

src/frontend/servers/y-provider/__tests__/content.test.ts
```diff
@@ -0,0 +1,207 @@
...
+      .post('/api/content')
+      .set('origin', origin)
+      .set('authorization', apiKey)
+      .set('content-type', 'application/json')
+      .send({
+        content: 'invalid-base64-content!@#',
+        format: 'json',
+      });
+
+    expect(response.status).toBe(500);
```
StephanMeijer: Instead of `500 Internal Server Error` it'd be neat to do a `400 Bad Request`.

src/frontend/servers/y-provider/__tests__/content.test.ts
```diff
@@ -0,0 +1,207 @@
...
+      if (format === 'json') {
+        const parsedContent = response.body.content;
+        expect(Array.isArray(parsedContent)).toBe(true);
+        expect(parsedContent.length).toBe(2);
+        expect(parsedContent[0].type).toBe('heading');
+        expect(parsedContent[1].type).toBe('paragraph');
+        expect(parsedContent[0].content[0].type).toBe('text');
+        expect(parsedContent[0].content[0].text).toBe('Test Document');
+        expect(parsedContent[1].content[0].type).toBe('text');
+        expect(parsedContent[1].content[0].text).toBe('This is test content.');
```
StephanMeijer: I'd do a `expect(...).toStrictEqual()`. You are not testing the output fully, constraining what comes out of it beyond the specific fields you are testing, etc.

src/frontend/servers/y-provider/__tests__/content.test.ts
StephanMeijer: Sources:

- https://testsmells.org/pages/testsmells.html#ConditionalTestLogic
- https://dev.to/ethancarlsson/dont-write-conditional-logic-in-tests-5aa
- https://github.com/vitest-dev/eslint-plugin-vitest/blob/main/docs/rules/no-conditional-expect.md

src/frontend/servers/y-provider/__tests__/content.test.ts
StephanMeijer: See also:

- https://github.com/suitenumerique/docs/blob/main/src/frontend/servers/y-provider/__tests__/convert.test.ts#L21
- https://github.com/suitenumerique/docs/blob/main/src/frontend/servers/y-provider/__tests__/convert.test.ts#L150

docker/auth/realm.json
```diff
@@ -26,7 +26,7 @@
   "oauth2DeviceCodeLifespan": 600,
   "oauth2DevicePollingInterval": 5,
   "enabled": true,
-  "sslRequired": "external",
+  "sslRequired": "none",
```
StephanMeijer: Is this committed accidentally or on purpose?

sylvinus: This fixes Keycloak requiring HTTPS on macOS because it doesn't recognize the host IP as local. We had to to the same for Messages: https://github.com/suitenumerique/messages/commit/7d8daaeaec97ccc3835f03feea501d1b11b38c90

sylvinus: I may do a separate PR if you prefer

sylvinus:
@StephanMeijer Thanks for the review! Your approach is more concise indeed and I implemented all your feedbacks I think.

StephanMeijer:
Will give it a new review today or tomorrow!

docker/auth/realm.json
StephanMeijer: > I may do a separate PR if you prefer

I have no preference, just seemed like something I would accidentally commit.

StephanMeijer:
@sylvinus Can I propose a different style of writing this tomorrow? Need some time to get behind my editor,.

sylvinus:
@StephanMeijer feel free to push on the branch, frankly I don't care that much about the internals, I just badly need the API as a user ;)

sylvinus:
@StephanMeijer hi! Any news on this? Thanks !!

sylvinus:
I've deployed this to our instance and begun testing. One thing I've just realized is that we need to register custom blocks to the server-side editor so they are rendered properly as HTML or Markdown.

That seems to introduce a dependency from impress to the y-provider, that didn't exist yet. Or should we put custom blocks in a new, third package?

sylvinus:
FYI, prior art with a route that generated markdown by parsing the yjs blob : https://github.com/suitenumerique/docs/commit/06c26a7179517bb80060306e02a86c5ea367e1a2

Curious to discuss the benefits of both approaches!

francois-ferrandis:
Hello,

This feature could be very useful to the RDV Service Public[^1] team, as we are currently looking for a replacement to https://headwayapp.co/ as an product announcement system and came across this guide[^2].

This leads me to the question: is this API stable? Is it safe to consider it as backwards compatible? Do you plan on introducing versioned APIs?

Thank you so much for all your work, we love Docs and can't way to see what comes next! :bow: :pray: :star_struck: 

[^1]: https://beta.gouv.fr/startups/rdv-service-public.html
[^2]: https://docs.suite.anct.gouv.fr/docs/0b65cc5b-2d72-408a-b5c1-0ff8d2a7a479/

PR #1219, opened 2025-07-24
✨(frontend) add multi columns support for editor

AntoLC:
## Purpose

We add multi columns support for editor, now you can add columns to your document.
Works with export.

## Careful ⚠️ 

📄AGPL feature.

## Demo

https://github.com/user-attachments/assets/9eebd8b8-629c-4f0a-a8f8-54e08a61a977

### DIFF
--- a/src/frontend/apps/impress/src/features/docs/doc-editor/components/BlockNoteEditor.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-editor/components/BlockNoteEditor.tsx
@@ -28,8 +28,13 @@ import { randomColor } from '../utils';
 import { BlockNoteSuggestionMenu } from './BlockNoteSuggestionMenu';
 import { BlockNoteToolbar } from './BlockNoteToolBar/BlockNoteToolbar';
 import { CalloutBlock, DividerBlock } from './custom-blocks';
+import XLMultiColumn from './xl-multi-column';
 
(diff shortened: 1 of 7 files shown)


src/frontend/apps/impress/src/features/docs/doc-editor/components/BlockNoteEditor.tsx
```diff
@@ -116,7 +126,11 @@ export const BlockNoteEditor = ({ doc, provider }: BlockNoteEditorProps) => {
         },
         showCursorLabels: showCursorLabels as 'always' | 'activity',
       },
-      dictionary: locales[lang as keyof typeof locales],
+      dictionary: {
+        ...locales[lang as keyof typeof locales],
+        multi_column:
+          multiColumnLocales[lang as keyof typeof multiColumnLocales],
```
lunika: `multiColumnLocales` can be undefined so this code will crash the application

AntoLC: Yes you'r right.

AntoLC: If we switch to Vitest, I think we should be able to test correctly this part, for the moment it is complicated.

Issue #1143, opened 2025-07-09
Move email sending in a celery task


lunika:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**

Emails are sent during a request, sometimes the connection to the server email can take time or fail.

**Describe the solution you'd like**

Move the email sending in a celery task to avoid the side effect listed before.

HaroWana:
Bonjour,
J'aimerais m'occuper de cette issue si c'est ok pour vous. :)

lunika:
@HaroWana You are assigned on this issue.
Thanks

PR #1236, opened 2025-07-29
⚡(backend) move email sending to celery

HaroWana:
Makes email sending asynchronous by moving email operations to Celery tasks, preventing API endpoints from blocking.

This pull request addresses issue #1143


src/backend/core/tasks/mail.py
```diff
@@ -22,3 +22,24 @@ def send_ask_for_access_mail(ask_for_access_id):
...
             )
+
+
+@app.task
+def send_mail(document_id, subject, emails, context=None, language=None):
+    """Send email from a document template."""
+    try:
+        document = models.Document.objects.get(id=document_id)
+        document.send_email(subject, emails, context, language)
+    except models.Document.DoesNotExist:
```
lunika: I would suggest to not fail silently. Let the exception bubbles, your monitoring system will warn you about this error. Because it is not normal to not find the document here.

src/backend/core/tasks/mail.py
```diff
@@ -22,3 +22,24 @@ def send_ask_for_access_mail(ask_for_access_id):
...
+
+
+@app.task
+def send_invitation_mail(document_id, email, role, sender_id, language=None):
+    """Send invitation email for a document."""
+    try:
+        document = models.Document.objects.get(id=document_id)
+        sender = models.User.objects.get(id=sender_id)
+        document.send_invitation_email(email, role, sender, language)
+    except (models.Document.DoesNotExist, models.User.DoesNotExist):
```
lunika: Same here

src/backend/core/tasks/mail.py
HaroWana: That's a good point. Removing the try/except clauses should allow the exception to bubble up then, right?
Thank you !

lunika (approved):
Tell me if you have trouble

virgile-dev:
I see that this PR has staled should we close it and assign it to someone else?

PR #1243, opened 2025-07-30
🐛(frontend) include root parent in search

AntoLC:
## Purpose

The root parent was not included in the search.

## Proposal

When searching for documents, the root parent document is now included in the search results if it matches the search query.

### DIFF
--- a/src/frontend/apps/impress/src/features/docs/doc-search/components/DocSearchSubPageContent.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-search/components/DocSearchSubPageContent.tsx
@@ -45,6 +45,16 @@ export const DocSearchSubPageContent = ({
   const docsData: QuickSearchData<Doc> = useMemo(() => {
     const subDocs = subDocsData?.pages.flatMap((page) => page.results) || [];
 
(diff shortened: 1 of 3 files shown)


src/frontend/apps/impress/src/features/docs/doc-search/components/DocSearchSubPageContent.tsx
```diff
@@ -45,6 +45,13 @@ export const DocSearchSubPageContent = ({
   const docsData: QuickSearchData<Doc> = useMemo(() => {
     const subDocs = subDocsData?.pages.flatMap((page) => page.results) || [];
 
+    if (
+      treeContext?.root &&
```
Ovgodd: If I may suggest, wrapping this condition in a variable could improve readability, little exemple : 

const rootDoc = treeContext?.root;
const shouldIncludeRootInResults =
  !!rootDoc && rootDoc.title?.toLowerCase().includes(search.toLowerCase());


```suggestion
     if (shouldIncludeRootInResults) {
  subDocs.unshift(rootDoc);
}

```

lunika (approved):

Issue #1136, opened 2025-07-07
header choice should be constrained


Erwan-le-Gall:
label: accessibility

When I insert a new header

Then I should be constrained to:

- immediately preceding higher level
- same level
- immediately following loxer level

because headers hierarchy should not have "holes" in it (it must be continuous).

This is a requirement for accessibility compliance.

Erwan-le-Gall:
ping @cyberbaloo

AntoLC:
I find that strange, what if you just want to have a big text somewhere, it would not be possible anymore ?

WDYT @rl-83 ?

https://github.com/user-attachments/assets/d49a8d5d-53dd-4c09-ad07-ce1cf58fa1e5

Erwan-le-Gall:
You're not supposed to use markup based solely on its apparence. You **must** use it because of its **semantic**.
Blind people don't give a sh* about how bold and big the text is. But they rely on the semantic to dynamically extract a ToC (table of content), which they use to navigate around in the document. Especially a long document.
A text mut be marked as a title because it is really a title.
Same for SEO, by the way…

Erwan-le-Gall:
So the hierarchy of titles must be logical : h1 followed by h1 or h2, h2 followed by h1 or h2 or h3 and so on. No "holes".

AntoLC:
Still, it does not solve on how do we get a bigger text after having set a h1 size and h2 size.
If we do that, we should first introduce a **size feature** I think, to be able to size texts other than titles.

Blocking user interactions is a shame I find, it will be complicated with the AI feature as well.

You can as well "warn" users instead of blocking them.

robin-lecomte:
@Erwan-le-Gall, do you have any examples of software that uses this approach? It could help us better understand the behavior you're expecting.

Erwan-le-Gall:
Nope.
Example:
`<h1>First` level title</h1>
<p>blah</p>
<h…
  [1]
  [2]

  [3], [4],[…] should not be offered.
`

Erwan-le-Gall:
screenshot:
text is:
> Titre
un essai
sous-titre

Titre is H1
<img width="399" height="419" alt="Image" src="https://github.com/user-attachments/assets/c5e888b4-5869-474f-8614-7786de932f94" />

Erwan-le-Gall:
…and sous-titre is H2…
[...]

PR #1251, opened 2025-07-31
Fix/1136 heading order

Ovgodd:
## Purpose

This PR introduces a heading accessibility filter in the slash menu to progressively limit heading levels (H1 → H2 → H3) in the document editor. It also adds an end-to-end test to validate the filtering logic.

that solve issue : [1136](https://github.com/suitenumerique/docs/issues/1136)

## Proposal

- Add useHeadingAccessibilityFilter hook to control heading availability
- Integrate filter in BlockNoteSuggestionMen
- https://github.com/user-attachments/assets/31dd9296-bb49-4a94-bbc6-edf4b5682c42
- Ensure only one H1 is allowed per document for accessibility compliance
- Allow multiple H2 and H3 levels, which is semantically and accessibly correct

 Write E2E test to ensure correct progressive filtering behavior (Playwright)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,9 @@ and this project adheres to
 - ♻️(frontend) redirect to doc after duplicate #1175
 - 🔧(project) change env.d system by using local files #1200
 - ⚡️(frontend) improve tree stability #1207
-- ⚡️(frontend) improve accessibility #1232
+- ⚡️(frontend) improve accessibility
+  - #1232
+  - #1251
 - 🛂(frontend) block drag n drop when not desktop #1239
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


AntoLC (commented):
Let's wait a bit with this PR to be sure about the behavior that we want to implement.
We need to validate altogether with @rl-83 and @virgile-dev.

Ovgodd:
> Let's wait a bit with this PR to be sure about the behavior that we want to implement. We need to validate altogether with @rl-83 and @virgile-dev.

I’ve set this PR to draft in the meantime

AntoLC:
Let's close it, I added the label `keep track` to open it easily if necessary.

PR #1276, opened 2025-08-07
Index to search

sampaccoud:
## Purpose

We want to add fulltext (and semantic in a second phase) search to Docs.

The goal is to enable efficient and scalable search across document content by pushing relevant data to a dedicated search backend, such as OpenSearch. The backend should be pluggable. 

## Proposal

- [x] Add indexing logic in a search indexer that can be declared as a backend
- [x] Implement indexing for the Find backend. See [corresponding PR in Find](https://github.com/suitenumerique/find/pull/16)
- [ ] Implement search views as a proxy

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -980,6 +980,11 @@ class DocumentViewSet(
             {"id": str(duplicated_document.id)}, status=status.HTTP_201_CREATED
         )
 
+    # TODO
+    # @drf.decorators.action(detail=False, methods=["get"])
+    # def search(self, request, *args, **kwargs):
+    #    index.search()
+
     @drf.decorators.action(detail=True, methods=["get"], url_path="versions")
     def versions_list(self, request, *args, **kwargs):
         """
--- a/src/backend/core/management/commands/index.py
+++ b/src/backend/core/management/commands/index.py
@@ -1,0 +1,28 @@
+"""
+Handle search setup that needs to be done at bootstrap time.
+"""
+
+import logging
+import time
+
+from django.core.management.base import BaseCommand
+
+from ...services.search_indexers import FindDocumentIndexer
+
+logger = logging.getLogger("docs.search.bootstrap_search")
+
+
+class Command(BaseCommand):
+    """Index all documents to remote search service"""
+
+    help = __doc__
+
+    def handle(self, *args, **options):
+        """Launch and log search index generation."""
+        logger.info("Starting to regenerate Find index...")
+        start = time.perf_counter()
+
+        FindDocumentIndexer().index()
+
+        duration = time.perf_counter() - start
+        logger.info(f"Search index regenerated in {duration:.2f} seconds.")
--- a/src/backend/core/models.py
+++ b/src/backend/core/models.py
@@ -439,32 +439,35 @@ class Document(MP_Node, BaseModel):
     def save(self, *args, **kwargs):
         """Write content to object storage only if _content has changed."""
         super().save(*args, **kwargs)
-
         if self._content:
-            file_key = self.file_key
-            bytes_content = self._content.encode("utf-8")
+            self.save_content(self._content)
 
-            # Attempt to directly check if the object exists using the storage client.
-            try:
-                response = default_storage.connection.meta.client.head_object(
-                    Bucket=default_storage.bucket_name, Key=file_key
-                )
-            except ClientError as excpt:
-                # If the error is a 404, the object doesn't exist, so we should create it.
-                if excpt.response["Error"]["Code"] == "404":
-                    has_changed = True
-                else:
-                    raise
+    def save_content(self, content):
+        """Save content to object storage."""
+
+        file_key = self.file_key
+        bytes_content = content.encode("utf-8")
+
+        # Attempt to directly check if the object exists using the storage client.
+        try:
+            response = default_storage.connection.meta.client.head_object(
+                Bucket=default_storage.bucket_name, Key=file_key
+            )
+        except ClientError as excpt:
+            # If the error is a 404, the object doesn't exist, so we should create it.
+            if excpt.response["Error"]["Code"] == "404":
+                has_changed = True
             else:
-                # Compare the existing ETag with the MD5 hash of the new content.
-                has_changed = (
-                    response["ETag"].strip('"')
-                    != hashlib.md5(bytes_content).hexdigest()  # noqa: S324
-                )
+                raise
+        else:
+            # Compare the existing ETag with the MD5 hash of the new content.
+            has_changed = (
+                response["ETag"].strip('"') != hashlib.md5(bytes_content).hexdigest()  # noqa: S324
+            )
 
-            if has_changed:
-                content_file = ContentFile(bytes_content)
-                default_storage.save(file_key, content_file)
+        if has_changed:
+            content_file = ContentFile(bytes_content)
+            default_storage.save(file_key, content_file)
 
     def is_leaf(self):
         """
--- a/src/backend/core/services/search_indexers.py
+++ b/src/backend/core/services/search_indexers.py
@@ -1,0 +1,170 @@
+"""Document search index management utilities and indexers"""
+
+import logging
+from abc import ABC, abstractmethod
+from collections import defaultdict
+
+from django.conf import settings
+
+import requests
+
+from core import models, utils
+
+logger = logging.getLogger(__name__)
+
+
+def get_batch_accesses_by_users_and_teams(paths):
+    """
+    Get accesses related to a list of document paths,
+    grouped by users and teams, including all ancestor paths.
+    """
+    print("paths: ", paths)
+    ancestor_map = utils.get_ancestor_to_descendants_map(paths, steplen=models.Document.steplen)
+    ancestor_paths = list(ancestor_map.keys())
+    print("ancestor map: ", ancestor_map)
+    print("ancestor paths: ", ancestor_paths)
+
+    access_qs = models.DocumentAccess.objects.filter(
+        document__path__in=ancestor_paths
+    ).values("document__path", "user__sub", "team")
+
+    access_by_document_path = defaultdict(lambda: {"users": set(), "teams": set()})
+
+    for access in access_qs:
+        ancestor_path = access["document__path"]
+        user_sub = access["user__sub"]
+        team = access["team"]
+
+        for descendant_path in ancestor_map.get(ancestor_path, []):
+            if user_sub:
+                access_by_document_path[descendant_path]["users"].add(str(user_sub))
+            if team:
+                access_by_document_path[descendant_path]["teams"].add(team)
+
+    return dict(access_by_document_path)
+
+
+class BaseDocumentIndexer(ABC):
+    """
+    Base class for document indexers.
+
+    Handles batching and access resolution. Subclasses must implement both
+    `serialize_document()` and `push()` to define backend-specific behavior.
+    """
+
+    def __init__(self, batch_size=None):
+        """
+        Initialize the indexer.
+
+        Args:
+            batch_size (int, optional): Number of documents per batch.
+                Defaults to settings.SEARCH_INDEXER_BATCH_SIZE.
+        """
+        self.batch_size = batch_size or settings.SEARCH_INDEXER_BATCH_SIZE
+
+    def index(self):
+        """
+        Fetch documents in batches, serialize them, and push to the search backend.
+        """
+        last_id = 0
+        while True:
+            documents_batch = list(
+                models.Document.objects.filter(
+                    id__gt=last_id,
+                ).order_by("id")[: self.batch_size]
+            )
+
+            if not documents_batch:
+                break
+
+            doc_paths = [doc.path for doc in documents_batch]
+            last_id = documents_batch[-1].id
+            accesses_by_document_path = get_batch_accesses_by_users_and_teams(doc_paths)
+
+            serialized_batch = [
+                self.serialize_document(document, accesses_by_document_path)
+                for document in documents_batch
+            ]
+            self.push(serialized_batch)
+
+    @abstractmethod
+    def serialize_document(self, document, accesses):
+        """
+        Convert a Document instance to a JSON-serializable format for indexing.
+
+        Must be implemented by subclasses.
+        """
+
+    @abstractmethod
+    def push(self, data):
+        """
+        Push a batch of serialized documents to the backend.
+
+        Must be implemented by subclasses.
+        """
+
+
+class FindDocumentIndexer(BaseDocumentIndexer):
+    """
+    Document indexer that pushes documents to La Suite Find app.
+    """
+
+    def serialize_document(self, document, accesses):
+        """
+        Convert a Document to the JSON format expected by La Suite Find.
+
+        Args:
+            document (Document): The document instance.
+            accesses (dict): Mapping of document ID to user/team access.
+
+        Returns:
+            dict: A JSON-serializable dictionary.
+        """
+        doc_path = document.path
+        text_content = utils.base64_yjs_to_text(document.content)
+        return {
+            "id": str(document.id),
+            "title": document.title,
+            "content": text_content,
+            "depth": document.depth,
+            "path": document.path,
+            "numchild": document.numchild,
+            "created_at": document.created_at.isoformat(),
+            "updated_at": document.updated_at.isoformat(),
+            "users": list(accesses.get(doc_path, {}).get("users", set())),
+            "groups": list(accesses.get(doc_path, {}).get("teams", set())),
+            "reach": document.computed_link_reach,
+            "size": len(text_content.encode("utf-8")),
+            "is_active": not bool(document.ancestors_deleted_at),
+        }
+
+    def push(self, data):
+        """
+        Push a batch of documents to the Find backend.
+
+        Args:
+            data (list): List of document dictionaries.
+        """
+        url = getattr(settings, "SEARCH_INDEXER_URL", None)
+        if not url:
+            raise RuntimeError(
+                "SEARCH_INDEXER_URL must be set in Django settings before indexing."
+            )
+
+        secret = getattr(settings, "SEARCH_INDEXER_SECRET", None)
+        if not secret:
+            raise RuntimeError(
+                "SEARCH_INDEXER_SECRET must be set in Django settings before indexing."
+            )
+        try:
+            response = requests.post(
+                url,
+                json=data,
+                headers={"Authorization": f"Bearer {secret}"},
+                timeout=10,
+            )
+            response.raise_for_status()
+        except requests.exceptions.HTTPError as e:
+            logger.error("HTTPError: %s", e)
+            logger.error("Response content: %s", response.text)
+            raise
--- a/src/backend/core/services/search_indexers.py
+++ b/src/backend/core/services/search_indexers.py
@@ -138,6 +138,13 @@ class FindDocumentIndexer(BaseDocumentIndexer):
             "is_active": not bool(document.ancestors_deleted_at),
         }
 
+    # TODO:
+    # def search(self, ):
+    #     # Include access token (resource server)
+    #     find.post(find_url, ...HTTP_AUTHORIZATION=access_token...)
+
+    # def format_response():
+
     def push(self, data):
         """
         Push a batch of documents to the Find backend.
--- a/src/backend/core/tests/test_services_search_indexers.py
+++ b/src/backend/core/tests/test_services_search_indexers.py
@@ -1,0 +1,243 @@
+"""Tests for Documents search indexers"""
+
+from unittest.mock import patch
+
+import pytest
+
+from core import factories, utils
+from core.services.search_indexers import FindDocumentIndexer
+
+pytestmark = pytest.mark.django_db
+
+
+def test_push_raises_error_if_search_indexer_url_is_none(settings):
+    """
+    Indexer should raise RuntimeError if SEARCH_INDEXER_URL is None or empty.
+    """
+    settings.SEARCH_INDEXER_URL = None
+    indexer = FindDocumentIndexer()
+
+    with pytest.raises(RuntimeError) as exc_info:
+        indexer.push([])
+
+    assert "SEARCH_INDEXER_URL must be set in Django settings before indexing." in str(
+        exc_info.value
+    )
+
+
+def test_push_raises_error_if_search_indexer_url_is_empty(settings):
+    """
+    Indexer should raise RuntimeError if SEARCH_INDEXER_URL is empty string.
+    """
+    settings.SEARCH_INDEXER_URL = ""
+    indexer = FindDocumentIndexer()
+
+    with pytest.raises(RuntimeError) as exc_info:
+        indexer.push([])
+
+    assert "SEARCH_INDEXER_URL must be set in Django settings before indexing." in str(
+        exc_info.value
+    )
+
+
+def test_push_raises_error_if_search_indexer_secret_is_none(settings):
+    """
+    Indexer should raise RuntimeError if SEARCH_INDEXER_SECRET is None or empty.
+    """
+    settings.SEARCH_INDEXER_SECRET = None
+    indexer = FindDocumentIndexer()
+
+    with pytest.raises(RuntimeError) as exc_info:
+        indexer.push([])
+
+    assert (
+        "SEARCH_INDEXER_SECRET must be set in Django settings before indexing."
+        in str(exc_info.value)
+    )
+
+
+def test_push_raises_error_if_search_indexer_secret_is_empty(settings):
+    """
+    Indexer should raise RuntimeError if SEARCH_INDEXER_SECRET is empty string.
+    """
+    settings.SEARCH_INDEXER_SECRET = ""
+    indexer = FindDocumentIndexer()
+
+    with pytest.raises(RuntimeError) as exc_info:
+        indexer.push([])
+
+    assert (
+        "SEARCH_INDEXER_SECRET must be set in Django settings before indexing."
+        in str(exc_info.value)
+    )
+
+
+def test_services_search_indexers_serialize_document_returns_expected_json():
+    """
+    It should serialize documents with correct metadata and access control.
+    """
+    user_a, user_b = factories.UserFactory.create_batch(2)
+    document = factories.DocumentFactory()
+    factories.DocumentFactory(parent=document)
+
+    factories.UserDocumentAccessFactory(document=document, user=user_a)
+    factories.UserDocumentAccessFactory(document=document, user=user_b)
+    factories.TeamDocumentAccessFactory(document=document, team="team1")
+    factories.TeamDocumentAccessFactory(document=document, team="team2")
+
+    accesses = {
+        document.path: {
+            "users": {str(user_a.sub), str(user_b.sub)},
+            "teams": {"team1", "team2"},
+        }
+    }
+
+    indexer = FindDocumentIndexer()
+    result = indexer.serialize_document(document, accesses)
+
+    assert set(result.pop("users")) == {str(user_a.sub), str(user_b.sub)}
+    assert set(result.pop("groups")) == {"team1", "team2"}
+    assert result == {
+        "id": str(document.id),
+        "title": document.title,
+        "depth": 1,
+        "path": document.path,
+        "numchild": 1,
+        "content": utils.base64_yjs_to_text(document.content),
+        "created_at": document.created_at.isoformat(),
+        "updated_at": document.updated_at.isoformat(),
+        "reach": document.link_reach,
+        'size': 13,
+        "is_active": True,
+    }
+
+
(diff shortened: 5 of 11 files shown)


src/backend/core/api/serializers.py
```diff
@@ -821,3 +821,17 @@ class MoveDocumentSerializer(serializers.Serializer):
...
+
+    q = serializers.CharField(required=True)
+
+    def validate_q(self, value):
+        """Ensure the text field is not empty."""
+
+        if len(value.strip()) == 0:
+            raise serializers.ValidationError("Text field cannot be empty.")
+
+        return value
```
qbey:
```suggestion
    q = serializers.CharField(required=True, allow_blank=False)
```

You may also add `trim_whitespace=True`

src/backend/core/api/viewsets.py
```diff
@@ -997,6 +1001,38 @@ def duplicate(self, request, *args, **kwargs):
...
+        Returns a DRF response containing the filtered, annotated and ordered document list.
+        The filtering allows full text search through the opensearch indexation app "find".
+        """
+        access_token = request.session.get("oidc_access_token")
+
+        serializer = serializers.FindDocumentSerializer(data=request.query_params)
+        serializer.is_valid(raise_exception=True)
+
+        try:
+            indexer = FindDocumentIndexer()
```
qbey: I guess this class should come from settings, because not everyone will have an indexer. I also think this view might fallback on searching locally on title if no indexer is configured.

src/backend/core/services/search_indexers.py
```diff
@@ -0,0 +1,264 @@
...
+            data (dict): search data
+            token (str): OICD token
+
+        Returns:
+            dict: A JSON-serializable dictionary.
+        """
+        url = getattr(settings, "SEARCH_INDEXER_QUERY_URL", None)
+
+        if not url:
+            raise RuntimeError(
```
qbey:
```suggestion
            raise ImproperlyConfigured(
```

src/backend/core/services/search_indexers.py
```diff
@@ -0,0 +1,264 @@
...
+        Retrieve documents from the Find app API.
+
+        Args:
+            data (dict): search data
+            token (str): OICD token
+
+        Returns:
+            dict: A JSON-serializable dictionary.
+        """
+        url = getattr(settings, "SEARCH_INDEXER_QUERY_URL", None)
```
qbey:
```suggestion
        url = settings.SEARCH_INDEXER_QUERY_URL
```

src/backend/core/services/search_indexers.py
```diff
@@ -0,0 +1,264 @@
...
+                url,
+                json=data,
+                headers={"Authorization": f"Bearer {token}"},
+                timeout=10,
+            )
+            response.raise_for_status()
+            return response.json()
+        except requests.exceptions.HTTPError as e:
+            logger.error("HTTPError: %s", e)
+            logger.error("Response content: %s", response.text)  # type: ignore
```
qbey: Log the error only once

src/backend/core/services/search_indexers.py
```diff
@@ -0,0 +1,264 @@
...
+        Args:
+            data (dict): search data
+            token (str): OICD token
+
+        Returns:
+            dict: A JSON-serializable dictionary.
+        """
+        url = getattr(settings, "SEARCH_INDEXER_QUERY_URL", None)
+
+        if not url:
```
qbey: I would do this check in the `__init__` method (or if the document indexer is configurable by setting, in the settings directly: you can't use FindDocumentIndexer without defining the URLs, it should raise the sooner as possible)

src/backend/core/api/viewsets.py
```diff
@@ -997,6 +1001,38 @@ def duplicate(self, request, *args, **kwargs):
...
+            queryset = indexer.search(
+                text=serializer.validated_data.get("q", ""),
+                user=request.user,
+                token=access_token,
+            )
+        except RuntimeError:
```
qbey:
```suggestion
        except ImproperlyConfigured:
```

src/backend/core/api/viewsets.py
```diff
@@ -997,6 +1001,38 @@ def duplicate(self, request, *args, **kwargs):
...
+        access_token = request.session.get("oidc_access_token")
+
+        serializer = serializers.FindDocumentSerializer(data=request.query_params)
+        serializer.is_valid(raise_exception=True)
+
+        try:
+            indexer = FindDocumentIndexer()
+            queryset = indexer.search(
+                text=serializer.validated_data.get("q", ""),
+                user=request.user,
```
qbey: What is the purpose of passing the user here ? the token already does the job, no?

src/backend/core/models.py
```diff
@@ -1169,6 +1185,15 @@ def get_abilities(self, user):
         }
 
 
+@receiver(signals.post_save, sender=DocumentAccess)
```
qbey: We try to follow "Use signals as a last resort" (Two scoops of Django): is there a problem calling this from the `save` method?

src/backend/core/services/search_indexers.py
```diff
@@ -0,0 +1,264 @@
...
+            logger.error("Response content: %s", response.text)  # type: ignore
+            raise
+
+    def format_response(self, data: dict):
+        """
+        Retrieve documents ids from Find app response and return a queryset.
+        """
+        return models.Document.objects.filter(pk__in=[d["_id"] for d in data])
+
+    def push(self, data):
```
qbey, line 285: Same comments as before in this method

src/backend/core/tests/commands/test_index.py
```diff
@@ -0,0 +1,52 @@
...
+        factories.UserDocumentAccessFactory(document=no_title_doc, user=user)
+
+    accesses = {
+        str(doc.path): {"users": [user.sub]},
+        str(empty_doc.path): {"users": [user.sub]},
+        str(no_title_doc.path): {"users": [user.sub]},
+    }
+
+    def sortkey(d):
+        return d["id"]
```
qbey: Use https://docs.python.org/3/library/operator.html#operator.itemgetter instead

src/backend/core/tests/commands/test_index.py
```diff
@@ -0,0 +1,52 @@
...
+
+    with mock.patch.object(FindDocumentIndexer, "push") as mock_push:
+        call_command("index")
+
+        push_call_args = [call.args[0] for call in mock_push.call_args_list]
+
+        assert len(push_call_args) == 1  # called once but with a batch of docs
+        assert sorted(push_call_args[0], key=sortkey) == sorted(
```
qbey: Interesting, actually I think the documents sorting should be deterministic, in case we need tu run several times the index command => I think we should change the index command to sort documents by creation date or something ^^ 

We can keep it this way for now, but we surely need to add a comment in the "index" management command.

src/backend/core/tests/commands/test_index.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def sortkey(d):
+        return d["id"]
+
+    with mock.patch.object(FindDocumentIndexer, "push") as mock_push:
+        call_command("index")
+
+        push_call_args = [call.args[0] for call in mock_push.call_args_list]
+
+        assert len(push_call_args) == 1  # called once but with a batch of docs
```
qbey: Don't you want to simply check the assert_called_once, then use the first value?

qbey (commented):
First review, I know work is still ongoing and I did not read all the tests... :)

src/backend/core/api/viewsets.py
joehybird: it was to get the visited document ids. It could be done outside this method with `get_visited_document_ids_of(user)` in the view.

src/backend/core/tests/commands/test_index.py
joehybird: the sort is by id because the indexation is done by batch : loop + `id__gt=prev_batch_last_id`

src/backend/core/tests/documents/test_api_documents_search.py
```diff
@@ -0,0 +1,142 @@
...
+def test_api_documents_search_endpoint_is_none(settings):
+    """Missing SEARCH_INDEXER_QUERY_URL should throw an error"""
+    settings.SEARCH_INDEXER_QUERY_URL = None
+
+    user = factories.UserFactory()
+
+    client = APIClient()
+    client.force_login(user)
+
+    response = APIClient().get("/api/v1.0/documents/search/", data={"q": "alpha"})
```
qbey: Too much `APIClient()` ^^'

```
response = client.get
```

src/backend/core/tests/documents/test_api_documents_search.py
```diff
@@ -0,0 +1,142 @@
...
+
+    assert response.status_code == 401
```
qbey: Is it the expected status code for misconfiguration?

src/backend/core/tests/documents/test_api_documents_search.py
```diff
@@ -0,0 +1,142 @@
...
+def test_api_documents_search_invalid_params(settings):
+    """Validate the format of documents as returned by the search view."""
+    settings.SEARCH_INDEXER_QUERY_URL = "http://find/api/v1.0/search"
+
+    user = factories.UserFactory()
+
+    client = APIClient()
+    client.force_login(user)
+
+    response = APIClient().get("/api/v1.0/documents/search/")
```
qbey:
```suggestion
    client = APIClient()
    client.force_login(user)

    response = client.get("/api/v1.0/documents/search/")
```

src/backend/core/tests/documents/test_api_documents_search.py
```diff
@@ -0,0 +1,142 @@
...
+
+    assert response.status_code == 400
+    assert response.json() == {"q": ["This field is required."]}
+
+    response = APIClient().get("/api/v1.0/documents/search/", data={"q": "    "})
```
qbey:
```suggestion
    response = client.get("/api/v1.0/documents/search/", data={"q": "    "})
```

src/backend/core/api/viewsets.py
```diff
@@ -1003,6 +1007,38 @@ def duplicate(self, request, *args, **kwargs):
...
+            indexer = get_document_indexer_class()()
+            queryset = indexer.search(
+                text=serializer.validated_data.get("q", ""),
+                user=request.user,
+                token=access_token,
+            )
+        except ImproperlyConfigured:
+            return drf.response.Response(
+                {"detail": "The service is not properly configured."},
+                status=status.HTTP_401_UNAUTHORIZED,
```
qbey: Why a 401?

src/backend/impress/settings.py
```diff
@@ -938,6 +957,11 @@ class Test(Base):
     # Tests are raising warnings because the /data/static directory does not exist
     STATIC_ROOT = None
 
+    # Setup indexer configuration to make test working on the CI.
+    SEARCH_INDEXER_SECRET = "ThisIsAKeyForTest"  # noqa
+    SEARCH_INDEXER_URL = "http://localhost:8081/api/v1.0/documents/index/"
+    SEARCH_INDEXER_QUERY_URL = "http://localhost:8081/api/v1.0/documents/search/"
+
```
qbey: I'm not sure to understand. 

My two cents:
 - you should only set the settings in the tests which requires it
 - there must be a way to disable the indexation, because not every instance of docs will have an index, and it must be the default case.
 
 => I would remove those settings, determine which settings enables the indexation (I think SEARCH_INDEXER_CLASS should be None by default, and if None, then the document must not be indexed). Add new tests on the documents save, with SEARCH_INDEXER_CLASS not None, to check the indexation task etc.

src/backend/core/api/viewsets.py
joehybird: I agree... why ? A 500 is a better idea :sweat_smile:

src/backend/core/api/viewsets.py
```diff
@@ -1030,6 +1037,57 @@ def duplicate(self, request, *args, **kwargs):
...
+
+        serializer = serializers.FindDocumentSerializer(data=request.query_params)
+        serializer.is_valid(raise_exception=True)
+
+        indexer = default_document_indexer()
+
+        if not indexer:
+            queryset = self.get_queryset()
+            filterset = DocumentFilter(
+                {"title": serializer.validated_data.get("q", "")}, queryset=queryset
```
qbey: `q` is required, better be explicit about it.

```suggestion
                {"title": serializer.validated_data["q"], queryset=queryset
```

src/backend/core/api/viewsets.py
```diff
@@ -1030,6 +1037,57 @@ def duplicate(self, request, *args, **kwargs):
...
+        Returns a DRF response containing the filtered, annotated and ordered document list.
+        The filtering allows full text search through the opensearch indexation app "find".
+        """
+        access_token = request.session.get("oidc_access_token")
+        user = request.user
+
+        serializer = serializers.FindDocumentSerializer(data=request.query_params)
+        serializer.is_valid(raise_exception=True)
+
+        indexer = default_document_indexer()
```
qbey: nit: this is not a `default_document_indexer`, it's the `document_indexer` ^^

src/backend/core/api/viewsets.py
```diff
@@ -1030,6 +1037,57 @@ def duplicate(self, request, *args, **kwargs):
...
+
+        if not indexer:
```
qbey: Either add a comment to explain, or update the docstring to explain the behavior with and without an external indexer.

src/backend/core/management/commands/index.py
```diff
@@ -0,0 +1,31 @@
...
+class Command(BaseCommand):
+    """Index all documents to remote search service"""
+
+    help = __doc__
+
+    def handle(self, *args, **options):
+        """Launch and log search index generation."""
+        logger.info("Starting to regenerate Find index...")
+        start = time.perf_counter()
+        count = FindDocumentIndexer().index()
```
qbey: Should also use `default_document_indexer` and log a warning if the command is called without any external index

src/backend/core/services/search_indexers.py
```diff
@@ -0,0 +1,314 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@cache
+def default_document_indexer():
+    """Returns default indexer service is enabled and properly configured."""
+
+    # For this usecase an empty indexer class is not an issue but a feature.
+    if not getattr(settings, "SEARCH_INDEXER_CLASS", None):
```
qbey: The setting exists, be explicit about it.
```suggestion
    if not settings.SEARCH_INDEXER_CLASS:
```

src/backend/core/tasks/find.py
```diff
@@ -0,0 +1,88 @@
...
+    # Prevents some circular imports
+    # pylint: disable=import-outside-toplevel
+    from core import models  # noqa: PLC0415
+    from core.services.search_indexers import (  # noqa: PLC0415
+        get_batch_accesses_by_users_and_teams,
+        get_document_indexer_class,
+    )
+
+    doc = models.Document.objects.get(pk=document_id)
+    indexer = get_document_indexer_class()()
```
qbey: `default_document_indexer` should be used here, and is not defined it should add a warning and early return.

src/backend/core/services/search_indexers.py
```diff
@@ -0,0 +1,314 @@
...
+
+    try:
+        return get_document_indexer_class()()
+    except ImproperlyConfigured as err:
+        logger.error("Document indexer is not properly configured : %s", err)
+        return None
+
+
+@cache
+def get_document_indexer_class():
```
qbey: I would merge it in `default_document_indexer` because there is no use of having this function alone and it adds a level of cache that might be confusing.

src/backend/core/services/search_indexers.py
```diff
@@ -0,0 +1,314 @@
...
+
+    qs = models.LinkTrace.objects.filter(user=user)
+
+    docs = (
+        queryset.exclude(accesses__user=user)
+        .filter(
+            deleted_at__isnull=True,
+            ancestors_deleted_at__isnull=True,
+        )
+        .filter(pk__in=Subquery(qs.values("document_id")))
```
qbey, line 89: Does having twice `.filter` changes anything?

src/backend/core/services/search_indexers.py
```diff
@@ -0,0 +1,314 @@
...
+    )
+
+    return list({str(id) for id in docs.values_list("pk", flat=True)})
```
qbey: Why do you need to return a list instead of a set? Why don't directly do a `return docs.order_by("pk").values_list("pk", flat=True).distinct()` (to be tested, I'm not sure it's better in terms of performance)

src/backend/core/services/search_indexers.py
```diff
@@ -0,0 +1,314 @@
...
+
+    def index(self):
+        """
+        Fetch documents in batches, serialize them, and push to the search backend.
+        """
+        last_id = 0
+        count = 0
+
+        while True:
+            documents_batch = list(
```
qbey, line 146: Why do you need to use `list` here ?

src/backend/core/tasks/find.py
```diff
@@ -0,0 +1,88 @@
...
+    Trigger indexation task with debounce a delay set by the SEARCH_INDEXER_COUNTDOWN setting.
+
+    Args:
+        document (Document): The document instance.
+    """
+    if document.deleted_at or document.ancestors_deleted_at:
+        return
+
+    key = document_indexer_debounce_key(document.pk)
+    countdown = getattr(settings, "SEARCH_INDEXER_COUNTDOWN", 1)
```
qbey: Setting exists, be explicit
```suggestion
    countdown =settings.SEARCH_INDEXER_COUNTDOWN
```

src/backend/core/signals.py
```diff
@@ -0,0 +1,33 @@
...
+        transaction.on_commit(partial(trigger_document_indexer, instance))
+
+
+@receiver(signals.post_save, sender=models.DocumentAccess)
+def document_access_post_save(sender, instance, created, **kwargs):  # pylint: disable=unused-argument
+    """
+    Asynchronous call to the document indexer at the end of the transaction.
+    """
+    if not created and default_document_indexer() is not None:
+        transaction.on_commit(partial(trigger_document_indexer, instance.document))
```
qbey: I tried to remove the signals and place this in the `save` method and is works well (at least the tests does not raise any issue)

src/backend/core/tests/test_models_documents.py
```diff
@@ -1411,3 +1416,239 @@ def test_models_documents_compute_ancestors_links_paths_mapping_structure(
...
+    user = factories.UserFactory()
+
+    with transaction.atomic():
+        doc1, doc2, doc3 = factories.DocumentFactory.create_batch(3)
+
+        factories.UserDocumentAccessFactory(document=doc1, user=user)
+        factories.UserDocumentAccessFactory(document=doc2, user=user)
+        factories.UserDocumentAccessFactory(document=doc3, user=user)
+
+    time.sleep(0.2)  # waits for the end of the tasks
```
qbey: Isn't there a way with pytest-celerey (https://docs.celeryq.dev/en/latest/userguide/testing.html) to prevent such wait?

src/backend/core/tests/test_models_documents.py
```diff
@@ -1411,3 +1416,239 @@ def test_models_documents_compute_ancestors_links_paths_mapping_structure(
...
+    time.sleep(0.1)  # waits for the end of the tasks
+
+    with mock.patch.object(FindDocumentIndexer, "push") as mock_push:
+        # Simulate 1 waiting task
+        cache.set(document_indexer_debounce_key(doc.pk), 1)
+
+        # save doc to trigger the indexer, but nothing should be done since
+        # the counter is over 0
+        with transaction.atomic():
+            doc.save()
```
qbey: If there is not any change, should we call the indexer?

src/backend/core/tests/test_models_documents.py
```diff
@@ -1411,3 +1416,239 @@ def test_models_documents_compute_ancestors_links_paths_mapping_structure(
...
+@mock.patch.object(FindDocumentIndexer, "push")
+@pytest.mark.django_db(transaction=True)
+def test_models_documents_post_save_indexer(mock_push, indexer_settings):
+    """Test indexation task on document creation"""
+    indexer_settings.SEARCH_INDEXER_COUNTDOWN = 0
+
+    user = factories.UserFactory()
+
+    with transaction.atomic():
+        doc1, doc2, doc3 = factories.DocumentFactory.create_batch(3)
```
qbey: Don't you also want to test on document creation only (without access I mean?)

src/backend/core/signals.py
```diff
@@ -0,0 +1,33 @@
...
+    Asynchronous call to the document indexer at the end of the transaction.
+    Note : Within the transaction we can have an empty content and a serialization
+    error.
+    """
+    if default_document_indexer() is not None:
+        transaction.on_commit(partial(trigger_document_indexer, instance))
+
+
+@receiver(signals.post_save, sender=models.DocumentAccess)
+def document_access_post_save(sender, instance, created, **kwargs):  # pylint: disable=unused-argument
```
qbey, line 26: Deletion should also trigger new indexation

src/backend/core/tests/test_services_search_indexers.py
```diff
@@ -0,0 +1,568 @@
...
+    """
+    Import indexer class defined in setting SEARCH_INDEXER_CLASS.
+    """
+    indexer_settings.SEARCH_INDEXER_CLASS = (
+        "core.tests.test_services_search_indexers.FakeDocumentIndexer"
+    )
+
+    assert get_document_indexer_class() == import_string(
+        "core.tests.test_services_search_indexers.FakeDocumentIndexer"
+    )
```
qbey: Can it be simplified?
```suggestion
    assert get_document_indexer_class() == FakeDocumentIndexer
```

joehybird: For the assertion it is not exactly the same class :thinking: : an `app.core.tests.test_services_search_indexers.FakeDocumentIndexer` is returned by import_module instead of FakeDocumentIndex.

src/backend/core/services/search_indexers.py
joehybird: I agree it is not really useful with the new behaviour.

src/backend/core/services/search_indexers.py
joehybird: hum... `distinct('pk')` should be ok.

src/backend/core/api/viewsets.py
```diff
@@ -1030,6 +1037,68 @@ def duplicate(self, request, *args, **kwargs):
...
+    @drf.decorators.action(detail=False, methods=["get"], url_path="search")
+    @method_decorator(refresh_oidc_access_token)
+    def search(self, request, *args, **kwargs):
+        """
+        Returns a DRF response containing the filtered, annotated and ordered document list.
+
+        Applies filtering based on request parameter 'q' from `FindDocumentSerializer`.
+        Depending of the configuration it can be:
+         - A fulltext search through the opensearch indexation app "find" if the backend is
+           enabled (see SEARCH_BACKEND_CLASS)
```
```suggestion
           enabled (see SEARCH_INDEXER_CLASS)
```
I don't see the `SEARCH_BACKEND_CLASS` settings

src/backend/core/api/viewsets.py
```diff
@@ -1030,6 +1037,68 @@ def duplicate(self, request, *args, **kwargs):
...
+                raise drf.exceptions.ValidationError(filterset.errors)
+
+            queryset = filterset.filter_queryset(queryset).order_by("-updated_at")
+
+            return self.get_response_for_queryset(
+                queryset,
+                context={
+                    "request": request,
+                },
+            )
```
lunika: You can maybe extract this in a private method ?

src/backend/core/api/viewsets.py
```diff
@@ -1030,6 +1037,68 @@ def duplicate(self, request, *args, **kwargs):
...
+        )
+
+        queryset = queryset.filter(pk__in=results).order_by("-updated_at")
+
+        return self.get_response_for_queryset(
+            queryset,
+            context={
+                "request": request,
+            },
+        )
```
lunika, line 1135: Same here ?

src/backend/core/tasks/search.py
```diff
@@ -0,0 +1,89 @@
...
+    # check if the counter : if still up, skip the task. only the last one
+    # within the countdown delay will do the query.
+    if indexer_debounce_release(document_id) > 0:
+        logger.info("Skip document %s indexation", document_id)
+        return
+
+    indexer = get_document_indexer()
+
+    if indexer is None:
+        return
```
lunika: Can be do earlier to abort the task before checking the debounce ?

src/backend/core/services/search_indexers.py
```diff
@@ -0,0 +1,303 @@
...
+    grouped by users and teams, including all ancestor paths.
+    """
+    ancestor_map = utils.get_ancestor_to_descendants_map(
+        paths, steplen=models.Document.steplen
+    )
+    ancestor_paths = list(ancestor_map.keys())
+
+    access_qs = models.DocumentAccess.objects.filter(
+        document__path__in=ancestor_paths
+    ).values("document__path", "user__sub", "team")
```
lunika, line 54:
```suggestion
    ).values("document__path", "user__sub", "team").iterator() 
```

As it is used only in the task, maybe it is interested to use `.iterator()` ?

src/backend/core/tasks/search.py
```diff
@@ -0,0 +1,89 @@
+"""Trigger document indexation using celery task."""
```
lunika, line 1: You are mixing `find` and `search indexers` in the PR. IMO keeping `search_indexer` here make sense, this what you are doing. (for the module name)

src/backend/core/tests/commands/test_index.py
```diff
@@ -0,0 +1,65 @@
...
+        factories.UserDocumentAccessFactory(document=empty_doc, user=user)
+        factories.UserDocumentAccessFactory(document=no_title_doc, user=user)
+
+    accesses = {
+        str(doc.path): {"users": [user.sub]},
+        str(empty_doc.path): {"users": [user.sub]},
+        str(no_title_doc.path): {"users": [user.sub]},
+    }
+
+    with mock.patch.object(FindDocumentIndexer, "push") as mock_push:
```
lunika: Instead of mocking the `push`, why not using `responses` ?

src/backend/core/services/search_indexers.py
lunika: To force the queryset execution ?

src/backend/core/services/search_indexers.py
```diff
@@ -0,0 +1,303 @@
...
+            response = requests.post(
+                self.indexer_url,
+                json=data,
+                headers={"Authorization": f"Bearer {self.indexer_secret}"},
+                timeout=10,
+            )
+            response.raise_for_status()
+        except requests.exceptions.HTTPError as e:
+            logger.error("HTTPError: %s", e)
+            raise
```
lunika:
```suggestion
          response = requests.post(
              self.indexer_url,
              json=data,
              headers={"Authorization": f"Bearer {self.indexer_secret}"},
              timeout=10,
          )
          response.raise_for_status()
```

I would let the exception raises without catching it. Sentry is configured to handle it.

src/backend/core/services/search_indexers.py
```diff
@@ -0,0 +1,303 @@
...
+                self.search_url,
+                json=data,
+                headers={"Authorization": f"Bearer {token}"},
+                timeout=10,
+            )
+            response.raise_for_status()
+            return response.json()
+        except requests.exceptions.HTTPError as e:
+            logger.error("HTTPError: %s", e)
+            raise
```
lunika: Same here, no try/catch needed IMO

src/backend/core/tasks/search.py
joehybird: good point :+1:

src/backend/core/tasks/search.py
joehybird: we can do a `indexer_debounce_lock(document.pk)  > 1` to skip the job start in the `trigger_document_indexer` method. But I'm not sure about the side effects I have to think about it.

src/backend/core/models.py
joehybird: I tried to move the triggers within the model but it creates circular import issues (pylint is really stubborn on this point). Either we keep the signals or move some code of the model like `soft_delete` or `restore` outside the model.

src/backend/core/tests/test_models_documents.py
joehybird: I've changed the mechanism for the task so this is not an issue now.

src/backend/core/tests/test_models_documents.py
joehybird: I think it is easier to limit the number of queries than detect changes (can be just the permissions).

qbey (approved):

PR #1286, opened 2025-08-13
🔧(keycloak) Fix https required issue in dev mode.

joehybird:
## Purpose

On some environments keycloak returns a 'HTTPS required' message on login.
The same issue was fixed in drive by changing the 'sslRequired' value from 'external' to 'none' in all realms configuration (see https://github.com/suitenumerique/drive/issues/291).

Also upgrade keycloak up to 26.3.2

### DIFF
--- a/src/frontend/apps/e2e/__tests__/app-impress/home.spec.ts
+++ b/src/frontend/apps/e2e/__tests__/app-impress/home.spec.ts
@@ -131,7 +131,7 @@ test.describe('Home page', () => {
 
     // Keyclock login page
     await expect(
-      page.locator('.login-pf-page-header').getByText('impress'),
+      page.locator('.login-pf #kc-header-wrapper').getByText('impress'),
     ).toBeVisible();
   });
 });
--- a/src/frontend/apps/e2e/__tests__/app-impress/utils-common.ts
+++ b/src/frontend/apps/e2e/__tests__/app-impress/utils-common.ts
@@ -56,7 +56,7 @@ export const keyCloakSignIn = async (
   const password = `password-e2e-${browserName}`;
 
   await expect(
-    page.locator('.login-pf-page-header').getByText('impress'),
+    page.locator('.login-pf #kc-header-wrapper').getByText('impress'),
   ).toBeVisible();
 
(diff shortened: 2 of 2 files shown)


compose.yml
```diff
@@ -184,22 +184,21 @@ services:
...
     volumes:
       - ./docker/auth/realm.json:/opt/keycloak/data/import/realm.json
     command:
       - start-dev
       - --features=preview
       - --import-realm
-      - --proxy=edge
-      - --hostname-url=http://localhost:8083
-      - --hostname-admin-url=http://localhost:8083/
+      # - --proxy=edge
```
lunika:
```suggestion
```
If it's not used, can it be removed?

compose.yml
```diff
@@ -184,22 +184,21 @@ services:
...
+      - --hostname=http://localhost:8083
       - --hostname-strict=false
-      - --hostname-strict-https=false
       - --health-enabled=true
       - --metrics-enabled=true
     healthcheck:
-      test: ["CMD", "curl", "--head", "-fsS", "http://localhost:8080/health/ready"]
+      test: ['CMD-SHELL', '[ -f /tmp/HealthCheck.java ] || echo "public class HealthCheck { public static void main(String[] args) throws java.lang.Throwable { java.net.URI uri = java.net.URI.create(args[0]); System.exit(java.net.HttpURLConnection.HTTP_OK == ((java.net.HttpURLConnection)uri.toURL().openConnection()).getResponseCode() ? 0 : 1); } }" > /tmp/HealthCheck.java && java /tmp/HealthCheck.java http://localhost:9000/health/live']
```
lunika: I wonder if it's not easier to remove the healthcheck instead of having to maintain this?

joehybird: does it seems easier to maintain ? can be moved in a script.
`['CMD-SHELL', 'exec 3<>/dev/tcp/localhost/9000; echo -e "GET /health/ready/live HTTP/1.1\r\nHost: localhost\r\nConnection: close\r\n\r\n" >&3; grep "HTTP/1.1 200 OK" <&3']`

lunika (approved):

Issue #1283, opened 2025-08-11
Docs Sharing


ahofbeck:
## Bug Report

**Problematic behavior**
I set the whole doc with "share documents" to Link settings "connected" and "reading"
<img width="856" height="84" alt="Image" src="https://github.com/user-attachments/assets/649b3eec-20fe-4643-877d-ea0a9760391c" />

I want to give granular access for writing to a sub-page and their responding sub-pages.
<img width="276" height="345" alt="Image" src="https://github.com/user-attachments/assets/d695627f-2b8a-47f9-b87e-02e304c4b821" />

But changing from read to write and sharing the link is not allowed
<img width="983" height="248" alt="Image" src="https://github.com/user-attachments/assets/48ce486f-246e-41ac-aa16-b3ff3206ec04" />

**Expected behavior/code**
I am able to either invite people to sub-pages for writing or provide an edit link which I can remove after the page content was created.

**Steps to Reproduce**
1. set doc to "connected" and "read"
2. change sub share settings to write
3. setting is not allowed to be modified from read to write

**Environment**
- Docs version: 3.5.0
- Instance url: own deployment

AntoLC:
Thank you for sharing this issue 🙏 
A fixed is proposed here: https://github.com/suitenumerique/docs/pull/1287

ahofbeck:
Thank you! This helps a lot if this is solved. In the next iterations of Docs it would be interesting to have right-management for subpages account-based, so that one is able to specify editors for specific pages. But sharing the link already helps a lot. Keep up the good work here!

AntoLC:
Fine grained right-management are on the roadmap yes.

----

In waiting for the patch to be released, you still have a way to make it works.

<img width="737" height="202" alt="Image" src="https://github.com/user-attachments/assets/7658686e-dcc7-4ffc-8019-2ab13b23f1a9" />

Steps: 
- From "Connected" update to "Public"
- From "Public" update back to "Connected"
- Now you can switch from "Reading" to "Editing" without problem

PR #1287, opened 2025-08-13
🐛Link role update

AntoLC:
## Purpose

When a child document is `restricted`, but have a different computation (`authenticated`), the link role could be updated from `link-configuration` and gives a 200 response, but the change did not have any effect on the computation because of the `link_reach` being still `restricted`.

```
Request URL https://docs.numerique.gouv.fr/api/v1.0/documents/[DOC_ID]/link-configuration/
Request Method PUT
Payload:{"link_role": "editor"}
```

image

See: https://github.com/suitenumerique/docs/issues/1283

## Proposal

We added some validations steps to be sure the `link-configuration` can be updated correctly.
We adapted the front to follow the new types requirement on `link-configuration` (link reach mandatory).

 - [x] 🥅(backend) link role could be updated when restricted document
 - [x] 🩹(frontend) add computed_link_reach on PUT link-configuration

### DIFF
--- a/src/frontend/apps/e2e/__tests__/app-impress/doc-header.spec.ts
+++ b/src/frontend/apps/e2e/__tests__/app-impress/doc-header.spec.ts
@@ -434,7 +434,9 @@ test.describe('Doc Header', () => {
   test('it pins a document', async ({ page, browserName }) => {
     const [docTitle] = await createDoc(page, `Pin doc`, browserName);
 
-    await page.getByLabel('Open the document options').click();
+    await page
+      .getByRole('button', { name: 'Open the document options' })
+      .click();
 
     // Pin
     await page.getByText('push_pin').click();
@@ -453,11 +455,15 @@ test.describe('Doc Header', () => {
     await expect(leftPanelFavorites.getByText(docTitle)).toBeVisible();
 
     await row.getByText(docTitle).click();
-    await page.getByLabel('Open the document options').click();
+    await page
+      .getByRole('button', { name: 'Open the document options' })
+      .click();
 
     // Unpin
     await page.getByText('Unpin').click();
-    await page.getByLabel('Open the document options').click();
+    await page
+      .getByRole('button', { name: 'Open the document options' })
+      .click();
     await expect(page.getByText('push_pin')).toBeVisible();
 
     await page.goto('/');
--- a/src/frontend/apps/impress/src/features/docs/doc-management/types.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-management/types.tsx
@@ -60,7 +60,7 @@ export interface Doc {
   path: string;
   is_favorite: boolean;
   link_reach: LinkReach;
-  link_role: LinkRole;
+  link_role?: LinkRole;
   nb_accesses_direct: number;
   nb_accesses_ancestors: number;
   computed_link_reach: LinkReach;
--- a/src/frontend/apps/impress/src/features/docs/doc-management/utils.ts
+++ b/src/frontend/apps/impress/src/features/docs/doc-management/utils.ts
@@ -1,7 +1,7 @@
 import emojiRegex from 'emoji-regex';
 import * as Y from 'yjs';
 
-import { Doc, LinkReach, LinkRole } from './types';
+import { Doc, LinkReach } from './types';
 
 export const base64ToYDoc = (base64: string) => {
   const uint8Array = Buffer.from(base64, 'base64');
@@ -18,7 +18,7 @@ export const getDocLinkReach = (doc: Doc): LinkReach => {
   return doc.computed_link_reach ?? doc.link_reach;
 };
 
-export const getDocLinkRole = (doc: Doc): LinkRole => {
+export const getDocLinkRole = (doc: Doc): Doc['link_role'] => {
   return doc.computed_link_role ?? doc.link_role;
 };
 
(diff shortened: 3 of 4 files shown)


AntoLC:
I find a bit strange to have `link_role`  by default in "Reader" when the reach is by default "Restricted", should not be "None" ?

https://github.com/suitenumerique/docs/blob/95c4e066bcf64fb68b444877b3ef5c89ea37506a/src/backend/core/models.py#L370-L377

src/backend/core/api/serializers.py
```diff
@@ -493,6 +493,23 @@ class Meta:
             "link_reach",
         ]
 
+    def validate(self, attrs):
+        """Validate that link_role is compatible with link_reach."""
+        link_reach = attrs.get("link_reach", self.instance.link_reach)
+        link_role = attrs.get("link_role")
+
+        # If link_reach is restricted, link_role should not be set to anything meaningful
+        if link_reach == models.LinkReachChoices.RESTRICTED and link_role in [
```
sampaccoud: I think you should use the existing get_select_options method to ensure unicity of the logic. Also, should we add something to abilities for the frontend to know? This makes me wonder how did you bump into this and was it really a functional issue because actually, link_role is ignored and irrelevant when link_reach is restricted.

AntoLC: Ok, actually I was misleaded, the root problem is how `max_role` is computed when a child is still in a restricted type (after being created).
https://github.com/suitenumerique/docs/blob/077575283c7fe1e56050d97a87bfa31494d931ba/src/backend/core/choices.py#L106-L113

src/backend/core/choices.py
```diff
@@ -99,17 +99,17 @@ def get_equivalent_link_definition(ancestors_links):
...
-                if link["link_reach"] == max_reach
-            ),
-            key=LinkRoleChoices.get_priority,
-        )
+        eligible_roles = [
+            link["link_role"]
+            for link in ancestors_links
+            if (link["link_reach"] == max_reach or 
+                link["link_reach"] == LinkReachChoices.RESTRICTED)
+        ]
```
sampaccoud: I don't get it. The link rôle of a restricted document is irrelevant so why take it into account? Can you try explaining in PR description what is the functional problem you face?

src/backend/core/api/serializers.py
sampaccoud: If I understand the problem correctly from the issue, the backend is correct but the frontend should set link_role to authenticated AND link_role to edit for it to be overriden.... doing what tou propose would lead to side effects and the problem would still be the same downstream or for authenticated vs public...

sampaccoud: Said differently: you can't expect to override an authenticated computed role by setting a different role for another level of reach!!

Technically in the frontend,  you should use the computed link_reach to set your override and not the current document link reach

src/backend/core/choices.py
AntoLC: I tried to be more descriptive. 
By following the issue https://github.com/suitenumerique/docs/issues/1283, you can reproduce the problem in production.

src/backend/core/api/serializers.py
sampaccoud: And I agree that there should be a validation to make sure that only allowed combinations are passed as per get_select_options

AntoLC: > Technically in the frontend, you should use the computed link_reach to set your override and not the current document link reach

The front does not set the current document link reach.
```
Request URL https://docs.numerique.gouv.fr/api/v1.0/documents/[DOC_ID]/link-configuration/
Request Method PUT
Payload:{"link_role": "editor"}
```

But yes, if we force the `link_reach` in the payload with the computed one, the problem does not occur anymore.
I suppose `link_reach` should be mandatory so to avoid mismatch, plus yes a validation to not be able to have these  kind of payload: `{"link_reach": "restricted", "link_role": "editor"}`

AntoLC: I adapted the serializer validation with what was made initially, I think it covers all the cases. 
Feel free to take over.

src/backend/core/api/serializers.py
```diff
@@ -486,13 +486,77 @@ class LinkDocumentSerializer(serializers.ModelSerializer):
...
+                        "link_role": link_role,
+                        "link_reach": link_reach,
+                        "allowed_roles": ", ".join(allowed_roles)
+                        if allowed_roles
+                        else "none",
+                    }
+                }
+            )
+
+        return attrs
```
lunika, line 570:
```suggestion

        # Restricted reach: link_role must be None
        if link_reach == models.LinkReachChoices.RESTRICTED:
            if link_role is not None:
                raise serializers.ValidationError(
                    {
                        "link_role": (
                            "Cannot set link_role when link_reach is 'restricted'. "
                            "Link role must be null for restricted reach."
                        )
                    }
                )
            return attrs

        # Non-restricted: link_role must be in allowed roles
        if link_role not in allowed_roles:
            allowed_roles_str = ", ".join(allowed_roles) if allowed_roles else "none"
            raise serializers.ValidationError(
                {
                    "link_role": (
                        f"Link role '{link_role}' is not allowed for link reach '{link_reach}'. "
                        f"Allowed roles: {allowed_roles_str}"
                    )
                }
            )

        return attrs
```

PR #1295, opened 2025-08-22
🐛(back) allow / and = characters in user sub field

### LINKED ISSUE #1280: Docs incompatible with ADFS OIDC provider
## Bug Report

**Problematic behavior**
I was setting up docs for myself and decided to use the already setup windows server domain controller as the OIDC provider (via ADFS) for convenience. But after throwing all the services together and `compose up`ping everything I got stuck on not being able to login. Looking into container's logs I saw this:

<details>
<summary>Container logs</summary>

```
backend-1     | lasuite.oidc_login.backends INFO Creating user dlvyg+PkAtlCmoJtTqbT+W8AJ743J2PaovD5oe6mrHs=
backend-1     | django.request ERROR Internal Server Error: /api/v1.0/callback/
backend-1     | Traceback (most recent call last):
backend-1     |   File "/usr/local/lib/python3.13/site-packages/django/core/handlers/exception.py", line 55, in inner
backend-1     |     response = get_response(request)
backend-1     |   File "/usr/local/lib/python3.13/site-packages/django/core/handlers/base.py", line 197, in _get_response
backend-1     |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
backend-1     |   File "/usr/local/lib/python3.13/site-packages/django/views/generic/base.py", line 105, in view
backend-1     |     return self.dispatch(request, *args, **kwargs)
backend-1     |            ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
backend-1     |   File "/usr/local/lib/python3.13/site-packages/django/views/generic/base.py", line 144, in dispatch
backend-1     |     return handler(request, *args, **kwargs)
backend-1     |   File "/usr/local/lib/python3.13/site-packages/mozilla_django_oidc/views.py", line 124, in get
backend-1     |     self.user = auth.authenticate(**kwargs)
backend-1     |                 ~~~~~~~~~~~~~~~~~^^^^^^^^^^
backend-1     |   File "/usr/local/lib/python3.13/site-packages/django/views/decorators/debug.py", line 75, in sensitive_variables_wrapper
backend-1     |     return func(*func_args, **func_kwargs)
backend-1     |   File "/usr/local/lib/python3.13/site-packages/django/contrib/auth/__init__.py", line 114, in authenticate
backend-1     |     user = backend.authenticate(request, **credentials)
backend-1     |   File "/usr/local/lib/python3.13/site-packages/lasuite/oidc_login/backends.py", line 113, in authenticate
backend-1     |     user = super().authenticate(request, **kwargs)
backend-1     |   File "/usr/local/lib/python3.13/site-packages/mozilla_django_oidc/auth.py", line 326, in authenticate
backend-1     |     return self.get_or_create_user(access_token, id_token, payload)
backend-1     |            ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
backend-1     |   File "/usr/local/lib/python3.13/site-packages/lasuite/oidc_login/backends.py", line 261, in get_or_create_user
backend-1     |     user = self.create_user(claims)
backend-1     |   File "/usr/local/lib/python3.13/site-packages/lasuite/oidc_login/backends.py", line 277, in create_user
backend-1     |     user.save()
backend-1     |     ~~~~~~~~~^^
backend-1     |   File "/app/core/models.py", line 222, in save
backend-1     |     super().save(*args, **kwargs)
backend-1     |     ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
backend-1     |   File "/usr/local/lib/python3.13/site-packages/django/contrib/auth/base_user.py", line 65, in save
backend-1     |     super().save(*args, **kwargs)
backend-1     |     ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
backend-1     |   File "/app/core/models.py", line 103, in save
backend-1     |     self.full_clean()
backend-1     |     ~~~~~~~~~~~~~~~^^
backend-1     |   File "/usr/local/lib/python3.13/site-packages/django/db/models/base.py", line 1679, in full_clean
backend-1     |     raise ValidationError(errors)
backend-1     | django.core.exceptions.ValidationError: {'sub': ['Enter a valid sub. This value may contain only letters, numbers, and @/./+/-/_/: characters.']}
```

</details>

After doing a bunch of research I found out that ADFS returns a hash value of a set of properties in the `sub` claim, which always contains a `=` char and sometimes others (like `/`, etc). And it's not possible to alter this behavior on the ADFS side as far as I understand, since it's meant to be "stable"

**Expected behavior/code**
ADFS OIDC Just Works™️

**Steps to Reproduce**
1. Set up a Windows Server domain controller with ADFS as a OIDC provider
2. Configure Docs to use the said DC for open id auth
3. Navigate to the Docs & try to login
4. After successfully loggin in & returning to the Docs observe a 500 response code & errors in the `backend` container's logs 

**Environment**
- Docs version: v3.5.0
- Instance url: private

**Possible Solution**
Update validations for the `sub` property to allow usage of `=`, `/` and other special characters that may appear in the `sub` claim

lunika:
## Purpose

We want to keep a restricted list of allowed characters in the user sub field. We allow now the = and /

## Proposal

- [x] 🐛(back) allow / and = characters in user sub field

Fix #1280

### DIFF
--- a/src/backend/core/models.py
+++ b/src/backend/core/models.py
@@ -137,21 +137,21 @@ class User(AbstractBaseUser, BaseModel, auth_models.PermissionsMixin):
     """User model to work with OIDC only authentication."""
 
     sub_validator = validators.RegexValidator(
-        regex=r"^[\w.@+-:]+\Z",
+        regex=r"^[\w.@+-:=/]+\Z",
         message=_(
             "Enter a valid sub. This value may contain only letters, "
-            "numbers, and @/./+/-/_/: characters."
+            "numbers, and @.+-_:=/ characters."
         ),
     )
 
     sub = models.CharField(
         _("sub"),
         help_text=_(
-            "Required. 255 characters or fewer. Letters, numbers, and @/./+/-/_/: characters only."
+            "Required. 255 characters or fewer. Letters, numbers, and @.+-_:=/ characters only."
         ),
         max_length=255,
-        unique=True,
         validators=[sub_validator],
+        unique=True,
         blank=True,
         null=True,
     )
--- a/src/backend/core/tests/test_models_users.py
+++ b/src/backend/core/tests/test_models_users.py
@@ -44,3 +44,27 @@ def test_models_users_send_mail_main_missing():
         user.email_user("my subject", "my message")
 
     assert str(excinfo.value) == "User has no email address."
+
+
+@pytest.mark.parametrize(
+    "sub,is_valid",
+    [
+        ("valid_sub.@+-:=/", True),
+        ("invalid sub", False),
+    ],
+)
+def test_models_users_sub_validator(sub, is_valid):
+    """The "sub" field should be validated."""
+    user = factories.UserFactory()
+    user.sub = sub
+    if is_valid:
+        user.full_clean()
+    else:
+        with pytest.raises(
+            ValidationError,
+            match=(
+                "Enter a valid sub. This value may contain only letters,"
+                " numbers, and @.+-_:=/ characters."
+            ),
+        ):
+            user.full_clean()


src/backend/core/models.py
```diff
@@ -137,21 +137,21 @@ class User(AbstractBaseUser, BaseModel, auth_models.PermissionsMixin):
...
-            "numbers, and @/./+/-/_/: characters."
+            "numbers, and @.+-_:=/ characters."
         ),
     )
 
     sub = models.CharField(
         _("sub"),
         help_text=_(
-            "Required. 255 characters or fewer. Letters, numbers, and @/./+/-/_/: characters only."
+            "Required. 255 characters or fewer. Letters, numbers, and @.+-_:=/ characters only."
```
qbey: Actually the OIDC spec says the sub can contain up to 255 ASCII characters (following RFC 20 https://www.rfc-editor.org/rfc/rfc20.txt)


> Subject Identifier. A locally unique and never reassigned identifier within the Issuer for the End-User, which is intended to be consumed by the Client, e.g., 24400320 or AItOawmwtWwcT0k51BayewNvutrJUqsvl6qs7A4. It MUST NOT exceed 255 ASCII [[RFC20]](https://openid.net/specs/openid-connect-core-1_0.html#RFC20) characters in length. The sub value is a case-sensitive string.

Should we be less restrictive?

qbey (commented):
Please check:

- [ ] add a changelog?
- [ ] update migrations (no-op)?

We can merge it like this, but I wonder whether we should be less restrictive on the `sub` field, wdyt?

src/backend/core/models.py
lunika: So we should remove the validator?

lunika:
@qbey ASCII characters are allowed now.

PR #1297, opened 2025-08-22
🐛(backend) allow creator to delete subpages

### LINKED ISSUE #1193: ♻️(backend) Editor can delete subpages
## Problem

As it is now an editor can create a subpage, but only a owner can delete it (even an admin cannot).

## Mission

The requested comportement is "_if I can create I can delete_", so an editor should be able to delete a subpage. If the doc is public editable, anybody (even not authenticated users) should be able to delete a subpage.

To know what to do, the frontend follow the `destroy` ability :
https://github.com/suitenumerique/docs/blob/ca09f9a158062e2d589a4b559e4950dd13663ade/src/frontend/apps/impress/src/features/docs/doc-tree/components/DocTreeItemActions.tsx#L94

lunika:
## Purpose

An editor who created a subpages should be allowed to delete it.
We change the abilities to be coherent between the creation and the
deletion.

## Proposal

- [x] 🐛(backend) allow editor to delete subpages

Fixes #1193

### DIFF
--- a/src/backend/core/models.py
+++ b/src/backend/core/models.py
@@ -753,6 +753,12 @@ class Document(MP_Node, BaseModel):
         can_update = (
             is_owner_or_admin or role == RoleChoices.EDITOR
         ) and not is_deleted
+        can_create_children = can_update and user.is_authenticated
+        can_destroy = (
+            is_owner
+            if self.is_root()
+            else (is_owner_or_admin or (user.is_authenticated and self.creator == user))
+        )
 
         ai_allow_reach_from = settings.AI_ALLOW_REACH_FROM
         ai_access = any(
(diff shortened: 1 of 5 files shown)


sampaccoud:
Good point as the owner can still restore from trashbin...

AntoLC (approved):
Working as expected, but I find it is lot of power for the editors, I am wondering if an editor should be able to delete only his own subpages.

```py
can_destroy = is_owner or is_admin or is_creator 
```

virgile-dev:
@AntoLC I agree with you ! Let’s go for what you proposed

lunika:
The implementation I made changed a bit. I think it should be reviewed again.
@qbey can you take a look please ? I made some modification on the document queryset, I would like to have your input.

src/backend/core/models.py
```diff
@@ -762,6 +762,12 @@ def get_abilities(self, user):
         can_update = (
             is_owner_or_admin or role == RoleChoices.EDITOR
         ) and not is_deleted
+        can_create_children = can_update and user.is_authenticated
+        can_destroy = (
+            is_owner
+            if self.is_root()
```
qbey, line 759: I don't understand, why admin cannot delete root documents (and I don't see the related test)?

lunika: This is the legacy behavior. And the related test is here https://github.com/suitenumerique/docs/blob/main/src/backend/core/tests/documents/test_api_documents_delete.py#L50

Issue #156, opened 2024-08-02
✨Comments


AntoLC:
## Feature Request

Can make comments on text of the editor.
Check with designer or other similar products how to integrate this comment feature.

virgile-dev:
Erwan Legal recommended Hypothes.is. 
https://github.com/hypothesis (W3C Web anotation) compatible. 
Comments are stored in an external database. 
The service could be used in other softwares.

virgile-dev:
@rl-83 sounds great thanks for the great work !

virgile-dev:
Comments are out in the latest BlockNotejs release. Get ready for comments ! https://github.com/TypeCellOS/BlockNote/releases/tag/v0.25.0

lunika:
We have to design the backend part first. We want to store the comment in the backend application

lunika:
The last Figma version is available here: https://www.figma.com/design/qdCWR4tTUr7vQSecEjCyqO/DOCS?node-id=9286-15611

For the backend part of this issue, this design should be applied:

- A new model `Comment` should be added. This model is linked to a user (but nullable, the feature is accessible to anonymous users) and to a document.
- A new role `commentator` (or `reviewer` ?) should be added. This role priority is between `virewer` and `editor` (`viewer` < `commentator` < `editor`)
- A `DocumentCommentViewset` viewset should be added. This item will be related to a document (add it to the `document_related_router` router).
- The model should implement the `get_abilities` method to use the `ResourceWithAccessPermission` permission.
- As the `comment` role is also available in the `link_role`, this viewset is accessible to anonymous users. The availability to access it should be determined in the `get_abilities` method of the `Comment` model.

The comment documentation for BlockNote is available here: https://www.blocknotejs.org/docs/features/collaboration/comments

PR #1309, opened 2025-08-28
✨(backend) Comments on text editor

lunika:
## Purpose

Describe in #156 this PR add the comment feature in the backend application.

A user, authenticated or not, depending on the link_reach and link_role configuration, can add one or more comment to a Document.
A new role and link_role `commentator` are also added. This role is higher than `reader` but lower than `editor`.
If the link_reach is public and the link_role is commentator, then anonymous user can leave a comment.

Missing:
- Compatilibility check with what blocknote is expected.
- Is it the back or the front that is managing comment reply ?
- What other properties than `content` is expected ?

## Proposal

- [x] ✨(backend) add comment viewset
- [x]  ✨(backend) add Comment model
- [x]  ✨(backend) add commentator role


lunika:
Commits are integrated in #1330 and I will continue my work in it.

PR #1315, opened 2025-08-29
👷(CI) trigger bundle-size-check only if necessary

AntoLC:
## Purpose

We don't need to run the bundle-size-check job if the app didn't change.
Only if the `yarn.lock` file or the app have changed, the bundle-size-check job will be triggered.


lunika:
Nice, we can probably use this github action for many others jobs. Like linting and running back only if the backend has changed, etc.

PR #1323, opened 2025-09-01
🔧(ci) always run all git-lint steps

lunika:
## Purpose

git-lint steps are independant and we would like to have all checks at once. Using the `if: always()` instruction should ensure all steps should be run event if the previous fails.

## Proposal

- [x] 🔧(ci) always run all git-lint steps

### DIFF
--- a/.github/workflows/impress.yml
+++ b/.github/workflows/impress.yml
@@ -25,14 +25,18 @@ jobs:
       - name: show
         run: git log
       - name: Enforce absence of print statements in code
+        if: always()
         run: |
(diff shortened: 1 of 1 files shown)


lunika:
@olaurendeau Hi, I made this PR to take into account your feedback. 
Here all steps are always run, you have all the info at once.
I made a fixup commit, with a print message. And an other commit that does not match gitlint rules. 
WDYT ?

olaurendeau:
Hey @lunika 
Nice thank you :)
I understand that "Merge" commit are not allowed in the process, should this also be checked by the CI to avoid @AntoLC to explain it to contributors 😅

PR #1330, opened 2025-09-02
Feat/comments frontend

AntoLC:
## Purpose

Implementation of the comment feature frontend side.


Ovgodd (commented):
Great work  the feature works really well!
it would be nice to move the focus directly into the “Write a comment” field when the comment panel opens, so keyboard users can start typing immediately.

Also, once a comment is created, it should be possible to reopen the comment thread using the keyboard. Wrapping the commented text in a focusable element could help with that ? wdyt

src/frontend/apps/impress/src/features/docs/doc-editor/components/comments/CommentToolbarButton.tsx
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  return (
+    <Box $direction="row" className="--docs--comment-toolbar-button">
+      <Components.Generic.Toolbar.Button
+        className="bn-button"
+        onClick={() => {
+          editor.comments?.startPendingComment();
+        }}
+        isDisabled={hasActiveUnresolvedThread}
```
Ovgodd: a little a11y suggestion
// Announce that this button opens a dialog-like panel (comment thread)
// Helps screen readers understand the interaction pattern
```
aria-haspopup="dialog"
aria-expanded={false} // updated to true when the comment panel is open
```

AntoLC: I added `aria-haspopup="dialog"`, but there is not an easy way to know when the dialog open or close (aria-expanded), I can catch the open on click, but when the user click out of the thread, he can close the dialog, and I don't have a easy way to catch that, the only way I found is to had a observer mechanism on the dialog, I find it is a bit overkill.


https://github.com/suitenumerique/docs/blob/902a1e9c862decc13b55e5b9c4f1a51b80305d2f/src/frontend/apps/impress/src/features/docs/doc-editor/components/comments/CommentToolbarButton.tsx#L38

PR #1332, opened 2025-09-03
🐛(frontend) fix display bug on homepage

AntoLC:
## Purpose

A section in the homepage was not displaying correctly anymore. This commit fixes the issue.

## Demo

image

### DIFF
--- a/src/frontend/apps/impress/src/features/home/components/HomeSection.tsx
+++ b/src/frontend/apps/impress/src/features/home/components/HomeSection.tsx
@@ -126,6 +126,7 @@ export const HomeSection = ({
             {title}
           </Text>
           <Text
+            as="div"
             $variation="700"
             $weight="400"
             $size={isSmallMobile ? 'ml' : 'md'}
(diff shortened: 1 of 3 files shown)


src/frontend/apps/impress/src/features/home/components/HomeSection.tsx
```diff
@@ -126,6 +126,7 @@ export const HomeSection = ({
             {title}
           </Text>
           <Text
+            as="div"
```
Ovgodd, line 129: I would recommend using a `p` tag instead of a `div` for better semantics.
If you want to keep the same visual style without the default margin applied to `p`, you can simply add:
wdyt ? 
```suggestion
           $css={css`
              margin: 0;
            `}
            as="p"
```

Ovgodd (changes requested):
Just a little thing ;)

src/frontend/apps/impress/src/features/home/components/HomeSection.tsx
AntoLC: I don't think we should use a "p" tag here, because it contains "div" and others "p" tag in it.

image

Ovgodd: Ahh yes, I agree with you, div is a better choice.

Ovgodd (approved):
good!

PR #1333, opened 2025-09-03
⚡️(frontend) improve export table width calculation

AntoLC:
## Purpose

Sometimes we do not have the width of some columns in a table. In such cases, we need to calculate a fallback width to ensure the table is rendered correctly.

## Proposal

We were previously using 120 points as the fallback width, but this has been improved to better fit the content.
We now check the size left and distribute it among the unknown columns.

### Before:
image

### After
image


virgile-dev:
Linking the BlockNote issue I opened : https://github.com/TypeCellOS/BlockNote/issues/1979

PR #1340, opened 2025-09-05
♻️(tilt) use hem dev-backend chart

lunika:
## Purpose

Remove usage of bitnami charts and use our own dev-backend charts instead.

## Proposal

- [X] ♻️(tilt) use hem dev-backend chart


rouja (approved):
LGTM

PR #1343, opened 2025-09-05
🔒️(backend) configure throttle on every viewsets

lunika:
## Purpose
We want to configure the throttle on all doc's viewsets. In order to
monitor them, we use the MonitoredScopedRateThrottle class and a custom
callback capturing the message in sentry at the warning level.

## Proposal

- [x] 🔒️(backend) configure throttle on every viewsets
- [x] ⬆️(backend) upgrade django-lasuite to version 0.0.14


qbey (approved):

src/backend/impress/settings.py
```diff
@@ -367,8 +370,46 @@ class Base(Configuration):
...
+            "document_ask_for_access": values.Value(
+                default="30/minute",
+                environ_name="API_DOCUMENT_ASK_FOR_ACCESS_THROTTLE_RATE",
+                environ_prefix=None,
+            ),
+            "config": values.Value(
+                default="30/minute",
+                environ_name="API_CONFIG_THROTTLE_RATE",
+                environ_prefix=None,
+            ),
```
AntoLC, line 407: Could this be in this list as well ? 
https://github.com/suitenumerique/docs/blob/0ac9f059b6212781000c3d490d1ae7885cdef067/src/backend/impress/settings.py#L662-L671

Issue #899, opened 2025-04-22
Modal


cyberbaloo:
**Problematic behavior** 

The modal close button has no label.
The modal text must be structured in a `<p>` element and should not have an `aria-label` attribute.
The modal has no title.
The labels for the "cancel" and "delete" buttons are not meaningful and could be confusing.

**Expected behavior/code**
`<div class="ReactModal__Content ReactModal__Content--after-open c__modal c__modal--medium" tabindex="-1" role="dialog" aria-modal="true">`

**Possible Solution**

Add an attribute `aria-label="Fermer la fenêtre"`
Structure the modal text in a `<p>` element and remove the `aria-label`

Add a h1 title with an attribut `id` for this window and add an attribute `aria-describedby="content_of_atttribute_id_h1"` 

Replace the content of the attribute `aria-label="Fermer la modale"` by `"Annuler la suppression"`
For the second button, replace with `Supprimer le document"`

<img width="630" alt="Image" src="https://github.com/user-attachments/assets/832ffa22-08b9-4dd8-97f8-15a8a199eeb9" />

PR #1349, opened 2025-09-08
✨(frontend) improve modal a11y: structure, labels, and title

Ovgodd:
## Purpose

Improve the accessibility of the modal by structuring content properly, adding a descriptive title, and clarifying button labels.

issue : [899](https://github.com/suitenumerique/docs/issues/899)

## Proposal
- [x]  Structure modal text inside a p element
- [x]  Add a h1 title with id and reference it via aria-describedby
- [x]  Replace ambiguous aria-labels for "cancel" and "delete" buttons
- [x]  Replace ambiguous aria-labels for "cancel" and "delete" buttons
- [x] Add a sr-only class to visually hide the modal title while keeping it accessible to screen readers

The sr-only class hides content from visual display but keeps it readable by screen readers. This ensures assistive technologies can announce the modal's purpose without affecting the UI layout.


src/frontend/apps/impress/src/features/docs/doc-export/components/ModalExport.tsx
```diff
@@ -165,6 +166,8 @@ export const ModalExport = ({ onClose, doc }: ModalExportProps) => {
           $width="100%"
         >
           <Text
+            as="h1"
```
AntoLC, line 169: When you change to `h1`, it get automatically a margin, so to keep the modal as it was before, you must adapt the margin as well.
Same for all the other components where `h1` was added.

```suggestion
            as="h1"
            $margin="0"
```

Ovgodd: Ok ! noted

PR #1356, opened 2025-09-09
🐛(backend) fix wrong permission check on sub page duplicate action [DO NOT MERGE]

### LINKED ISSUE #1329: wrong permission check on sub page duplicate action
**Problematic behavior**

A user with only read access to a document and it sub documents should not be able to duplicate a sub document in the document tree.
Moreover, it should not become owner of the duplicated sub document

**Expected behavior/code**

A user with read role should not be allowed to duplicate a sub document.

**Steps to Reproduce**

https://github.com/user-attachments/assets/defe12f8-b072-4a0b-80c1-91d77a3ade26

lunika:
## Purpose

A user with only read access to a document and its sub documents should not be able to duplicate a sub document in the document tree. To avoid this we have to compute the direct parent abilities to determine if it can create a new children. Doing so computes the abilities in cascade from the document to the root tree. To save some queries and not compute again and again the same ability, we create a abilities cache. The hard part is to invalidate this cache. The cache is invalidated if a related DocumentAccess instance is updated or deleted and also if the document link_reach or link_role is updated. To introspect the modification made on the model itself when the user saves it, we decided to use the library django-dirtyfields

## Proposal

- [x] 🐛(backend) fix wrong permission checj on sub page duplicate action

Fixes #1329

### DIFF
--- a/src/backend/core/models.py
+++ b/src/backend/core/models.py
@@ -404,6 +404,8 @@ class Document(DirtyFieldsMixin, MP_Node, BaseModel):
 
     objects = DocumentManager()
 
+    FIELDS_TO_CHECK = ["link_reach", "link_role"]
+
     class Meta:
         db_table = "impress_document"
         ordering = ("path",)
@@ -434,7 +436,7 @@ class Document(DirtyFieldsMixin, MP_Node, BaseModel):
         invalidate_abilities_cache = False
         if not self._state.adding:
             changed_fields = self.get_dirty_fields()
-            if "link_reach" in changed_fields or "link_role" in changed_fields:
+            if any(field in changed_fields for field in self.FIELDS_TO_CHECK):
                 invalidate_abilities_cache = True
 
         super().save(*args, **kwargs)
@@ -482,7 +484,7 @@ class Document(DirtyFieldsMixin, MP_Node, BaseModel):
         if self.is_root():
             return Document.objects.none()
 
-        paths = [self.path[0:pos] for pos in range(0, len(self.path), self.steplen)[1:]]
+        paths = [self.path[0:pos] for pos in range(self.steplen, len(self.path), self.steplen)]
         return (
             Document.objects.select_related("creator")
             .filter(path__in=paths)
@@ -759,13 +761,13 @@ class Document(DirtyFieldsMixin, MP_Node, BaseModel):
         """Actual link role on the document."""
         return self.computed_link_definition["link_role"]
 
-    def _compute_abilities_cache_key(self):
+    def _get_abilities_cache_key(self):
         """Generate a unique cache key for each document."""
         return f"document:abilities:{self.path!s}"
 
     def _get_abilities_cache_for_user(self, user):
         """Return the abilities cache for the document and user."""
-        key = self._compute_abilities_cache_key()
+        key = self._get_abilities_cache_key()
         document_abilities = cache.get(key, {})
 
         user_id = user.id if user.is_authenticated else "anonymous"
@@ -774,7 +776,7 @@ class Document(DirtyFieldsMixin, MP_Node, BaseModel):
 
     def _set_abilities_cache_for_user(self, user, abilities):
         """Set the abilities cache for the document and user."""
-        key = self._compute_abilities_cache_key()
+        key = self._get_abilities_cache_key()
         document_abilities = cache.get(key, {})
         user_id = user.id if user.is_authenticated else "anonymous"
         document_abilities[user_id] = abilities
@@ -782,7 +784,7 @@ class Document(DirtyFieldsMixin, MP_Node, BaseModel):
 
     def invalidate_abilities_cache(self):
         """Invalidate the abilities cache for the document."""
-        key = self._compute_abilities_cache_key()
+        key = self._get_abilities_cache_key()
         cache.delete(key)
 
         # Invalidate in cascade the abilities for all children
--- a/src/backend/core/models.py
+++ b/src/backend/core/models.py
@@ -28,6 +28,7 @@ from django.utils.translation import get_language, override
 from django.utils.translation import gettext_lazy as _
 
 from botocore.exceptions import ClientError
+from dirtyfields import DirtyFieldsMixin
 from rest_framework.exceptions import ValidationError
 from timezone_field import TimeZoneField
 from treebeard.mp_tree import MP_Node, MP_NodeManager, MP_NodeQuerySet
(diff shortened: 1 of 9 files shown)


src/backend/core/models.py
```diff
@@ -353,7 +354,7 @@ def get_queryset(self):
 
 
 # pylint: disable=too-many-public-methods
-class Document(MP_Node, BaseModel):
+class Document(DirtyFieldsMixin, MP_Node, BaseModel):
```
qbey, line 357: Don't you want to define the `FIELDS_TO_CHECK` attribute?

https://django-dirtyfields.readthedocs.io/en/stable/customisation.html#checking-a-limited-set-of-model-fields

src/backend/core/models.py
```diff
@@ -463,6 +474,42 @@ def is_leaf(self):
...
+        """
+        if self.is_root():
+            return Document.objects.none()
+
+        paths = [self.path[0:pos] for pos in range(0, len(self.path), self.steplen)[1:]]
+        return (
+            Document.objects.select_related("creator")
+            .filter(path__in=paths)
+            .order_by("depth")
+        )
```
qbey: Hmmm.... `range(0, len(self.path), self.steplen)[1:]` is actually `range(self.steplen, len(self.path), self.steplen)`?

src/backend/core/models.py
```diff
@@ -463,6 +474,42 @@ def is_leaf(self):
         """
         return not self.has_deleted_children and self.numchild == 0
 
+    def get_ancestors(self):
```
qbey, line 479: Why do you need to override the base `MP_Node` method ?

src/backend/core/models.py
```diff
@@ -426,11 +427,21 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._ancestors_link_definition = None
         self._computed_link_definition = None
+        self._cached_parent_obj = None
```
qbey, line 432: Why do you need to override the MP_Node attribute?

src/backend/core/models.py
```diff
@@ -712,10 +759,44 @@ def computed_link_role(self):
         """Actual link role on the document."""
         return self.computed_link_definition["link_role"]
 
+    def _compute_abilities_cache_key(self):
```
qbey: nit: "compute" what a big word for a simple function... "get"? or even simply a property?

src/backend/core/tests/documents/test_api_documents_children_list.py
```diff
@@ -267,10 +267,10 @@ def test_api_documents_children_list_authenticated_public_or_authenticated_paren
     child1, child2 = factories.DocumentFactory.create_batch(2, parent=document)
     factories.UserDocumentAccessFactory(document=child1)
 
-    with django_assert_num_queries(10):
+    with django_assert_num_queries(17):
```
qbey, line 270: And it's under control? ^^'

src/backend/core/tests/documents/test_api_documents_retrieve.py
```diff
@@ -500,7 +500,7 @@ def test_api_documents_retrieve_authenticated_related_parent():
             "cors_proxy": True,
             "content": True,
             "destroy": access.role in ["administrator", "owner"],
-            "duplicate": True,
+            "duplicate": access.role != "reader",
```
qbey, line 503: Reading the test, I don't even know if it's what we want or not... this has become too complexe for me ^^'

src/backend/impress/settings.py
```diff
@@ -808,6 +808,12 @@ class Base(Configuration):
         ),
     }
 
+    DOCUMENT_ABILITIES_CACHE_TIMEOUT = values.IntegerValue(
+        default=60 * 60,  # 1 hour
```
qbey, line 812: It could even be 12 hours, no? Leave it like that, we will fine tune in deployment

src/backend/core/tests/documents/test_api_documents_children_list.py
```diff
@@ -98,7 +98,7 @@ def test_api_documents_children_list_anonymous_public_parent(django_assert_num_q
 
     with django_assert_num_queries(9):
         APIClient().get(f"/api/v1.0/documents/{document.id!s}/children/")
-    with django_assert_num_queries(5):
+    with django_assert_num_queries(4):
```
qbey, line 101: Why?

src/backend/core/tests/documents/test_api_documents_children_list.py
```diff
@@ -187,7 +187,7 @@ def test_api_documents_children_list_authenticated_unrelated_public_or_authentic
     child1, child2 = factories.DocumentFactory.create_batch(2, parent=document)
     factories.UserDocumentAccessFactory(document=child1)
 
-    with django_assert_num_queries(9):
+    with django_assert_num_queries(11):
```
qbey, line 190: Why?

src/backend/core/tests/documents/test_api_documents_children_list.py
lunika: The abilities are in cache, there is no need to make a query to retrieve the roles in the accesses

src/backend/core/models.py
lunika: Ok I will change it for `get`.

src/backend/core/tests/documents/test_api_documents_children_list.py
lunika: With out the cache it will be 36 :man_shrugging:

src/backend/core/tests/documents/test_api_documents_children_list.py
lunika: This number of queries depends on the number of children. There is one  more query for each child.

src/backend/core/models.py
lunika: To add the `select_releated("creator")` to save one query. The creator is used in the abilities.

src/backend/core/models.py
lunika: To satisfy pylint. And `_cached_parent_obj` is not declare in the parent class, it is created "on the fly" in the `get_parent` method.

sampaccoud:
The philosophy behind the duplicate feature is: "if you can see it you can duplicate it". 
This is to avoid tricky copies which were raising bugs (media access, etc) and removing the possibility to link copies.
So I don't think this is a bug and i think we should not merge this PR.

More generally on the implementation, heritage is more complex than looking at the parent and caching get_abilities is more complicated than proposed here. I think it would trigger tricky bugs, and complicate group sharing. Also, the heavy lifting is done in the queryset and the get_abilities method is not where the time is spent so I think we should cache higher in the view if one day we do. But that's far from necessary with our current traffic so let's keep it live until we face a problem ?

PR #1361, opened 2025-09-10
💄(frontend) remove margin from modal title

AntoLC:
## Purpose

Recent improvement changes the modal title with a h1 tag, h1 tag adds margin by default.
We remove the margin from the h1 tag to stick to the design system.

## Demo

image


Ovgodd (approved):
💯

PR #1377, opened 2025-09-12
💬(copy) wrong permission mentionned on move doc

virgile-dev:
## Purpose

To move a doc you have to be at least admin of the destination doc.
It was written editor in the alert banner.

## Proposal
Change editor for administrator

## External contributions


virgile-dev:
In the making of the PR I was thinking that the current permission setting is confusing.
We allow editors to create sub-docs why not allow them to move docs inside a document where they are editors. I feel we should change the permissions not the message. @lunika

lunika:
> In the making of the PR I was thinking that the current permission setting is confusing. We allow editors to create sub-docs why not allow them to move docs inside a document where they are editors. I feel we should change the permissions not the message. @lunika

True. But it depends on the target position maybe ? Should we allow the "Move to my docs" for editors too ?
We should probably take the time to think about other situation where an admin role is needed and you can already do more or less the same thing as an editor.

lunika (approved):

PR #1385, opened 2025-09-15
🐛(backend) duplicate sub docs as root for reader user

lunika:
## Purpose

Reader user should be able to duplicate a doc in the doc tree. It should be created a new doc at the root level.

## Proposal

- [x] 🐛(backend) duplicate sub docs as root for reader user

Fixes https://github.com/suitenumerique/docs/issues/1329

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -941,37 +941,64 @@ class DocumentViewSet(
         in the payload.
         """
         # Get document while checking permissions
-        document = self.get_object()
+        document_to_duplicate = self.get_object()
 
         serializer = serializers.DocumentDuplicationSerializer(
             data=request.data, partial=True
         )
         serializer.is_valid(raise_exception=True)
         with_accesses = serializer.validated_data.get("with_accesses", False)
-        is_owner_or_admin = document.get_role(request.user) in models.PRIVILEGED_ROLES
+        user_role = document_to_duplicate.get_role(request.user)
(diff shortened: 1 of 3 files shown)


src/backend/core/api/viewsets.py
```diff
@@ -941,37 +941,68 @@ def duplicate(self, request, *args, **kwargs):
...
-        extracted_attachments = set(extract_attachments(document.content))
-        attachments = list(extracted_attachments & set(document.attachments))
-        duplicated_document = document.add_sibling(
+        extracted_attachments = set(extract_attachments(document_to_duplicate.content))
+        attachments = list(
+            extracted_attachments & set(document_to_duplicate.attachments)
+        )
+        if (
+            not document_to_duplicate.is_root()
+            and user_role == models.RoleChoices.READER
```
sampaccoud: You could use a comparison instead and check that it is inferior to EDITOR so that it will continue to work when we introduce the COMMENT role

src/backend/core/api/viewsets.py
```diff
@@ -941,37 +941,68 @@ def duplicate(self, request, *args, **kwargs):
...
+        ):
+            duplicated_document = models.Document.add_root(
+                creator=self.request.user,
+                title=capfirst(
+                    _("copy of {title}").format(title=document_to_duplicate.title)
```
sampaccoud: Maybe factorize the code that computes the title of the duplicated document with the code below

Issue #115, opened 2024-07-01
✨Export to a word document


AntoLC:
## Feature Request

As it is now, we can export to a PDF, by inserting our content inside a template. We want to be able to do the same but to a word document.

## Code resources:

https://github.com/numerique-gouv/impress/blob/main/src/frontend/apps/impress/src/features/docs/doc-tools/components/ModalPDF.tsx

## Proposal

- Use the same component, but rename it with some more general (ModalExport by exmaple).
- Add a select or a radio input to choose between export to PDF or Word.

## What we have

![image](https://github.com/numerique-gouv/impress/assets/25994652/fc6f9a3b-d13d-4473-80cb-18c98d9576e0)

PR #1388, opened 2025-09-16
✨(frontend) fix tree keyboard toggle when children not yet loaded

Ovgodd:
## Purpose

Fix tree node lazy-loading behavior when toggling via keyboard.
Previously, children loaded only if the node had already been expanded via mouse.

## Proposal

- [x] Ensure useLoadChildrenOnOpen triggers fetch on first keyboard toggle
- [x] Prevent multiple fetches with hasLoadedRef

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to
 
 - ♿(frontend) improve accessibility:
   - #1354
+  - ✨fix tree keyboard toggle when children not yet loaded #1388
 
 ### Fixed
 
--- a/src/frontend/apps/impress/src/features/docs/doc-tree/components/DocSubPageItem.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-tree/components/DocSubPageItem.tsx
@@ -20,6 +20,7 @@ import { useLeftPanelStore } from '@/features/left-panel';
 import { useResponsiveStore } from '@/stores';
 
 import { useKeyboardActivation } from '../hooks/useKeyboardActivation';
+import { useLoadChildrenOnOpen } from '../hooks/useLoadChildrenOnOpen';
 
 import SubPageIcon from './../assets/sub-page-logo.svg';
 import { DocTreeItemActions } from './DocTreeItemActions';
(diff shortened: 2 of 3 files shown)


AntoLC (commented):
I think it is better to do the fix in the ui-kit repository as it will benefit for all: https://github.com/suitenumerique/ui-kit/

The problem comes from `TreeViewItem` component and how `handleLoadChildren` is loaded.
https://github.com/suitenumerique/ui-kit/blob/b93531e51a2c74527cd0773cb01292cb9a07be3d/src/components/tree-view/TreeViewItem.tsx#L131-L139

The onClick on the arrow should not trigger `handleClick` but set the node to open:
```ts
onClick={(e) => {
   e.stopPropagation();
   node.toggle();
 }}
```

We should then have a useEffect listening `node.isOpen`, when isOpen is at true, we trigger `handleLoadChildren` if not already done. By doing so, it will fix this issue, because the `useEffect` will be trigger when toggling with the keyboard.

Something like that:
```ts
useEffect(() => {
    if (isLeaf || hasLoadedChildren || node.data.value.hasLoadedChildren) {
      return;
    }

    setIsLoading(true);
    context?.treeData.handleLoadChildren(node.data.value.id).then(() =>
    setIsLoading(false));
  }, [node.isOpen]);

Ovgodd:
> I think it is better to do the fix in the ui-kit repository as it will benefit for all: https://github.com/suitenumerique/ui-kit/
> 
> The problem comes from `TreeViewItem` component and how `handleLoadChildren` is loaded. https://github.com/suitenumerique/ui-kit/blob/b93531e51a2c74527cd0773cb01292cb9a07be3d/src/components/tree-view/TreeViewItem.tsx#L131-L139
> 
> The onClick on the arrow should not trigger `handleClick` but set the node to open:
> 
> ```ts
> onClick={(e) => {
>    e.stopPropagation();
>    node.toggle();
>  }}
> ```
> 
> We should then have a useEffect listening `node.isOpen`, when isOpen is at true, we trigger `handleLoadChildren` if not already done. By doing so, it will fix this issue, because the `useEffect` will be trigger when toggling with the keyboard.
> 
> Something like that:
> 
> ```ts
> useEffect(() => {
>     if (isLeaf || hasLoadedChildren || node.data.value.hasLoadedChildren) {
>       return;
>     }
> 
>     setIsLoading(true);
>     context?.treeData.handleLoadChildren(node.data.value.id).then(() =>
>     setIsLoading(false));
>   }, [node.isOpen]);
> ```
 
Ok I agree yes. Thanks for the detailed explanation !

Ovgodd:
I’m closing this PR because the fix has been handled directly in the UI Kit.
This way, all applications using the UI Kit will benefit from it: [#115](https://github.com/suitenumerique/ui-kit/pull/115)
.

PR #1396, opened 2025-09-17
✨(frontend) improve NVDA navigation in DocShareModal

Ovgodd:
## Purpose

Fix UX and accessibility violations in DocShareModal where QuickSearch auto-selected the first result on open, causing NVDA to immediately announce "John Doe impress@impress.world" despite focus being in the search input. 

This violated  UX principles (user consent, predictability) and WCAG

## Proposal
- [x] Remove auto-selection useEffect that triggered premature NVDA announcements
- [x] Implement user interaction tracking, only navigate after user action
- [x] Fix ARIA modal structure - remove nested role="dialog", use aria-labelledby
- [x] Add proper combobox ARIA relationships (aria-controls, aria-expanded, role="listbox")
- [x] Control expansion state - only "expanded" after user interaction

### DIFF
--- a/src/frontend/apps/e2e/__tests__/app-impress/doc-editor.spec.ts
+++ b/src/frontend/apps/e2e/__tests__/app-impress/doc-editor.spec.ts
@@ -757,15 +757,21 @@ test.describe('Doc Editor', () => {
     await expect(searchContainer.getByText(docChild2)).toBeVisible();
     await expect(searchContainer.getByText(randomDoc)).toBeHidden();
 
-    // use keydown to select the second result
+    await page.keyboard.press('ArrowDown');
     await page.keyboard.press('ArrowDown');
     await page.keyboard.press('Enter');
 
-    const interlink = page.getByRole('link', {
-      name: 'child-2',
+    // Wait for the search container to disappear, indicating selection was made
+    await expect(searchContainer).toBeHidden();
+
+    // Wait for the interlink to be created and rendered
+    const editor = page.locator('.ProseMirror.bn-editor');
+
+    const interlink = editor.getByRole('link', {
+      name: docChild2,
     });
 
-    await expect(interlink).toBeVisible();
+    await expect(interlink).toBeVisible({ timeout: 10000 });
     await interlink.click();
 
     await verifyDocName(page, docChild2);
--- a/src/frontend/apps/e2e/__tests__/app-impress/doc-member-create.spec.ts
+++ b/src/frontend/apps/e2e/__tests__/app-impress/doc-member-create.spec.ts
@@ -26,9 +26,8 @@ test.describe('Document create member', () => {
 
     await page.getByRole('button', { name: 'Share' }).click();
 
-    const inputSearch = page.getByRole('combobox', {
-      name: 'Quick search input',
-    });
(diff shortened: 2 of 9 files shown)


src/frontend/apps/e2e/__tests__/app-impress/doc-member-create.spec.ts
```diff
@@ -16,97 +16,80 @@ test.describe('Document create member', () => {
...
+      { email: 'user1@test.com', full_name: 'User One' },
+      { email: 'user2@test.com', full_name: 'User Two' },
+    ];
+
+    await page.route('**/users/?q=*', (route) =>
+      route.fulfill({
+        status: 200,
+        contentType: 'application/json',
+        body: JSON.stringify(users),
+      }),
```
AntoLC: We don't want to mock this test, this test assert not only the frontend but the backend as well.

src/frontend/apps/impress/src/components/quick-search/QuickSearchInput.tsx
```diff
@@ -16,6 +16,10 @@ type Props = {
   placeholder?: string;
   children?: ReactNode;
   withSeparator?: boolean;
+  listId?: string;
+  hasResults?: boolean;
```
AntoLC: I don't think you need `hasResults`, `isExpanded` should be enough. 
Then here: https://github.com/suitenumerique/docs/blob/28f34ce25c017c1fb6eaaa9dddd6cd0a260f3ee8/src/frontend/apps/impress/src/components/quick-search/QuickSearch.tsx#L86

You can just do: 
```ts
 isExpanded={isExpanded ?? hasResults}
```

src/frontend/apps/e2e/__tests__/app-impress/doc-editor.spec.ts
```diff
@@ -760,17 +760,22 @@ test.describe('Doc Editor', () => {
...
     await expect(searchContainer.getByText(docChild1)).toBeVisible();
     await expect(searchContainer.getByText(docChild2)).toBeVisible();
+    await expect(searchContainer.getByText(docChild2)).toBeVisible();
     await expect(searchContainer.getByText(randomDoc)).toBeHidden();
 
-    // use keydown to select the second result
-    await page.keyboard.press('ArrowDown');
-    await page.keyboard.press('Enter');
+    // Click directly on the second result instead of using keyboard navigation
+    await searchContainer.getByText(docChild2).click();
```
AntoLC: We use the keyboard by purpose, to assert that the navigation by keyboard is working.

src/frontend/apps/e2e/__tests__/app-impress/doc-member-create.spec.ts
AntoLC, line 1: I don't understand why we have to update this test, the development does not seem to touch really this part.
I can see it is still mocking the users request, it should not, it is not asserting anymore that the users are correctly displayed.

Ovgodd: Yes I agree, that was a mistake !

AntoLC (approved):

PR #1402, opened 2025-09-17
🐛(frontend) fix 404 page when reload 403 page

AntoLC:
## Purpose

When users were reloading a 403 page, they were redirected to the 404 page because of Nextjs routing mechanism. 
This PR fixes this issue by removing the 403 page from the pages directory and creating a component that is used directly in the layout when a 403 error is detected.

## Proposal

- [x] 🐛(frontend) fix 404 page when reload 403 page
- [x] 🥅(frontend) improve meta 401 page
- [x] 🛂(frontend) invalidate doc query when lost connection


Ovgodd (approved):
LGTM !

Issue #1390, opened 2025-09-16
Hide decorative icons, reduce noise from assistive technologies


cyberbaloo:
## Bug Report

**Problematic behavior**
Some icons are decorative and must be hidden from assistive technologies.

**Expected behavior/code**
Material icons are being rendered when they shouldn't be. (example: `<span class="sc-34780e96-0 sc-262648e5-0 iiwUsF foIyfo --docs--icon-bg material-icons" data-testid="public-icon">vpn_lock</span>`)

**Steps to Reproduce**
1. Search for `<span>` elements that contain "material-icons" in the CSS class name

**Possible Solution**

1. Add `aria-hidden="true"`
2. And the icon will no longer make noise in assistive technologies, the content will be much clearer because hearing "vpn_lock" when it's just the decorative icon of the document that is visible to everyone since the document is accessible to anyone who is logged in.

**Additional context/Screenshots**

<img width="522" height="62" alt="Image" src="https://github.com/user-attachments/assets/bbb124b5-1258-4d96-9196-54ce0f22ed34" />

PR #1404, opened 2025-09-18
✨(frontend) hide decorative icons from assistive tech with aria-hidden

Ovgodd:
## Purpose

Improve accessibility by hiding decorative icons from assistive technologies using aria-hidden. This prevents screen readers from announcing icons that are purely visual

This solves issue : [1390](https://github.com/suitenumerique/docs/issues/1390)

## Proposal

- [x] Add aria-hidden logic to <Icon /> based on presence of ARIA labels
- [x] Ensure decorative icons default to aria-hidden="true" when appropriate
- [x] Respect existing ARIA attributes when explicitly set by component props

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to
   - ♿ improve accessibility by adding landmark roles to layout #1394
   - ♿ add document visible in list and openable via enter key #1365
   - ♿ add pdf outline property to enable bookmarks display #1368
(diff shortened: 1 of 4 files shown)


Ovgodd:
> Looks good, but I think you will have some tests e2e to fix, the icon will not be accessible in some places.

Yep indeed !

Ovgodd:
> Looks good, but I think you will have some tests e2e to fix, the icon will not be accessible in some places.

All good, all e2e tests pass.

PR #1411, opened 2025-09-19
♻️(frontend) replace Arial font-family with token font

AntoLC:
## Purpose

In some components, the Arial font was still used because of a centering problem.

## Proposal

We removed all instances of Arial and replaced them with the current font token, the centering problems
were fixed by adding "contain: content;" to the css.

- [x] 🍱(frontend) replace Marianne font
- [x] ♻️(frontend) replace Arial font-family with token font


Ovgodd (approved):
Good !

PR #1441, opened 2025-10-01
✨(frontend) exclude h4-h6 headings from table of contents

Ovgodd:
## Purpose

Fix the document outline behavior where `h4 `to `h6 `headings were incorrectly considered as higher-level (similar to <h1>).
This caused the table of contents to display misleading hierarchy and confused navigation.

## Proposal
- [ ]  Exclude `h4` to `h6 `headings from the generated table of contents


src/frontend/apps/impress/src/features/docs/doc-table-content/components/TableContent.tsx
AntoLC, line 1: I don't think you should do the filter here, but here instead: https://github.com/suitenumerique/docs/blob/e842ae8d8a75d6e0cd94b4bcb6122a5318d8d8c5/src/frontend/apps/impress/src/features/docs/doc-editor/stores/useHeadingStore.tsx#L32

Ovgodd: Yes indeed, thank you

Issue #1410, opened 2025-09-19
🐛Can change emoji when reader role


AntoLC:
## Bug Report

When I am a reader, I can still locally change the Emoji in the emoji picker.

It is a minor issue as this change is visible only for the current user (not dispatch).

<img width="560" height="448" alt="Image" src="https://github.com/user-attachments/assets/cce418c8-6b84-4477-a3f9-29e1009512af" />

PR #1449, opened 2025-10-02
🔒(frontend) prevent readers from changing callout emoji

Ovgodd:
## Purpose

Ensure proper permissions handling in callouts by preventing users with reader role from editing the emoji.

https://github.com/user-attachments/assets/37f71fb8-ef3d-4cc7-b356-3c649aa9a0d3

issue : [1410](https://github.com/suitenumerique/docs/issues/1410)

## Proposal

- [x] Restrict callout emoji editing to users with sufficient permissions
- [x] Readers can still view callouts but cannot modify the emoji
- [x] Editors and above keep full editing rights
- [x] Remove UX feedback in reader mode to avoid users thinking they can take action.


AntoLC (approved):
Looks good ! 🎉

Issue #1445, opened 2025-10-01
Small UX quirk on column


virgile-dev:
## Bug Report

**Problematic behavior**
When you create 2 columns you get this kind of issues

<img width="1795" height="670" alt="Image" src="https://github.com/user-attachments/assets/d1c32b74-6d75-44b9-89d5-e94801423f90" />

<img width="1795" height="670" alt="Image" src="https://github.com/user-attachments/assets/cb47d46c-4e0b-4b1d-ac9f-e5417bef4fdd" />

**Expected behavior/code**
The place holder shouldn't overlap on the next column. User think their second column was not created.

**Steps to Reproduce**
1. Do this...
2. Then this...
3. And then the bug happens!

**Environment**
- Docs version:
- Instance url:

**Possible Solution**

**Additional context/Screenshots**
These issues are not present on blocknotejs.org

PR #1455, opened 2025-10-06
🐛(frontend) fix overlapping placeholders in multi-column layout

Ovgodd:
## Purpose

Fix visual issues with placeholders overlapping when multiple columns are created.
Ensure that text wraps correctly, avoiding confusion for users when adding a second or third column.

https://github.com/user-attachments/assets/523ad157-272a-4ed7-a326-2a41e9e18c3f

issue : [1445](https://github.com/suitenumerique/docs/issues/1445)

## Proposal
- [ ]  Prevent overlapping of placeholder content in multi-column layout
- [ ]  Apply correct word wrapping for text in 2–3 column configurations
- [ ]  Improve user perception of column creation success


src/frontend/apps/impress/src/features/docs/doc-editor/styles.tsx
```diff
@@ -16,6 +16,17 @@ export const cssEditor = (readonly: boolean) => css`
       font-weight: 400;
     }
 
+    .bn-block-content[data-is-empty-and-focused][data-content-type='paragraph']
+      .bn-inline-content:has(> .ProseMirror-trailingBreak:only-child)::before {
+      position: static !important; /* override absolute positioning from lib */
```
AntoLC: It is working ok but there is a few side effects by doing like that, this part become somehow clickable, making the cursor at the end.

In collaboration, we have a scrollbar appearing sometimes and some cases where it will move the editors line for other collaborator:

Image

https://github.com/user-attachments/assets/aab9c22a-bcd0-4cec-b839-0f0070395052


What about that instead, with ellipsis when reaching the end ? 
```css
.bn-block-content[data-is-empty-and-focused][data-content-type='paragraph']
      .bn-inline-content:has(> .ProseMirror-trailingBreak:only-child)::before {
      text-overflow: ellipsis;
      white-space: nowrap;
      overflow: hidden;
      width: inherit;
      height: inherit;
    }

    .bn-block-content[data-is-empty-and-focused][data-content-type='paragraph']
      .bn-inline-content:has(> .ProseMirror-trailingBreak:only-child) {
      position: relative;
    }
```

Image

Open to suggestion.

Ovgodd: Oh I didn't see those side effects, nice catch,

hm IMO your suggestion is good and far better.

AntoLC (approved):

Issue #1180, opened 2025-07-15
🐛 Panel bar adjustable


AntoLC:
## Feature Request

Past a certain deepness we cannot see the doc anymore. 
We need either to have a panel bar adjustable or a horizontal overflow.

<img width="451" height="516" alt="Image" src="https://github.com/user-attachments/assets/2ec24a14-d79c-496a-8cc6-82f41c82fea7" />

buildwithricky:
I'd be interested in contributing to this project

virgile-dev:
@buildwithricky you can check the label ‘good first issue’ for starters. 
Also have a look at the ‘CONTRIBUTING.md’ before contributing.

PR #1456, opened 2025-10-07
🐛(frontend) show full nested doc names with resizable panel

Ovgodd:
## Purpose

EDIT & FINAL: Following a discussion with @rl-83, we replaced the horizontal scroll approach with an adjustable panel bar using `react-resizable-panels`.

This change improves usability by allowing users to resize the document tree panel within defined limits. It ensures better visibility of nested documents without using scrollbars and aligns with the implementation used in our shared UI Kit (where resizing is handled via the `MainLayout`, not inside the document tree component).

added e2e tests to cover this new resizable behavior too.

issue : [1180](https://github.com/suitenumerique/docs/issues/1180)

https://github.com/user-attachments/assets/ec0ee816-7e70-4071-b616-b8031f7614f2

## Proposal

- [x] Add `react-resizable-panels` (v3.0.6) dependency
- [x] Implement `PanelGroup` with a horizontal resize handle in `MainLayout`
- [x] Persist panel size in `localStorage`
- [x] Set min width to 300px, max width to 450px (capped at 40% of viewport)
- [x] Keep mobile behavior unchanged (no resize handle)
- [x] Add e2e tests for resizable panel behavior

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to
 ## Fixed
 
 - 🐛(frontend) fix duplicate document entries in grid #1479
+- 🐛(frontend) show full nested doc names with ajustable bar #1456
 
 ## [3.8.2] - 2025-10-17
 
@@ -30,13 +31,16 @@ and this project adheres to
 
 - 🔥(backend) remove treebeard form for the document admin #1470
 
-
 ## [3.8.0] - 2025-10-14
 
 ### Added
 
 - ✨(frontend) add pdf block to the editor #1293
 - ✨List and restore deleted docs #1450
+
+### Fixed
+
+- 🐛(frontend) show full nested doc names with ajustable bar #1456
 
 ### Changed
 
--- a/src/frontend/apps/impress/package.json
+++ b/src/frontend/apps/impress/package.json
@@ -60,6 +60,7 @@
     "react-dom": "*",
     "react-i18next": "15.7.3",
     "react-intersection-observer": "9.16.0",
+    "react-resizable-panels": "3.0.6",
     "react-select": "5.10.2",
     "styled-components": "6.1.19",
     "use-debounce": "10.0.6",
--- a/src/frontend/apps/impress/src/features/docs/doc-tree/components/DocSubPageItem.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-tree/components/DocSubPageItem.tsx
@@ -163,6 +163,7 @@ export const DocSubPageItem = (props: TreeViewNodeProps<Doc>) => {
           aria-label={`${t('Open document {{title}}', { title: docTitle })}`}
           $css={css`
             text-align: left;
+            min-width: 0;
           `}
         >
           <Box $width="16px" $height="16px">
@@ -180,8 +181,10 @@ export const DocSubPageItem = (props: TreeViewNodeProps<Doc>) => {
               display: flex;
               flex-direction: row;
               width: 100%;
+              min-width: 0;
               gap: 0.5rem;
               align-items: center;
+              overflow: hidden;
             `}
           >
             <Text $css={ItemTextCss} $size="sm" $variation="1000">
--- a/src/frontend/apps/impress/src/features/docs/doc-tree/components/DocTree.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-tree/components/DocTree.tsx
@@ -184,7 +184,6 @@ export const DocTree = ({ currentDoc }: DocTreeProps) => {
         /* Remove outline from TreeViewItem wrapper elements */
         .c__tree-view--row {
           outline: none !important;
-
           &:focus-visible {
             outline: none !important;
           }
@@ -241,7 +240,7 @@ export const DocTree = ({ currentDoc }: DocTreeProps) => {
               }
             }
             &:hover,
-            &:focus-within {
+            &:focus-visible {
               .doc-tree-root-item-actions {
                 opacity: 1;
               }
--- a/src/frontend/apps/impress/src/layouts/MainLayout.tsx
+++ b/src/frontend/apps/impress/src/layouts/MainLayout.tsx
@@ -6,23 +6,21 @@ import { Box } from '@/components';
 import { useCunninghamTheme } from '@/cunningham';
 import { Header } from '@/features/header';
 import { HEADER_HEIGHT } from '@/features/header/conf';
-import { LeftPanel } from '@/features/left-panel';
-import { MAIN_LAYOUT_ID } from '@/layouts/conf';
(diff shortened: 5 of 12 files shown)


src/frontend/yarn.lock
AntoLC, line 1: The `yarn.lock` gives changes in local.

src/frontend/apps/impress/src/features/docs/doc-tree/components/DocSubPageItem.tsx
```diff
@@ -116,9 +116,11 @@ export const DocSubPageItem = (props: TreeViewNodeProps<Doc>) => {
           ? 'var(--c--theme--colors--greyscale-100)'
           : 'var(--c--theme--colors--greyscale-000)'};
         .light-doc-item-actions {
-          display: ${menuOpen || !isDesktop ? 'flex' : 'none'};
-          position: absolute;
+          display: flex;
+          position: sticky;
           right: 0;
+
+          opacity: ${menuOpen || !isDesktop ? '1' : '0'};
```
AntoLC: By doing like that the name is cropped far before needed, because the actions button take place even when not displayed.

src/frontend/apps/impress/src/features/docs/doc-tree/components/DocTree.tsx
AntoLC, line 1: Not introduce in this PR but I can see there is differences on how the actions buttons are displayed between the top parent and the children, I think it should be `&:focus-visible` here.
https://github.com/suitenumerique/docs/blob/f0590148dbf5392471a8ea847bfc2eb11f57e922/src/frontend/apps/impress/src/features/docs/doc-tree/components/DocTree.tsx#L243-L246

src/frontend/apps/impress/src/layouts/MainLayout.tsx
```diff
@@ -12,50 +18,186 @@ import { useResponsiveStore } from '@/stores';
...
+                $css={css`
+                  overflow-y: auto;
+                  overflow-x: clip;
+                `}
+              >
+                {children}
+              </Box>
+            </Panel>
+          </PanelGroup>
+        ) : (
```
AntoLC: It should have its own component I think.

src/frontend/apps/impress/src/layouts/MainLayout.tsx
```diff
@@ -12,50 +18,186 @@ import { useResponsiveStore } from '@/stores';
...
+                    : colorsTokens['greyscale-050']
+                }
+                $css={css`
+                  overflow-y: auto;
+                  overflow-x: clip;
+                `}
+              >
+                {children}
+              </Box>
+            </Panel>
```
AntoLC: I can see the code almost the same for the mobile part under, can we factorize this part.

src/frontend/apps/impress/src/layouts/MainLayout.tsx
```diff
@@ -12,50 +18,186 @@ import { useResponsiveStore } from '@/stores';
...
+              ref={ref}
+              order={0}
+              defaultSize={minPanelSize}
+              minSize={minPanelSize}
+              maxSize={maxPanelSize}
+            >
+              <LeftPanel />
+            </Panel>
+            <PanelResizeHandle
+              className="border-clr-surface-primary"
```
AntoLC: Not sure we use it.

src/frontend/apps/impress/src/layouts/MainLayout.tsx
```diff
@@ -12,50 +18,186 @@ import { useResponsiveStore } from '@/stores';
...
+      enableResize ? setMaxPanelSize(max) : setMaxPanelSize(min);
+    };
+
+    updatePanelSize();
+    window.addEventListener('resize', updatePanelSize);
+
+    return () => {
+      window.removeEventListener('resize', updatePanelSize);
+    };
+  }, [isDesktop, enableResize]);
```
AntoLC: This part can be improve I think, we use twice the same listener (`resize`), we could use only one.
If `isDesktop` is false, I have the feeling we don't need to listen anything.

When using `useEffect`, it is good practice to use `useCallback` around the function called inside, `calculateDefaultSize` should be inside a `useCallback`.

I think by having its own component we will be able to remove this notion of `isDesktop`, this component will be called only when desktop, removing complexity.

src/frontend/apps/impress/src/layouts/components/ResizableLeftPanel.tsx
```diff
@@ -0,0 +1,94 @@
...
+  return (
+    <PanelGroup autoSaveId="docs-left-panel-persistence" direction="horizontal">
+      <Panel
+        ref={ref}
+        order={0}
+        defaultSize={minPanelSize}
+        minSize={minPanelSize}
+        maxSize={maxPanelSize}
+      >
+        <LeftPanel />
```
AntoLC: Could it be a parameter as well, to be fully generic ?

src/frontend/apps/impress/package.json
```diff
@@ -60,6 +60,7 @@
     "react-dom": "*",
     "react-i18next": "15.7.3",
     "react-intersection-observer": "9.16.0",
+    "react-resizable-panels": "^3.0.6",
```
AntoLC:
```suggestion
    "react-resizable-panels": "3.0.6",
```

src/frontend/yarn.lock
AntoLC, line 1: You should not remove the `yarn.lock`, you should put back the previous one, then do `yarn`, it will only include the new dependency `react-resizable-panels` and what is needed with it.

src/frontend/apps/impress/src/layouts/MainLayout.tsx
```diff
@@ -10,52 +10,108 @@ import { LeftPanel } from '@/features/left-panel';
...
-              : colorsTokens['greyscale-050']
-          }
-          $css={css`
-            overflow-y: auto;
-            overflow-x: clip;
-          `}
-        >
-          {children}
-        </Box>
+        {renderContent()}
```
AntoLC: It is a bit anti-pattern like that, what about its own component MainLayoutContent ?

```suggestion
        <MainLayoutContent  enableResizablePanel={enableResizablePanel}>{children}</MainLayoutContent>
```

CHANGELOG.md
```diff
@@ -51,6 +51,7 @@ and this project adheres to
 - 🐛(frontend) reduce no access image size from 450 to 300 #1463
 - 🐛(frontend) preserve interlink style on drag-and-drop in editor #1460
 - ✨(frontend) load docs logo from public folder via url #1462
+- 🐛(frontend) show full nested doc names with ajustable bar #1456
```
AntoLC: Under the `Unreleased` part.

PR #1457, opened 2025-10-07
🐛(backend) filter invitation with case insensitive email

lunika:
## Purpose

A user can be invited and no control is made on the email case. Then,
when a new user is created, we are looking if there are pending
invitation and the lookup used is case sensitive. We change it using
__iexact which is case insensitive.

## Proposal

- [x] 🐛(backend) filter invitation with case insensitive email
- [x]  ♻️(backend) force saving invitation email in lowercase

### DIFF
--- a/src/frontend/apps/e2e/__tests__/app-impress/doc-member-create.spec.ts
+++ b/src/frontend/apps/e2e/__tests__/app-impress/doc-member-create.spec.ts
@@ -169,8 +169,15 @@ test.describe('Document create member', () => {
 
     const inputSearch = page.getByTestId('quick-search-input');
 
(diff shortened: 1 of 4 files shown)


src/backend/core/tests/documents/test_api_document_invitations.py
```diff
@@ -596,6 +596,32 @@ def test_api_document_invitations_create_cannot_invite_existing_users():
     }
 
 
+def test_api_document_invitations_create_lower_email():
+    """
+    INo matter the case, the email should be converted to lowercase.
```
AntoLC:
```suggestion
    No matter the case, the email should be converted to lowercase.
```

PR #1460, opened 2025-10-07
✨(frontend) preserve interlink style on drag-and-drop in editor

Ovgodd:
## Purpose

Fix DnD of links
Before this fix: both drag-and-drop and copy/paste turned them into plain <a> links, even when trying the “extend selection with left arrow” trick.

For copy-paste, the behavior is odd: I have to extend the selection to include the whole chip (e.g., Shift+Left / selecting just before the text); if I only select the text, it copies plain text and not the interlink.

https://github.com/user-attachments/assets/d4574542-0243-47f3-b568-3dfb7844917d

## Proposal

- [ ]  Add useInterlinkDropNormalizer hook.
- [ ]  On editor changes, handle sources: drop, paste, insert
- [ ]  Normalize affected blocks: convert <a href="/docs/:id">…</a> back to interlinkingLinkInline (url, docId, title).


Ovgodd:
> I found the root problem actually, here we use a `StyledLink` that render a `a` tag:
> 
> https://github.com/suitenumerique/docs/blob/2db509d4fa4c72dd91749c97671661fd0fff17b4/src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-inline-content/Interlinking/InterlinkingLinkInlineContent.tsx#L56
> 
> There is a conflict with the `a` tag, I think Blocknote or Tiptap take over our custom component during pasting or copy event.
> 
> An easy fix is to use a `button` tag instead of a `a` tag, like that we avoid the conflict, ex:
> 
> ```tsx
> const LinkSelected = ({ url, title }: LinkSelectedProps) => {
>   const { colorsTokens } = useCunninghamTheme();
>   const router = useRouter();
> 
>   return (
>     <BoxButton
>       onClick={(e) => {
>         e.preventDefault();
>         router.push(url);
>       }}
>       draggable="false"
>       $css={css`
>         display: inline;
>         padding: 0.1rem 0.4rem;
>         border-radius: 4px;
>         & svg {
>           position: relative;
>           top: 2px;
>           margin-right: 0.2rem;
>         }
>         &:hover {
>           background-color: ${colorsTokens['greyscale-100']};
>         }
>         transition: background-color 0.2s ease-in-out;
>       `}
>     >
>       <SelectedPageIcon width={11.5} />
>       <Text $weight="500" spellCheck="false" $size="16px" $display="inline">
>         {title}
>       </Text>
>     </BoxButton>
>   );
> };
> ```

Oh yes nice, and with that we can get rid of my useInterlinkDropNormalizer hook :o ! this is simpler ahha and it works

Issue #1221, opened 2025-07-24
Emoji dropdown is hidden by left sidebar


virgile-dev:
## Bug Report

**Before you file your issue**
- Check the other [issues](https://github.com/suitenumerique/docs/issues) before filing your own
- If your report is related to the ([BlockNote](https://github.com/TypeCellOS/BlockNote)) text editor, [file it on their repo](https://github.com/TypeCellOS/BlockNote/issues). If you're not sure whether your issue is with BlockNote or Docs, file it on our repo: if we support it, we'll backport it upstream ourselves 😊, otherwise we'll ask you to do so.

**Problematic behavior**

Emoji dropdown is hidden by left sidebar on some resolutions

**Expected behavior/code**
the emoji menu should display above

**Steps to Reproduce**
1. Do /emoji on a big rez 
2. Then see this

<img width="2145" height="1252" alt="Image" src="https://github.com/user-attachments/assets/d4f3fb0f-8da6-497a-b5a9-4ed3af1b466f" />

**Environment**
- Docs version: 3.4.2
- Instance url: docs.numerique.gouv.fr

**Possible Solution**

**Additional context/Screenshots**
Add any other context about the problem here. If applicable, add screenshots to help explain.

RoshanMuhammedR:
Can I work on this ?

virgile-dev:
@RoshanMuhammedR sure! I’ll try to reproduce first.

virgile-dev:
I can't reproduce anymore. Closing

PR #1466, opened 2025-10-08
🐛(frontend) force emoji picker to fixed position to stay visible on zoom

Ovgodd:
## Purpose

Improve the visibility of the emoji picker in the BlockNote editor by forcing its position to fixed. This prevents it from being visually hidden under the left sidebar, especially when zooming in or out.

issue : [1221](https://github.com/suitenumerique/docs/issues/1221)

In the screenshot, I intentionally forced the tranform / translate to the left, to demonstrate the behavior.

Capture d'écran 2025-10-08 130717

## Proposal

- [x] Add global CSS rule via createGlobalStyle to target emoji suggestion container 
- [x] Force position: fixed on emoji picker using a data attribute selector
- [x] Prevent clipping behind sidebar or other layout containers
- [x]  Verified via manual testing and visual confirmation on zoomed layout

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,8 @@ and this project adheres to
 - 🐛(frontend) fix overlapping placeholders in multi-column layout #1455
 - 🐛(backend) filter invitation with case insensitive email
 - 🐛(frontend) reduce no access image size from 450 to 300 #1463
+- 🐛(frontend) force emoji picker to fixed position to stay visible on zoom #1466
+
 
 ## [3.7.0] - 2025-09-12
 
--- a/src/frontend/apps/impress/src/features/docs/doc-editor/components/BlockNoteEditor.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-editor/components/BlockNoteEditor.tsx
@@ -13,6 +13,7 @@ import { useCreateBlockNote } from '@blocknote/react';
 import { HocuspocusProvider } from '@hocuspocus/provider';
 import { useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
+import { createGlobalStyle } from 'styled-components';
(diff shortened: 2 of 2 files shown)


AntoLC (commented):
Are we sure there is a bug ?
I don't reproduce it on the production environment, neither with the zoom:

https://github.com/user-attachments/assets/b11003e0-a566-431d-a0d2-54c540238fb7

Ovgodd:
> Are we sure there is a bug ? I don't reproduce it on the production environment, neither with the zoom:
> 
>  -.Docs.3.webm

hm indeed, On my side I wasn't able to reproduce the problem too, the "only way" to simulate the virgil's issue was to modify manually the transform to watch what happened IF the bug may come back,

wdyt ?

AntoLC:
> hm indeed, On my side I wasn't able to reproduce the problem too, the "only way" to simulate the virgil's issue was to modify manually the transform to watch what happened IF the bug may come back,
> 
> wdyt ?

If you don't reproduce the bug, you don't know if your fix works.
We should try to see with Virgile how he gets this bug to be able to reproduce it, if Virgile cannot reproduce the bug, we should close the issue, and the PR.

Ovgodd:
> > hm indeed, On my side I wasn't able to reproduce the problem too, the "only way" to simulate the virgil's issue was to modify manually the transform to watch what happened IF the bug may come back,
> > wdyt ?
> 
> If you don't reproduce the bug, you don't know if your fix works. We should try to see with Virgile how he gets this bug to be able to reproduce it, if Virgile cannot reproduce the bug, we should close the issue, and the PR.

Yep that makes sense

Ovgodd:
https://github.com/suitenumerique/docs/issues/1221,
Closing this PR, the bug could not be reproduced.

PR #1467, opened 2025-10-08
🔧(action) publish a comment when label preview is added

lunika:
## Purpose

The label preview will deploy a full environment. This environment is accessible using a specific url. This commit will publish a comment with the good url.

## Proposal

- [x] 🔧(action) publish a comment when label preview is added

### DIFF
--- a/.github/workflows/label_preview.yml
+++ b/.github/workflows/label_preview.yml
@@ -1,0 +1,27 @@
+name: Label Preview
+
+on:
+  pull_request:
+    types: [labeled, opened]
+
+permissions: 
+  pull-requests: write
+
+jobs:
+  comment:
+    runs-on: ubuntu-latest
+    if: contains(github.event.pull_request.labels.*.name, 'preview')
+    steps:
+      - uses: thollander/actions-comment-pull-request@v3
+        with:
+          message: |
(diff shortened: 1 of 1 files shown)


.github/workflows/label_preview.yml
```diff
@@ -0,0 +1,27 @@
...
+    if: contains(github.event.pull_request.labels.*.name, 'preview')
+    steps:
+      - uses: thollander/actions-comment-pull-request@v3
+        with:
+          message: |
+            :rocket: Preview will be available at [https://${{ github.event.pull_request.number }}-docs.ppr-docs.beta.numerique.gouv.fr/](https://${{ github.event.pull_request.number }}-docs.ppr-docs.beta.numerique.gouv.fr/)
+
+            You can use the existing account with these credentials:
+            - username: `docs`
+            - password: `docs`
```
AntoLC, line 22: Apparently it is the way to avoid `gitgardian` to give an alert: https://blog.gitguardian.com/setting-up-a-pre-commit-git-hook-with-gitguardian-shield-to-scan-for-secrets/

```suggestion
            - password: `docs`  # ggignore
```

.github/workflows/label_preview.yml
```diff
@@ -0,0 +1,27 @@
...
+
+            You can also create a new account if you want to.
```
AntoLC, line 24: I didn't succeed to create an account.

lunika: I tried, and it worked for me. Weird

.github/workflows/label_preview.yml
lunika: Yes but I will see it in the comment then.

PR #1496, opened 2025-10-16
🐛(backend) fix s3 version_id validation

lennartzellmer:
# Loosen URL `version_id` Path Validation to Match S3 Specification

## Purpose

The current validation for `version_id` is too strict. It assumes that a version ID consists only of lowercase characters, numbers, and hyphens. This holds true for MINIO, which uses UUIDs as version IDs. However, according to the S3 documentation, version IDs can be more opaque strings:

> Version IDs are Unicode, UTF-8 encoded, URL-ready, opaque strings that are no more than 1,024 bytes long. The following is an example: 3sL4kqtJlcpXroDTDmJ+rmSpXd3dIbrHY+MTRCxf3vjVBH40Nr8X8gdRQBpUMLUo
>
-> More details: https://docs.aws.amazon.com/AmazonS3/latest/userguide/versioning-workflows.html#version-ids

This discrepancy causes issues with Hetzner's S3-compatible storage, which uses opaque strings as version IDs. These IDs can contain uppercase and lowercase letters as well as special characters.

When running tests against a Hetzner S3 bucket with versioning enabled, the current validation fails:

image

Here’s an example of a Hetzner `version_id` that doesn’t conform to the UUID format:

```json
{
  ...
  "versions": [
    {
      "etag": "\"d153d2e52bbf6f443d6eb47439079b2f\"",
      "is_latest": false,
      "last_modified": "2025-10-16T14:55:33.471000Z",
      "version_id": "jH7VgK5xKEY-yWGxRpcDSRw30.p0r0m"
    },
    ...
  ]
}
```

### Solution

Loosen the version_id validation logic to accept any opaque string up to 1,024 characters, aligning with the official S3 specification.

### Why Tests Are Not Adjusted

The generation of version_id values is handled by the S3 implementation and is outside the scope of the test suite. The underlying issue is the assumption that all S3-compatible systems use UUID-like values. Since this is not a guaranteed or testable property across providers, no test adjustments are necessary.

All existing tests were executed successfully against both Hetzner S3 and MINIO after applying the fix:
SCR-20251016-tsor

---

## External contribution

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ and this project adheres to
 - 🐛(frontend) fix duplicate document entries in grid #1479
 - 🐛(frontend) show full nested doc names with ajustable bar #1456
 - 🐛(backend) fix trashbin list
+- ♿(frontend) improve accessibility:
+  - ♿(frontend) remove empty alt on logo due to Axe a11y error #1516
 
 ## [3.8.2] - 2025-10-17
 
--- a/src/frontend/apps/e2e/__tests__/app-impress/header.spec.ts
+++ b/src/frontend/apps/e2e/__tests__/app-impress/header.spec.ts
@@ -159,7 +159,7 @@ test.describe('Header: Override configuration', () => {
           logo: {
             src: '/assets/logo-gouv.svg',
             width: '220px',
-            alt: 'Gouvernement Logo',
+            alt: '',
           },
         },
       },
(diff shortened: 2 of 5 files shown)


lennartzellmer:
The matter has gained additional significance in light of MinIO’s recent decision to withdraw their official Docker distribution, showing a downshift in their commitment to open source and reliability... :( 
https://github.com/minio/minio/issues/21647

src/backend/core/api/viewsets.py
```diff
@@ -1102,7 +1102,7 @@ def versions_list(self, request, *args, **kwargs):
     @drf.decorators.action(
         detail=True,
         methods=["get", "delete"],
-        url_path="versions/(?P<version_id>[0-9a-z-]+)",
+        url_path=r"versions/(?P<version_id>[^/]{1,1024})",
```
```suggestion
        url_path=r"versions/(?P<version_id>[A-Za-z0-9._+\-=~]{1,1024})",
```

I would suggest something less permissive. Your proposal afraid me a bit and I don't want to put a regex so permissive in production. What I suggest is less permissive and I think enough to match your requirement.

WDYT ?

lennartzellmer: Thank you for reviewing the matter.
I completely understand your point regarding my regex - it may be a bit too broad. 
I tested your regex against a bigger set of Hetzner S3 (ceph) version IDs and there where no issues. 🙌
This would solve the issue! 🥳

lunika:
Sorry, I wanted to rebase your pull request, but as you are working on your main branch, pushing on it closed the pull request...

lunika:
Merged from this PR https://github.com/suitenumerique/docs/pull/1543

lunika:
Thank you for your contribution

lennartzellmer:
Awesome! Thank you!

PR #1509, opened 2025-10-21
🚸(frontend) separate viewers from editors

AntoLC:
## Purpose

We are now totally separating the viewers with the editors. 
We will not load the provider when we are in viewer mode, meaning the viewers will not be aware of other users and will not show their cursors anymore. 
We still get the document updates in real-time.


Ovgodd (approved):
Seems good to me !

PR #1516, opened 2025-10-23
✨(frontend) remove empty alt on logo due to Axe a11y error

Ovgodd:
## Purpose

Fix an accessibility issue raised by Axe: the logo image had an alt attribute, which was redundant with the surrounding link’s aria-label="Back to homepage". 

## Proposal

- [x] Set alt="" on the logo image to avoid redundancy
- [x] Ensure a11y remains covered by the parent link’s aria-label

### DIFF
--- a/src/frontend/apps/e2e/__tests__/app-impress/header.spec.ts
+++ b/src/frontend/apps/e2e/__tests__/app-impress/header.spec.ts
@@ -159,7 +159,7 @@ test.describe('Header: Override configuration', () => {
           logo: {
             src: '/assets/logo-gouv.svg',
             width: '220px',
-            alt: 'Gouvernement Logo',
+            alt: '',
           },
         },
       },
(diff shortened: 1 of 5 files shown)


src/frontend/apps/e2e/__tests__/app-impress/header.spec.ts
```diff
@@ -159,7 +159,7 @@ test.describe('Header: Override configuration', () => {
           logo: {
             src: '/assets/logo-gouv.svg',
             width: '220px',
-            alt: 'Gouvernement Logo',
```
AntoLC, line 162: Ok, so it means that we cannot override the `alt` anymore, we should modify the types so and remove the `alt` prop I suppose:
https://github.com/suitenumerique/docs/blob/246e0a930108ad54f92ca8e67880056fbbe17d1b/src/frontend/apps/impress/src/features/header/types.ts#L6

Same here, we should remove the `alt` part: 
https://github.com/suitenumerique/docs/blob/246e0a930108ad54f92ca8e67880056fbbe17d1b/src/helm/env.d/dev/configuration/theme/demo.json#L141

PR #1518, opened 2025-10-23
Fix/a11y axe critical errors

Ovgodd:
## Purpose

Improve accessibility (a11y) by aligning the document grid and editor with ARIA specifications and Axe recommandations.

## Proposal

- [x]  Move the "More documents" button outside the role="grid" container
- [x]  Ensure valid ARIA structure: grid > rowgroup > row
- [x]  Remove role="presentation" from Draggable and Droppable components
- [x]  Add aria-label to the BlockNoteView editor for better screen reader support


Ovgodd:
wrong branch

Issue #1222, opened 2025-07-24
Remove the templates dropdown in exports modal


virgile-dev:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
We never really worked on a real template. It was just to show it was possible.

**Describe the solution you'd like**
Remove the option all together meaning get rid of the template dropdown.

**Describe alternatives you've considered**
People can make the document however they like by exporting to docx and soon odt.

**Discovery, Documentation, Adoption, Migration Strategy**
Eport templates will be reintroduced when we do this feature : https://github.com/suitenumerique/docs/issues/680

**Do you want to work on it through a Pull Request?**

Issue #680, opened 2025-03-04
Custom header and footer for printing (Export Update)


virgile-dev:
@rl-83 could you add the specs and design that we discussed on this issue ?

virgile-dev:
Robin added a justify option to the feature https://github.com/suitenumerique/docs/issues/395

virgile-dev:
Closed by mistake

PR #1523, opened 2025-10-27
📝(documentation) Add custom export templates section with usage instructions and limitations

rvveber:
## Purpose

Documentation about how adding custom templates was lacking.


rvveber:
> Indeed it is a lacking documentation. I think we should wait a bit to merge this PR because we want to refactorise this template system, we have these 2 issues about it:
> 
> * #1222
> * #680
> 
> I would say that the best way to do template for the moment is like that:
> 
>  Feature.multi.columns.-.Docs.online-video-cutter.com.mp4

Hello, sure.
The video is already half of the documentation.
Whats left is just copy as HTML and paste into backend template.

AntoLC:
The templates feature has been removed (https://github.com/suitenumerique/docs/pull/1780), I think we can close this one.

PR #1547, opened 2025-10-31
🐛(frontend) fix doc timestamp display

buildwithricky:
## Purpose

Implemented the logic to show 'Just now' instead of '0 seconds ago' when the difference is under one second. 
Users can now see 'just now' when an edit happens.

## Proposal
- [x] Fixed the '0 seconds ago issue' to show 'Just Now' when edits happen.

Screenshot 2025-10-31 at 21 40 22

## External contributions

### DIFF
--- a/src/frontend/apps/impress/src/hooks/useDate.tsx
+++ b/src/frontend/apps/impress/src/hooks/useDate.tsx
@@ -10,7 +10,7 @@ const formatDefault: DateTimeFormatOptions = {
 };
 
 export const useDate = () => {
-  const { i18n } = useTranslation();
+  const { i18n, t } = useTranslation();
 
   const formatDate = (
     date: string,
@@ -22,7 +22,19 @@ export const useDate = () => {
   };
 
   const relativeDate = (date: string): string => {
-    return DateTime.fromISO(date).setLocale(i18n.language).toRelative() || '';
+    const dateToCompare = DateTime.fromISO(date);
+
+    if (!dateToCompare.isValid) {
+      return '';
+    }
+
+    const dateNow = DateTime.now();
+
+    const differenceInSeconds = dateNow.diff(dateToCompare).as('seconds');
+
+    return Math.abs(differenceInSeconds) >= 5
+      ? dateToCompare.toRelative({ base: dateNow, locale: i18n.language })
+      : t('just now');
   };
 
   const calculateDaysLeft = (date: string, daysLimit: number): number =>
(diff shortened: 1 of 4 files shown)


buildwithricky:
@rvveber

buildwithricky:
@rvveber where you able to review this yet the workflow hasn't started

src/frontend/apps/impress/src/hook/useDate.tsx
```diff
@@ -22,7 +22,19 @@ export const useDate = () => {
...
+      const diffInSeconds = Math.abs(dateTime.diffNow('seconds').seconds);
+      if (diffInSeconds < 1) {
+        return t('just now');
+      }
+      if (relative.includes('0 seconds') || relative.includes('0 second') ||
+        relative.match(/0\s*(second|seconde|segundo|秒)/i)) {
+        return t('just now');
+      }
+    }
+    return relative;
```
rvveber: Thanks for your contribution!

- Regex is unnecessary, we have input data so we should use it to calculate the difference
- Locale is only required for displaying and so it would be more readable to set it when displaying
- Always best to name variables explicitly and concisely
- Lets use a timeframe of 5 seconds to display just now for everything below it

Suggestion
```
  const dateToCompare = DateTime.fromISO(date);
  if (!dateToCompare.isValid) return '';
  const dateNow = DateTime.now();
  const differenceInSeconds = dateNow.diff(dateToCompare).as('seconds');

  return Math.abs(differenceInSeconds) >= 5
    ? dateToCompare.toRelative({ base: dateNow, locale: i18n.language })
    : t('just now');
```

Agree / Disagree?

rvveber (changes requested):

src/frontend/apps/impress/src/hook/useDate.tsx
buildwithricky: Agree , thanks for the feedback

buildwithricky:
@rvveber

virgile-dev:
@rl-83 FYI

rvveber:
@buildwithricky 
- Your commits need to be according to convention 
- Your commits need to have verified signatures.
- You need to add an entry into the changelog (e,g, `short explaination #1547`)
These points prevent me from beeing able to merge, could you fix that?

Btw. you don't need to wait for the Workflows on this Repository.
If you create a PR within your Fork
`buildwithricky:main <- buildwithricky:fix/705-doc-timestamp-display`
The workflow should run on your fork and you can check what the CI linter complains about - directly after you force push.

src/frontend/apps/impress/src/hooks/useDate.tsx
AntoLC, line 1: Could be nice to add a test, this file has some tests already made about it:
https://github.com/suitenumerique/docs/blob/main/src/frontend/apps/impress/src/features/docs/docs-grid/components/__tests__/DocsGridItemDate.test.tsx

Or here: https://github.com/suitenumerique/docs/blob/main/src/frontend/apps/impress/src/features/docs/doc-header/__tests__/DocHeaderInfo.spec.tsx

PR #1580, opened 2025-11-12
Mitigate role escalation in the ask_for_access viewset

lunika:
## Purpose

Via the ask_for_access viewset, it is possible for an administrator to escaladate an other user to an owner role and then take the control of the document by removing all other owner.

This PR ensure that an administrator can not accept role higher than the one he/she already have. Also, the serializer does not accept the owner role anymore.

## Proposal

- [x] 🔒️(backend) role in ask_for_access must be lower than user role
- [x] 🔒️(backend) remove owner as valid role for ask_for_access serializer

### DIFF
--- a/src/frontend/apps/e2e/__tests__/app-impress/doc-member-list.spec.ts
+++ b/src/frontend/apps/e2e/__tests__/app-impress/doc-member-list.spec.ts
@@ -160,6 +160,9 @@ test.describe('Document list members', () => {
       `You are the sole owner of this group, make another member the group owner before you can change your own role or be removed from your document.`,
     );
     await expect(soloOwner).toBeVisible();
+    await expect(
+      page.getByRole('menuitem', { name: 'Administrator' }),
+    ).toBeDisabled();
 
     await list.click({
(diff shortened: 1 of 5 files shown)


src/backend/core/api/serializers.py
```diff
@@ -810,11 +812,11 @@ class Meta:
         ]
         read_only_fields = ["id", "document", "user", "role", "created_at", "abilities"]
 
-    def get_abilities(self, invitation) -> dict:
+    def get_abilities(self, instance) -> dict:
         """Return abilities of the logged-in user on the instance."""
         request = self.context.get("request")
         if request:
-            return invitation.get_abilities(request.user)
+            return instance.get_abilities(request.user)
```
AntoLC, line 819: Could we modify the model to get `set_role_to` ability ? It is to know what are the roles available frontend side.
https://github.com/suitenumerique/docs/blob/b37f2aa555a93fcd584338d1a5f7cb4b23d1a501/src/backend/core/models.py#L1171

lunika: Done

Issue #1578, opened 2025-11-11
Table of contents/Summary fixed on top right corner of screen


Gu-Pa:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
When I scroll down a document, I no longer have access to the table of contents.

**Describe the solution you'd like**
I would like the table of contents in Docs to be fixed, meaning that it always appears at the top right of the screen and not at the top right of the document, so that when you scroll down, the table of contents is always visible at the top right of the screen.

**Describe alternatives you've considered**
None

<img width="2398" height="1288" alt="Image" src="https://github.com/user-attachments/assets/7d1a6151-9421-4dfe-aa2c-651c2def3f0a" />
<img width="2398" height="1288" alt="Image" src="https://github.com/user-attachments/assets/d4002ef9-fa4b-4e7b-b584-7d43657b198e" />

PR #1581, opened 2025-11-12
🐛(frontend) make summary button fixed to remain visible during scroll

Ovgodd:
## Purpose

Ensure the "Summary" button remains visible during user scrolls, especially on long documents

issue : https://github.com/suitenumerique/docs/issues/1578

## Proposal

- [x] Change button position from relative to fixed
- [x] Add right: 30px offset to keep it anchored to the viewport

### DIFF
--- a/src/frontend/apps/impress/src/features/docs/doc-table-content/components/TableContent.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-table-content/components/TableContent.tsx
@@ -108,9 +108,10 @@ export const TableContent = () => {
       $align="center"
       $padding={isHover ? 'xs' : '0'}
       $justify="center"
-      $position="relative"
+      $position="sticky"
       aria-label={t('Summary')}
       $css={css`
+        top: 0;
         border: 1px solid ${colorsTokens['greyscale-300']};
         overflow: hidden;
         border-radius: ${spacingsTokens['3xs']};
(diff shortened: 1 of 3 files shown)


src/frontend/apps/impress/src/features/docs/doc-table-content/components/TableContent.tsx
```diff
@@ -108,9 +108,10 @@ export const TableContent = () => {
       $align="center"
       $padding={isHover ? 'xs' : '0'}
       $justify="center"
-      $position="relative"
+      $position="fixed"
```
AntoLC: Could we make it sticky instead ?

```suggestion
      $position="sticky"
```

 If yes, to make it working you will have to set a height here (`height: 100vh;`): https://github.com/suitenumerique/docs/blob/e16f31341fd741a15542ca641755054912d2a167/src/frontend/apps/impress/src/features/docs/doc-editor/components/DocEditor.tsx#L101-L110

src/frontend/apps/impress/src/features/docs/doc-table-content/components/TableContent.tsx
```diff
@@ -108,9 +108,10 @@ export const TableContent = () => {
...
       aria-label={t('Summary')}
       $css={css`
+        right: 30px;
```
AntoLC: If you find sticky is nice, you can replace by:

```suggestion
        top: 0px;
```

Ovgodd: it is nice !

Issue #1573, opened 2025-11-07
Left panel expands when page resized


rl-83:
## Bug Report

The left panel expands when the window is resized, but it shouldn’t. The only way to enlarge it should be manually.
![Image](https://github.com/user-attachments/assets/628921e1-5cac-4bc2-8bf5-3fdbb794f056)

Ovgodd:
Hello @rl-83,

Following a discussion with @AntoLC forthis issue, we saw some flickering when resizing the window. 
We tested an alternative library, but the result wasn't convincing, it lacks accessibility support, is difficult to customize, and doesn't integrate well overall.

As for the current library, fixing the resize behavior, needed too put a "big bloc of code" just for a flickering result.

Would it be ok to put this issue on hold for now and keep the current implem while we look for a better long-term solution?

PR #1588, opened 2025-11-13
🐛(frontend) preserve left panel width on window resize

Ovgodd:
## Purpose

Fix a bug where the left panel unexpectedly expanded or shrank when the window was resized. The expected behavior is that the left panel width should remain constant unless manually adjusted by the user.

issue [1573](https://github.com/suitenumerique/docs/issues/1573)

https://github.com/user-attachments/assets/0cf330c7-1279-49c9-9059-c3d892a9e51c

## Proposal

- [x]  Remove global recalculation of min/max panel size on window resize
- [x]  Track panel width in pixels (savedWidthPxRef) and convert it to % only when needed
- [x]  Force the panel to maintain its width percentage equivalent after resize

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ and this project adheres to
 ### Security
 
 - mitigate role escalation in the ask_for_access viewset #1580
+- 🐛(frontend) preserve left panel width on window resize #1588
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


AntoLC (commented):
I can see some flickering: 

https://github.com/user-attachments/assets/d236e07c-a37a-45fe-b1d4-9962e03664ec

It is because [react-resizable-panels](https://github.com/bvaughn/react-resizable-panels) is percent based I think, so we have to compute the size every time to keep it fixed.

Could you give a try to this lib instead ?
https://github.com/johnwalley/allotment

AntoLC (approved):
Let's merge it.

Issue #1618, opened 2025-11-17
A11y: Localize aria-label in DocRoleDropdown


rvveber:
https://github.com/suitenumerique/docs/blob/a4e3168682bdfa91f5a526d80cdeef02df73488c/src/frontend/apps/impress/src/features/docs/doc-share/components/DocRoleDropdown.tsx#L113-L116

- **Problem**: The dropdown trigger hardcodes `aria-label="doc-role-dropdown"`, so assistive tech literally announces the English slug instead of a descriptive, localized label.
- **Impact**: Non-descriptive labels make it impossible for users to understand the control’s purpose, especially in translated interfaces.
- **Acceptance criteria / fix ideas**:
  - Replace the hardcoded string with a translated phrase (e.g., `t('Document role options')`).
  - Ensure the label matches the control’s function and is localized alongside the rest of the UI.

PR #1628, opened 2025-11-18
♿(frontend) improve screen reader support in DocShare modal with aria…

Ovgodd:
## Purpose

Improve accessibility in the DocShare modal by enhancing screen reader support.

this fix [this](https://github.com/suitenumerique/docs/issues/1618) issue too 

## Proposal

- [x] Add `ariaLabel` prop to `DocRoleDropdown` for contextual role change labels
- [x] Inject `ariaLabel` in `DocShareAccessRequest`, `Invitation`, and `Member` items
- [x] Implement live screen reader announcements in `DocShareModal` for:
  - Adding a user to the invite list
  - Removing a user from the invite list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to
 
 - ♿(frontend) improve accessibility:
   - ♿(frontend) improve share modal button accessibility #1626
+  - ♿(frontend) improve screen reader support in DocShare modal #1628
 - 🐛(frontend) fix toolbar not activated when reader #1640
 - 🐛(frontend) preserve left panel width on window resize #1588
 
@@ -51,7 +52,6 @@ and this project adheres to
 ### Removed
 
 - 🔥(backend) remove api managing templates
-
 
 ## [3.9.0] - 2025-11-10
 
--- a/src/frontend/apps/impress/src/features/docs/doc-share/components/DocShareModal.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-share/components/DocShareModal.tsx
@@ -76,6 +76,7 @@ export const DocShareModal = ({ doc, onClose, isRootDoc = true }: Props) => {
   const [selectedUsers, setSelectedUsers] = useState<User[]>([]);
   const [userQuery, setUserQuery] = useState('');
   const [inputValue, setInputValue] = useState('');
+  const [liveAnnouncement, setLiveAnnouncement] = useState('');
 
   const [listHeight, setListHeight] = useState<string>('400px');
   const canShare = doc.abilities.accesses_manage && isRootDoc;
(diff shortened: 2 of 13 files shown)


src/frontend/apps/impress/src/features/docs/doc-share/components/DocShareModal.tsx
```diff
@@ -175,12 +199,29 @@ export const DocShareModal = ({ doc, onClose, isRootDoc = true }: Props) => {
...
+          aria-atomic="true"
+          className="sr-only"
+          style={{
+            position: 'absolute',
+            left: '-10000px',
+            width: '1px',
+            height: '1px',
+            overflow: 'hidden',
+          }}
+        >
```
AntoLC: The style attach with `sr-only` is not enough ?

src/frontend/apps/impress/src/features/docs/doc-share/components/DocShareModal.tsx
```diff
@@ -88,6 +89,19 @@ export const DocShareModal = ({ doc, onClose, isRootDoc = true }: Props) => {
...
+      t(
+        '{{name}} added to invite list. Add more members or press Tab to select role and invite.',
+        {
+          name: userName,
+        },
+      ),
+    );
+    // Clear announcement after it's been read
+    setTimeout(() => setLiveAnnouncement(''), 100);
   };
```
AntoLC, line 105: I find a bit strange to do the announcement, we have a notification system, but I think we decided to not display a notification for that, because the invitation list will display the user that you just add, it is already like a announcement.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to
   - ♿(frontend) improve ARIA in doc grid and editor for a11y #1519
   - ♿(frontend) improve accessibility and styling of summary table #1528
   - ♿(frontend) add focus trap and enter key support to remove doc modal #1531
+  - ♿(frontend) improve screen reader support in DocShare modal #1628
```
AntoLC: During rebase, this part will have to go under `Unrelease`

src/frontend/apps/impress/src/features/docs/doc-share/components/DocShareModal.tsx
Ovgodd: oh yes, why I did that.

src/frontend/apps/impress/src/features/docs/doc-share/components/DocShareModal.tsx
Ovgodd: > I find a bit strange to do the announcement, we have a notification system, but I think we decided to not display a notification for that, because the invitation list will display the user that you just add, it is already like a announcement.

Hey ! the SR announcement is necessary for RGAA compliance. Users with visual "issues"cannot see the visual feedback when a user is added to the invite list. The aria-live region provides audio feedback only (no visual toast, this is a standard accessibility requirement

PR #1635, opened 2025-11-19
👷(CI) free disk space in some jobs

AntoLC:
## Purpose

An error appeared in the CI indicating that there is no space left on device on the e2e tests jobs.
To fix this, we free some disk space by removing unused docker images and some unnecessary large folders.


AntoLC:
Seems to have been the problem: https://github.com/suitenumerique/docs/pull/1629/commits/257de6d06895e979435d0e5ef52b34bfb3793684

PR #1637, opened 2025-11-19
🚸(backend) use unaccented full name for user search

### LINKED ISSUE #929: Use first name and last name for user search
## Bug Report

**Problematic behavior**
Sometimes it takes writing a whole email address to invite someone on a document. That's because we're only looking at email adresses while performing the search.

**Possible Solution**
Through OIDC we retrieve first and last name. We should use that info to provide results while the user types characters.

### LINKED ISSUE #1091: User search is sensitive to special characters
## Bug Report

**Problematic behavior**
If you search a user whose name contains a "é" you won't find it

**Expected behavior/code**
The search user function shouldn't be senstitive to that. It should translate "é" to "e"

**Steps to Reproduce**
1. Click on share
2. Type Josué
3. See no results
4. Type Josue
5. See results

## Possible solution

Search users is only based on the email, it should search on the full name as well.
https://github.com/suitenumerique/docs/blob/a09bd6c338a396262b8176cd01f0daaf68c52150/src/backend/core/api/viewsets.py#L164-L202

Ash-Crow:
## Purpose
We have the user full name through OIDC in the database, but the search only used the email field.

This change allows to search for a user by their first and/or last name (fix #929).

Given that user names are more likely than emails to include diacritics, it unaccents both the query and the database entry for search (fix #1091). It also unaccents for email so that internationalized domain names are managed whether or not the accent is included in the search.

## Proposal
- [x] Add the ability to search by full name
- [x] Unaccent both the query and the database search
- [x] Added an unaccented GIN index on the `email` and `full_name` fields of the User model.

Capture d’écran du 2025-11-19 14-44-22
Capture d’écran du 2025-11-19 11-30-59

## External contributions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to
 
 - ✨(backend) allow to create a new user in a marketing system
 
+### Changed
+
+- 🚸(backend) use unaccented full name for user search #1637
+
 ## [4.1.0] - 2025-12-09
 
 ### Added
@@ -64,7 +68,6 @@ and this project adheres to
 - ♻️(frontend) preserve @ character when esc is pressed after typing it #1512
 - ♻️(frontend) make summary button fixed to remain visible during scroll #1581
 - ♻️(frontend) pdf embed use full width #1526
-- 🚸(backend) use unaccented full name for user search #1637
 
 ### Fixed
 


src/backend/core/tests/test_api_users.py
```diff
@@ -76,6 +76,125 @@ def test_api_users_list_query_email():
...
+
+    response = client.get(
+        "/api/v1.0/users/?q=David",
+    )
+    assert response.status_code == 200
+    user_ids = [user["id"] for user in response.json()]
+    assert user_ids == [str(dave.id)]
+
+    response = client.get(
+        "/api/v1.0/users/?q=Bowman",
```
lunika: Can you add tests with case sensitivity check ? Does it work with `bowman` ?

lunika:
I think we can also introduce a new index on the User model ? One on the `fullname` colume and an other on the `email` column` ? WDYT ?

Ash-Crow:
> I think we can also introduce a new index on the User model ? One on the `fullname` colume and an other on the `email` column` ? WDYT ?

Good idea!

Ash-Crow:
It tried to create a GinIndex combined with Unaccent through an [OpClass](https://docs.djangoproject.com/en/5.2/ref/contrib/postgres/indexes/#opclass-expressions),

```python

indexes = [
    GinIndex(
        OpClass(Func(F("email"), function="unaccent"), name="gin_trgm_ops"),
        name="user_email_unaccent_trgm_idx",
    ),
    GinIndex(
        OpClass(Func(F("full_name"), function="unaccent"), name="gin_trgm_ops"),
        name="user_name_unaccent_trgm_idx",
    ),
]
```

but apparently it is impossibe because Unaccent is only STABLE, not IMMUTABLE (per [this StackOverflow thread](https://stackoverflow.com/questions/11005036/does-postgresql-support-accent-insensitive-collations))

There are a few workarounds but I'm not sure what is the best approach.

lunika:
Can you squash your commits ? They are all related, you can keep only one.

I added the preview label, once your push force made this PR will be available to test.

Ash-Crow:
I added the requested index.

For what is worth, I checked the queries through `EXPLAIN ANALYZE`:

```
Before the migration

 Planning Time: 0.137 ms
 Execution Time: 0.848 ms
(8 rows)

After the migration (same query)
 Planning Time: 0.266 ms
 Execution Time: 0.440 ms
(8 rows)

After the migration (removed redundant calls to Unaccent)

 Planning Time: 0.162 ms
 Execution Time: 0.335 ms
(8 rows)
```

PR #1640, opened 2025-11-20
🐛(frontend) fix toolbar not activated when reader

AntoLC:
## Purpose

When user was a reader of the document, the toolbar of the BlockNote editor was not activated, making it impossible to download resources like images. 
We add the toolbar even in viewer mode.
We block as well automatic document mutation from custom blocks when the editor is in viewer mode to avoid unwanted modifications.

### DIFF
--- a/src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-blocks/UploadLoaderBlock.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-blocks/UploadLoaderBlock.tsx
@@ -59,9 +59,15 @@ const UploadLoaderBlockComponent = ({
   editor,
 }: UploadLoaderBlockComponentProps) => {
   const mediaUrl = useMediaUrl();
+  const isEditable = editor.isEditable;
 
   useEffect(() => {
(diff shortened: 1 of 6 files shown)


src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-blocks/UploadLoaderBlock.tsx
```diff
@@ -59,9 +59,14 @@ const UploadLoaderBlockComponent = ({
...
 }: UploadLoaderBlockComponentProps) => {
   const mediaUrl = useMediaUrl();
+  const isEditable = editor.isEditable;
 
   useEffect(() => {
-    if (!block.props.blockUploadUrl || block.props.type !== 'loading') {
+    if (
+      !block.props.blockUploadUrl ||
+      block.props.type !== 'loading' ||
+      !isEditable
```
Ovgodd:
```suggestion
    const shouldSkip =   !block.props.blockUploadUrl ||
      block.props.type !== 'loading' ||
      !isEditable
    if (shouldSkip)
```

PR #1644, opened 2025-11-23
🌐(backend) internationalize demo

Ash-Crow:
## Purpose
This allows the demo to generate user and filenames with other locales than English, for all languages defined in the project's settings. In particular, it allows the generation of accented names, which were previously missing.

## Proposal
- [x] Use the list of languages defined in settings in the call to `Faker()`

## Image
Capture d’écran du 2025-11-23 11-46-05
Capture d’écran du 2025-11-23 11-48-36


virgile-dev:
This is really cool thanks @Ash-Crow !

PR #1648, opened 2025-11-24
✅(e2e) add test to compare generated PDF against reference template

Ovgodd:
## Purpose

Add a stable **PDF export regression test** that fails whenever the exported PDF for a full, complex document changes.

    1. `npx playwright test __tests__/app-impress/doc-export.spec.ts --update-snapshots -g "full document" --project=chromium`
       -> refresh the reference PDF whenever we intentionally change the export output.
    2. `npx playwright test __tests__/app-impress/doc-export.spec.ts -g "full document" --project=chromium`
       -> CI (and local runs without --update-snapshots) will compare the PDF to the reference
          and fail on any byte-level difference once the dynamic metadata has been sanitized.

## Proposal

- [x] Seed a doc with a **fixed title** and a **prebuilt BlockNote/Yjs content** (`content.txt`) via the REST API (`PATCH /documents/:id/`), using `X-CSRFToken` from Playwright cookies.
- [x] Export to PDF on **Chromium**, sanitize dynamic metadata (dates, IDs, font prefixes),
  then compare bytes against a **snapshot PDF**.
- [x] Document how to maintain the snapshot (`--update-snapshots` vs. normal run) directly in the test.
- [x] Add `print-datauris.js` to generate PNG/JPG/SVG **data URL helpers** and update `content.txt`
  so the reference document includes **embedded images** via data URLs.


AntoLC:
Closed in favor of https://github.com/suitenumerique/docs/pull/1762

PR #1662, opened 2025-11-27
⚡️(frontend) improve perf on upload and table of contents

AntoLC:
## Purpose

During investigations about some users having theirs docs quite slow, we found a method that slow down a lot the users typing on a big documents.

https://github.com/suitenumerique/docs/blob/2864669dde3da6c56df2298cf15c24ce52e1fd36/src/frontend/apps/impress/src/features/docs/doc-editor/hook/useUploadFile.tsx#L45

## Proposal

- [x] We are improving the way we listen a upload to avoid this methods.
- [x] We are improving the listeners on the table of content to listen only when necessary, we added a debounce as well.
- [x] We fixed a bug on the "Table of Content" about stickiness blocked after a page bigger that `100vh`.

### DIFF
--- a/src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-blocks/UploadLoaderBlock.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-blocks/UploadLoaderBlock.tsx
@@ -75,27 +75,31 @@ const UploadLoaderBlockComponent = ({
 
     loopCheckDocMediaStatus(url)
       .then((response) => {
-        // Replace the loading block with the resource block (image, audio, video, pdf ...)
-        try {
-          editor.replaceBlocks(
-            [block.id],
-            [
-              {
-                type: block.props.blockUploadType,
-                props: {
-                  url: `${mediaUrl}${response.file}`,
-                  showPreview: block.props.blockUploadShowPreview,
-                  name: block.props.blockUploadName,
-                  caption: '',
-                  backgroundColor: 'default',
-                  textAlignment: 'left',
-                },
-              } as never,
-            ],
-          );
-        } catch {
-          /* During collaboration, another user might have updated the block */
-        }
+        // Add random delay to reduce collision probability during collaboration
+        const randomDelay = Math.random() * 800;
+        setTimeout(() => {
+          // Replace the loading block with the resource block (image, audio, video, pdf ...)
+          try {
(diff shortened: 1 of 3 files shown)


src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-blocks/UploadLoaderBlock.tsx
```diff
@@ -75,27 +75,31 @@ const UploadLoaderBlockComponent = ({
...
-                  textAlignment: 'left',
-                },
-              } as never,
-            ],
-          );
-        } catch {
-          /* During collaboration, another user might have updated the block */
-        }
+        // Add random delay to reduce collision probability during collaboration
+        const randomDelay = Math.random() * 800;
```
Ovgodd, line 79: Smart I like it

src/frontend/apps/impress/src/features/docs/doc-editor/hook/useHeadings.tsx
```diff
@@ -9,11 +9,33 @@ export const useHeadings = (editor: DocsBlockNoteEditor) => {
...
+
+        if (
+          blockChanges.type !== 'update' ||
+          blockChanges.block.type !== 'heading'
+        ) {
+          return;
+        }
+
+        setHeadings(editor);
+      }, 500);
```
Ovgodd: maybe store it in const DEBOUNCE_DELAY = 500;

src/frontend/apps/impress/src/features/docs/doc-editor/hook/useUploadFile.tsx
```diff
@@ -36,73 +37,93 @@ export const useUploadFile = (docId: string) => {
...
-      if (!blocksChanges.length) {
-        return;
-      }
-
-      const blockChanges = blocksChanges[0];
-
+  /**
+   * Replace the resource block by a uploadLoader block to show analyzing status
+   */
+  const replaceBlockWithUploadLoader = useCallback(
```
Ovgodd, line 55: Would it make sense to extract this into a utility function if we plan to reuse it later?

AntoLC: It is hypothetical, let's do that when the situation would happen.

Issue #599, opened 2025-01-30
Move copy as markdown and HTML to the download menu


rl-83:
**Is your feature request related to a problem or unsupported use case? Please describe.**
The "Copy as Markdown" and "Copy as HTML" functions in the page dropdown menu don't seem useful for average users. The use case seems far too limited to be featured as prominent functions. This dropdown menu should primarily serve the everyday tasks of the majority of our users.
![Image](https://github.com/user-attachments/assets/5ec1fc8f-ec73-4b45-bc17-093fa27f7987)

**Describe the solution you'd like**
Remove both functions from the dropdown menu.

**Describe alternatives you've considered**
If these functions are deemed essential, we could move them to the download window. To make a decision, analyzing the actual usage of these functions with posthog would be great.
![Image](https://github.com/user-attachments/assets/016de589-d254-40d7-9c73-b632807d6474)

rvveber:
Fully agree with this proposal 👍🏼

arnaud-robin:
Also agree that we should move them to the export panel.

virgile-dev:
Agree on moving this to the export panel, but keep the html it can be useful.
When html or markdown are selected the template dropdown should be greyed or not displayed. @rl-83 can you update your proposal ?

AntoLC:
Could you provide a small design about it @rl-83 ?

rl-83:
Will be done next week!

rl-83:
Here’s my suggestion for adding HTML to the export options. If the HTML format is selected, the template choice becomes grayed out. Therefore, the two fields are reversed compared to their current setup.

-> [Resources here](https://www.figma.com/design/qdCWR4tTUr7vQSecEjCyqO/DOCS?node-id=8298-14380&t=UVbhYBw32LIFPTvH-1)

![Image](https://github.com/user-attachments/assets/8bcff200-1eeb-4a8d-9074-f9f371a30fa9)

virgile-dev:
Hey @rl-83 
I think for markdown export and html exports we need to think a bit more about it.
Docs will contain medias, medias have permissions. We should export a doc as zip file with the medias.

laem:
Can confirm this : I came here in the issues section to create new issue "Why can't I export the doc as markdown" only to find that it's another button. 

Regards, keep up the good work ;)

PR #1669, opened 2025-12-01
⚡️Enhance/html copy to download

Ovgodd:
## Purpose

Enable users to export an accessible HTML version of content, replacing the "Copy as HTML" option with a full export feature from the modal.

issue : [599](https://github.com/suitenumerique/docs/issues/599)

image

## Proposal

- [x] Move "Copy as HTML" to the export modal
- [x] Add a new accessible HTML export format (.html in ZIP)
- [x] Embed media files (images, SVG, PDF, video/audio) in the ZIP with clean filenames
- [x] Update export modal UI and labels
- [x] Add E2E test to validate HTML export ZIP structure and content


src/frontend/apps/impress/src/features/docs/doc-header/__tests__/DocToolBox.spec.tsx
AntoLC, line 1: You can remove totally this file, it was necessary to test the feature flag thanks to Analytics class but if we don't use it anymore, better to remove it totally.

src/frontend/apps/impress/src/features/docs/doc-header/__tests__/DocToolBoxLicence.spec.tsx
```diff
@@ -42,12 +42,13 @@ describe('DocToolBox - Licence', () => {
...
+    // Wait for the export modal to be visible, then assert on its content text.
+    await screen.findByTestId('modal-export-title');
     expect(
-      await screen.findByText(
-        'Download your document in a .docx, .odt or .pdf format.',
-      ),
+      screen.getByText(/Download your document in a .docx, .odt.*format\./i),
     ).toBeInTheDocument();
-  }, 10000);
+  }, 60000);
```
AntoLC: Do you really need 1mn ?

src/frontend/apps/impress/src/features/docs/doc-export/__tests__/ExportMIT.test.tsx
```diff
@@ -16,12 +16,12 @@ describe('useModuleExport', () => {
     const Export = await import('@/features/docs/doc-export/');
 
     expect(Export.default).toBeUndefined();
-  }, 10000);
+  }, 60000);
```
AntoLC: Same, is 1mn really necessary, seems big ?

src/frontend/apps/e2e/__tests__/app-impress/doc-header.spec.ts
```diff
@@ -408,7 +408,10 @@ test.describe('Doc Header', () => {
     expect(clipboardContent.trim()).toBe('# Hello World');
   });
 
-  test('It checks the copy as HTML button', async ({ page, browserName }) => {
+  test('It no longer shows the copy as HTML button', async ({
```
AntoLC: You can remove totally this test, we don't assert things that does not exist anymore.

src/frontend/apps/impress/src/features/docs/doc-export/components/ModalExport.tsx
```diff
@@ -142,13 +144,122 @@ export const ModalExport = ({ onClose, doc }: ModalExportProps) => {
       });
 
       blobExport = await exporter.toODTDocument(exportDocument);
+    } else if (format === DocDownloadFormat.HTML) {
+      const editorHtml = await editor.blocksToHTMLLossy();
```
AntoLC:
```suggestion
      const editorHtml = editor.blocksToHTMLLossy();
```

Have you seen that you have `editor.blocksToFullHTML` available as well ?

src/frontend/apps/impress/src/features/docs/doc-export/components/ModalExport.tsx
```diff
@@ -142,13 +144,122 @@ export const ModalExport = ({ onClose, doc }: ModalExportProps) => {
...
+              filename = `${baseName}.${extension}`;
+            }
+          }
+
+          element.setAttribute('src', filename);
+          mediaFiles.push({ filename, blob: fetched });
+        }),
+      );
+
+      const lang = i18next.language || 'fr';
```
AntoLC:
```suggestion
      const lang = i18next.language || fallbackLng;
```

From: https://github.com/suitenumerique/docs/blob/999c81499a087214c8de72575f202bbfc5e015a1/src/frontend/apps/impress/src/i18n/config.ts#L1-L2

src/frontend/apps/impress/src/features/docs/doc-export/components/ModalExport.tsx
```diff
@@ -245,12 +348,24 @@ export const ModalExport = ({ onClose, doc }: ModalExportProps) => {
...
+          clearable={false}
+          fullWidth
+          label={t('Template')}
+          options={templateOptions}
+          value={templateSelected}
+          disabled={format === DocDownloadFormat.HTML}
+          onChange={(options) =>
+            setTemplateSelected(options.target.value as string)
+          }
+        />
```
AntoLC: Can we totally removed the template choice when html ?

```suggestion
      { format !== DocDownloadFormat.HTML && 
        <Select
          clearable={false}
          fullWidth
          label={t('Template')}
          options={templateOptions}
          value={templateSelected}
          onChange={(options) =>
            setTemplateSelected(options.target.value as string)
          }
        />
      }

```

src/frontend/apps/impress/src/features/docs/doc-export/components/ModalExport.tsx
```diff
@@ -142,13 +144,122 @@ export const ModalExport = ({ onClose, doc }: ModalExportProps) => {
...
+      await Promise.all(
+        mediaElements.map(async (element, index) => {
+          const src = element.getAttribute('src');
+
+          if (!src) {
+            return;
+          }
+
+          const fetched = await exportCorsResolveFileUrl(doc.id, src);
+
```
AntoLC: I can see you use the CORS proxy, but I think it works only with images:
https://github.com/suitenumerique/docs/blob/999c81499a087214c8de72575f202bbfc5e015a1/src/backend/core/api/viewsets.py#L1551-L1555

src/frontend/apps/impress/src/features/docs/doc-export/components/ModalExport.tsx
```diff
@@ -142,13 +144,122 @@ export const ModalExport = ({ onClose, doc }: ModalExportProps) => {
...
+              // Ignore invalid URLs, keep default baseName.
+            }
+          }
+
+          let filename = baseName;
+
+          // Ensure the filename has an extension consistent with the blob MIME type.
+          const mimeType = fetched.type;
+          if (mimeType && !baseName.includes('.')) {
+            const subtype = mimeType.split('/')[1] || '';
```
AntoLC: I have the feeling it is error prone, what if the mimeType does not have a "/" ?

src/frontend/apps/impress/src/features/docs/doc-export/components/ModalExport.tsx
```diff
@@ -142,13 +144,122 @@ export const ModalExport = ({ onClose, doc }: ModalExportProps) => {
...
+
+            if (extension) {
+              filename = `${baseName}.${extension}`;
+            }
+          }
+
+          element.setAttribute('src', filename);
+          mediaFiles.push({ filename, blob: fetched });
+        }),
+      );
```
AntoLC: This part deserve its own function I think, it will increase readability.
It is doing lot of string manipulations as well, could be nice to have a few unit tests to be sure that it is working as expected, wdyt ?

src/frontend/apps/impress/src/features/docs/doc-export/components/ModalExport.tsx
```diff
@@ -142,13 +144,122 @@ export const ModalExport = ({ onClose, doc }: ModalExportProps) => {
...
+    <meta charset="utf-8" />
+    <title>${escapeHtml(documentTitle)}</title>
+  </head>
+  <body>
+    <main role="main">
+${editorHtmlWithLocalMedia}
+    </main>
+  </body>
+</html>`;
+
```
AntoLC: Maybe its own function to improve readability as well ?

src/frontend/apps/e2e/__tests__/app-impress/doc-export.spec.ts
```diff
@@ -187,6 +187,64 @@ test.describe('Doc Export', () => {
...
+
+    const download = await downloadPromise;
+    expect(download.suggestedFilename()).toBe(`${randomDoc}.zip`);
+
+    const zipBuffer = await cs.toBuffer(await download.createReadStream());
+
+    // ZIP files start with "PK\x03\x04"
+    expect(zipBuffer.length).toBeGreaterThan(4);
+    expect(zipBuffer[0]).toBe(0x50);
+    expect(zipBuffer[1]).toBe(0x4b);
```
AntoLC: Could we improve this part to check what is inside the zip file ?

The idea: 
```ts
import JSZip from 'jszip';

// In your test:
const zipBuffer = await cs.toBuffer(await download.createReadStream());

// Unzip and inspect contents
const zip = await JSZip.loadAsync(zipBuffer);

// Check that index.html exists
const indexHtml = zip.file('index.html');
expect(indexHtml).not.toBeNull();

// Read and verify HTML content
const htmlContent = await indexHtml!.async('string');
expect(htmlContent).toContain('Hello HTML ZIP');

// Check for media files
const mediaFiles = zip.file(/^media\//);
expect(mediaFiles.length).toBeGreaterThan(0);

// Verify the SVG image is included
const svgFile = mediaFiles.find(f => f.name.endsWith('.svg'));
expect(svgFile).toBeDefined();
```

src/frontend/apps/impress/src/features/docs/doc-header/__tests__/DocToolBoxLicence.spec.tsx
Ovgodd: Not anymore,it was only for debugging the test (the issue came from waiting for a specific text, which is now fixed).
thanks !

src/frontend/apps/impress/src/features/docs/doc-export/__tests__/ExportMIT.test.tsx
Ovgodd: The test was a bit flaky, I saw it fail occasionally and one run even took ~9s, so I put timeout to 1min a bit aggressively 😅
We could probably reduce it to 15–20s now ? Wdyt

src/frontend/apps/impress/src/features/docs/doc-export/components/ModalExport.tsx
Ovgodd: Yes, indeed, I’ve just switched to editor.blocksToFullHTML(), and it preserves the original document structure much better than blocksToHTMLLossy.
This is the best option BlockNote provides, I think.

src/frontend/apps/impress/src/features/docs/doc-export/components/ModalExport.tsx
Ovgodd: Oh yes, I see, the current CORS proxy only supports images.
So that means external audio / video (when they’re not uploaded to our backend) are rejected and can’t be added to the HTML ZIP? ( as you said sometimes mp3 and vidéos are not exported ) 
In that case, should we update the backend proxy to also allow audio and video so these files can be exported too?

src/frontend/apps/impress/src/features/docs/doc-export/components/ModalExport.tsx
Ovgodd: Yes, good catch,
I’ve updated it so we first check for the / is present otherwise we skip the extension logic and keep the base filename

src/frontend/apps/impress/src/features/docs/doc-export/components/ModalExport.tsx
Ovgodd: Yes I completely agree !

src/frontend/apps/impress/src/features/docs/doc-export/components/ModalExport.tsx
Ovgodd: Yep ! good catch

src/frontend/apps/e2e/__tests__/app-impress/doc-export.spec.ts
Ovgodd: yes I agree this is much better

src/frontend/apps/e2e/__tests__/app-impress/doc-export.spec.ts
```diff
@@ -187,6 +188,89 @@ test.describe('Doc Export', () => {
...
+
+    // Read and verify HTML content
+    const htmlContent = await indexHtml!.async('string');
+    console.log('HTML content preview:', htmlContent.substring(0, 1000));
+    expect(htmlContent).toContain('Hello HTML ZIP');
+
+    // Check for media files (they are at the root of the ZIP, not in a media/ folder)
+    // Media files are named like "1-test.svg" or "media-1.png" by deriveMediaFilename
+    const allFiles = Object.keys(zip.files);
+    console.log('All files in ZIP:', allFiles);
```
AntoLC:
```suggestion
```

src/frontend/apps/e2e/__tests__/app-impress/doc-export.spec.ts
```diff
@@ -187,6 +188,89 @@ test.describe('Doc Export', () => {
...
+    // Unzip and inspect contents
+    const zip = await JSZip.loadAsync(zipBuffer);
+
+    // Check that index.html exists
+    const indexHtml = zip.file('index.html');
+    expect(indexHtml).not.toBeNull();
+
+    // Read and verify HTML content
+    const htmlContent = await indexHtml!.async('string');
+    console.log('HTML content preview:', htmlContent.substring(0, 1000));
```
AntoLC:
```suggestion
```

src/frontend/apps/e2e/__tests__/app-impress/doc-export.spec.ts
```diff
@@ -187,6 +188,89 @@ test.describe('Doc Export', () => {
...
+    expect(htmlContent).toContain('Hello HTML ZIP');
+
+    // Check for media files (they are at the root of the ZIP, not in a media/ folder)
+    // Media files are named like "1-test.svg" or "media-1.png" by deriveMediaFilename
+    const allFiles = Object.keys(zip.files);
+    console.log('All files in ZIP:', allFiles);
+    const mediaFiles = allFiles.filter(
+      (name) => name !== 'index.html' && !name.endsWith('/'),
+    );
+    console.log('Media files found:', mediaFiles);
```
AntoLC:
```suggestion
```

src/frontend/apps/impress/src/features/docs/doc-export/__tests__/ExportMIT.test.tsx
AntoLC: Yes 15 or 20 max,  if it takes more than that it means there is a problem somewhere.

CHANGELOG.md
```diff
@@ -36,6 +37,9 @@ and this project adheres to
 - ⚡️(sw) stop to cache external resources likes videos #1655
 - 💥(frontend) upgrade to ui-kit v2 #1605
 - ⚡️(frontend) improve perf on upload and table of contents #1662
+
+### Fixed
+
```
AntoLC:
```suggestion
```

src/frontend/apps/impress/src/features/docs/doc-export/components/ModalExport.tsx
AntoLC: There are 2 types of "link" (internal / external), if the resources are internal like `https://docs.numerique.gouv.fr/media/...`, we have to download them and put them in the zip files because they will not be accessible otherwise; on the other hand, if the resources are external like `https://youtube.com/kefseklsfes.mp4`, I don't think we should download them but just let the `src` as it was, moreover we don't have the hand on the external resources they could be huge (1go) or even infected.

Issue #1160, opened 2025-07-10
Accessibilité - Sous Doc


selimmeziti:
## Accessibilité au clavier

La nouvelle fonctionnalité d'ajout de sous documents n'est pas accessible au clavier.
Les boutons d'ajout de document ne sont pas atteignable au clavier et l'usage de la souris est indispensable pour les afficher (survol).
Il est donc impossible à utiliser au clavier et sur écran tactile.

virgile-dev:
Hello @selimmeziti 
Thanks for the issue !
Docs welcomes an international community of contributors. Can we keep issues in English please ?
Feel free to use a translator on this one.
Cheers

Erwan-le-Gall:
# Keyboard accessibility

The new functionality to add subdocuments is not available thru keyboard.

The buttons to add documents can't be reached using keyboard. And using the mouse is mandatory to show them by hovering them.

It is therefore impossible to use only with keyboard or touch.

@selimmeziti @virgile-dev

AntoLC:
@Ovgodd is it completed ? Could we close this issue ?

Ovgodd:
> [@Ovgodd](https://github.com/Ovgodd) is it completed ? Could we close this issue ?

this issue is resolved in the UI Kit via the hook I introduced. For now, it's not yet effective in `docs` because we haven’t updated to the latest UI Kit version.

This upgrade is included in [PR #1605](https://github.com/suitenumerique/docs/pull/1605). Once this PR is merged, I’ll be able to confirm the fix is working in `docs`.

Keeping the issue open until we can validate it

PR #1681, opened 2025-12-02
♿️Improve keyboard accessibility for the document tree

Ovgodd:
## Purpose

Improve keyboard accessibility for the document tree by implementing F2 navigation to access action buttons (three dots menu and add child button) for both root items and sub-pages.

issues : [1160](https://github.com/suitenumerique/docs/issues/1160) 

https://github.com/user-attachments/assets/025a2d33-7129-4e72-b922-9df8631779b2

### Keyboard Navigation

| Key | Context | Action |
|-----|---------|--------|
| **Tab** | Document tree | Navigate between documents (root → sub-pages) |
| **F2** | Focused document | Move focus to first action button (emoji picker & options)  |
| **Enter** | Focused document | Open the document |
| **Enter** | Action button | Activate button (open menu / create child) |
| **Arrow Left/Right** | Action buttons | Navigate between buttons |
| **Escape** | Action buttons | Return focus to document |

### Changes
- [x] Use ref to handle focus
- [x] Display actions when document is focused (like hover)
- [x] Hide document focus outline when focus is on actions

### Accessibility

- ✅ Actions are visible when document is focused (keyboard users see them)
- ✅ Actions are not in tab order (reduces tab stops)
- ✅ F2 is the standard key for accessing toolbars (ARIA best practices)
- ✅ Arrow keys for navigation within toolbar (ARIA roving tabindex pattern)
- ✅ Escape returns to main content (standard pattern)
- ✅ Screen reader announces keyboard shortcuts (F2, arrows, Enter) when entering the tree


src/frontend/apps/impress/src/features/docs/doc-tree/components/DocTree.tsx
```diff
@@ -234,20 +273,25 @@ export const DocTree = ({ currentDoc }: DocTreeProps) => {
...
             &:hover,
-            &:focus-visible {
+            &:focus-visible,
+            &:focus-within {
               .doc-tree-root-item-actions {
                 display: flex;
                 opacity: 1;
               }
             }
+            /* Retirer le focus visuel du root item quand le focus est sur les actions */
```
AntoLC:
```suggestion
            /* Remove visual focus from the root item when focus is on the actions */
```

src/frontend/apps/impress/src/features/docs/doc-tree/components/DocSubPageItem.tsx
```diff
@@ -127,6 +130,13 @@ export const DocSubPageItem = (props: TreeViewNodeProps<Doc>) => {
           outline: none !important;
           box-shadow: 0 0 0 2px var(--c--globals--colors--brand-500) !important;
           border-radius: var(--c--globals--spacings--st);
+          .light-doc-item-actions {
+            display: flex;
+          }
+        }
+        /* Retirer le focus visuel du tree item quand le focus est sur les actions */
```
AntoLC: Comments should be in english.

src/frontend/apps/impress/src/features/docs/doc-tree/components/DocTreeItemActions.tsx
```diff
@@ -160,30 +171,54 @@ export const DocTreeItemActions = ({
...
         $gap="4px"
+        tabIndex={-1}
       >
         <DropdownMenu
           options={options}
           isOpen={isOpen}
           onOpenChange={onOpenChange}
         >
-          <Icon
+          <Box
```
AntoLC: What about `BoxButton` ?

src/frontend/apps/impress/src/features/docs/doc-tree/hooks/useActionableMode.ts
```diff
@@ -0,0 +1,83 @@
...
+      // TreeView may reclaim focus after this event cycle; setTimeout guarantees focus happens after
+      setTimeout(() => {
+        first.focus();
+      }, 0);
+    };
+
+    document.addEventListener('keydown', toActions, true);
+    return () => document.removeEventListener('keydown', toActions, true);
+    // node is a dependency, as it's checked for focus state
+  }, [node]);
```
AntoLC:
```suggestion
  }, [node?.isFocused]);
```

src/frontend/apps/impress/src/features/docs/doc-tree/hooks/useActionableMode.ts
```diff
@@ -0,0 +1,83 @@
...
+    const toActions = (e: KeyboardEvent) => {
+      if (e.key !== 'F2' || document.querySelector(SELECTORS.MODAL)) {
+        return;
+      }
+
+      // Only react if the node is currently focused
+      if (!node?.isFocused) {
+        return;
+      }
+
```
AntoLC: I am not fond of coupling with `dom-selectors`. If a modal is opened I suppose the node would not be focused anymore, so is it not enough that ?
```suggestion
      if (e.key !== 'F2' || !node?.isFocused) {
        return;
      }
```

src/frontend/apps/impress/src/features/docs/doc-tree/dom-selectors.ts
AntoLC, line 1: Can be removed as well.

src/frontend/apps/impress/src/features/docs/doc-tree/components/DocTree.tsx
```diff
@@ -88,18 +91,45 @@ export const DocTree = ({ currentDoc }: DocTreeProps) => {
     selectRoot();
   }, [selectRoot]);
 
-  // activate root document with enter or space
+  // Handle keyboard navigation for root item
   const handleRootKeyDown = useCallback(
     (e: React.KeyboardEvent) => {
+      // F2: focus first action button
+      if (e.key === 'F2' && !rootActionsOpen) {
```
AntoLC, line 98: Can we assert that in a e2e test ?

src/frontend/apps/impress/src/features/docs/doc-tree/components/DocTree.tsx
```diff
@@ -199,13 +230,20 @@ export const DocTree = ({ currentDoc }: DocTreeProps) => {
         }
       `}
     >
+      {/* Instructions clavier pour les lecteurs d'écran */}
```
AntoLC: English

src/frontend/apps/impress/src/features/docs/doc-tree/components/DocTreeItemActions.tsx
```diff
@@ -160,30 +173,52 @@ export const DocTreeItemActions = ({
...
+              display: flex;
+              align-items: center;
+              justify-content: center;
+
+              &:focus-visible {
+                outline: 2px solid var(--c--globals--colors--brand-500) !important;
+                outline-offset: -2px;
+                border-radius: var(--c--globals--spacings--st);
+              }
+            `}
```
AntoLC, line 206: BoxButton has already all that:
https://github.com/suitenumerique/docs/blob/c109288c1fb55ccb74dd8e9e50090c2073c572b5/src/frontend/apps/impress/src/components/BoxButton.tsx#L38-L53

src/frontend/apps/impress/src/features/docs/doc-tree/components/DocTreeItemActions.tsx
```diff
@@ -199,6 +234,14 @@ export const DocTreeItemActions = ({
             $variation="secondary"
             aria-label={t('Add a sub page')}
             data-testid="doc-tree-item-actions-add-child"
+            tabIndex={-1}
+            $css={css`
```
AntoLC, line 231: Same BoxButton has already that, you just need to set the correct `theme` and `variation` props: 
https://github.com/suitenumerique/docs/blob/c109288c1fb55ccb74dd8e9e50090c2073c572b5/src/frontend/apps/impress/src/components/BoxButton.tsx#L45-L51

src/frontend/apps/impress/src/features/docs/doc-tree/hooks/useActionableMode.ts
Ovgodd: This file has been deleted since our discussion about this implementation

src/frontend/apps/impress/src/features/docs/doc-tree/hooks/useActionableMode.ts
Ovgodd: This file has beed deleted since our discussion about this implementation

Ovgodd:
> I found the problem about the double `f2`, it is intercepted by `treeitem`.
> 
> Image
> It positions the focus on the emojipicker ^^ If you do `f2` then enter, the picker open.
> 
> I suppose it is not the best that : https://github.com/suitenumerique/ui-kit/blob/0b6f80100a85641c8227b4a61720722ec12691e2/src/components/tree-view/TreeView.tsx#L351-L359 From here: [suitenumerique/ui-kit@c40f7f6](https://github.com/suitenumerique/ui-kit/commit/c40f7f6c0d8573ae75f0f7eab35bb4be1b44c24f)
> 
> We need a way to pass props easily to this part, if we can have the hand on `props.attr` we should be able to add our own `onKeyDown`, but for that we need to move this line (l378 `{...props.attrs}`) to the line under `onKeyDown` (l390).(https://github.com/suitenumerique/ui-kit/blob/0b6f80100a85641c8227b4a61720722ec12691e2/src/components/tree-view/TreeView.tsx#L378-L379).

Ohhh it’s a really nice catch, I don’t know how I missed that the first F2 actually opens the emoji picker.

When I built this feature I only thought about the emoji picker later, so this is actually perfect, because we do need to be able to access it with the keyboard.

For your suggestion on the ui-kit side (moving `{...props.attrs}` below the `onKeyDown` in `TreeView.tsx`), I agree that the clean solution is to handle F2 directly in the row via `onKeyDown` instead of relying on workarounds here.

If I may, I propose to make some little fixup anyway ?  because we have an accessibility audit coming very soon, and updating the ui-kit + bumping it in this on docs will probably take a bit of time ? 

- first F2 moves focus (visually as well) to the emoji button when available
- second F2 moves focus to the document actions
- the SR announcement is updated to describe this behavior

Once the ui-kit exposes `onKeyDown` properly on the tree row, I’d be happy to follow up with another PR to remove this workaround and align the behavior with the upstream TreeView.

src/frontend/apps/impress/src/features/docs/doc-tree/components/DocTree.tsx
Ovgodd: yes of course !

src/frontend/apps/impress/src/features/docs/doc-tree/components/DocTreeItemActions.tsx
Ovgodd: Yes indeed, thanks ! I removed all duplicates

AntoLC (approved):
Looks good 👍

PR #1690, opened 2025-12-04
✨(auth) Add silent login

arnaud-robin:
## Problem

Right now users already logged in to the SSO have to click on the login button again to be connected. This extra step should not be necessary and some partners have complained

## Proposed solution

This PR uses the "silent=true" parameter in the lasuite login endpoint to avoid showing the IdP login screen if user already have an active session.

## Demo

https://github.com/user-attachments/assets/f9de05a8-6ba0-47e9-9c10-a2d287ab3b30

## ⚠️  Limitation

This feature does not work in development mode (localhost:3000 / localhost:8071). 
When frontend and backend are on different origins (and/or IdP is another origin), silent auth is much more likely to fail due to browser privacy rules and cookie context.
Same-domain setups tend to “just work” because everything is first-party and the IdP session is more reliably available.

## ⚠️  New env

To activate this feature:
```
FRONTEND_SILENT_LOGIN_ENABLED=true
```


arnaud-robin:
This implementation is heavily copied from @lebaudantoine work.

virgile-dev:
@AntoLC fyi

AntoLC (commented):
We need a test to assert that it is working, the tests don't seem happy btw.

If I understand correctly, even if our cookies is dead (frontend side), if the SSO recognize the user it will not prompt the user again, isn't it ?

arnaud-robin:
Yes, I’ll update the tests accordingly. I also plan to make this option configurable so that self-hosted instances using identity providers that don’t support this parameter don’t end up with a broken login flow.

But you're right, the goal is that even if the user has never logged into Docs before, or if the Docs session has expired, we still try to reuse an existing IdP session so the user doesn’t have to click the login button.

AntoLC:
:rocket: Preview will be available at [https://1609-docs.ppr-docs.beta.numerique.gouv.fr/](https://1690-docs.ppr-docs.beta.numerique.gouv.fr/)

You can use the existing account with these credentials:
- username: `docs`
- password: `docs`

You can also create a new account if you want to.

Once this Pull Request is merged, the preview will be destroyed.

src/frontend/apps/impress/src/pages/docs/[id]/index.tsx
```diff
@@ -171,19 +171,25 @@ const DocPage = ({ id }: DocProps) => {
     });
   }, [addTask, doc?.id, queryClient]);
 
-  if (isError && error) {
-    if ([404, 401].includes(error.status)) {
-      let replacePath = `/${error.status}`;
+  useEffect(() => {
+    if (isError && error?.status) {
+      if ([404, 401].includes(error.status)) {
+        let replacePath = `/${error.status}`;
```
Comment (Ovgodd) [inline src/frontend/apps/impress/src/pages/docs/[id]/index.tsx]: may I suggest early return here ?
I think it is more cleaner, readable and can avoid some undesired sideeffects ?
```suggestion
useEffect(() => {
  if (!isError || !error?.status) return;
  if (![401, 404].includes(error.status)) return;

  if (error.status === 401) {
    if (authenticated) {
      queryClient.setQueryData([KEY_AUTH], null);
    }
    setAuthUrl();
  }

  void replace(`/${error.status}`);
}, [isError, error?.status, replace, authenticated, queryClient]);

```

src/backend/impress/settings.py
```diff
@@ -548,6 +550,8 @@ class Base(Configuration):
     SESSION_COOKIE_NAME = "docs_sessionid"
 
     # OIDC - Authorization Code Flow
+    OIDC_AUTHENTICATE_CLASS = "lasuite.oidc_login.views.OIDCAuthenticationRequestView"
+    OIDC_CALLBACK_CLASS = "lasuite.oidc_login.views.OIDCAuthenticationCallbackView"
```
lunika: I would use `values.Value` to allow changing them. WDYT ?

Ovgodd (approved):

PR #1697, opened 2025-12-05
✨(helm) redirecting system

AntoLC:
## Purpose

We want to replace the `LaSuite` Docs landing page (https://docs.numerique.gouv.fr/home/) by the new one (https://lasuite.numerique.gouv.fr/produits/docs).
We will redirect to it if the app targets `/home/` page.
To target only our instance it will be configurable thanks to the helm charts.

## How to use

In your `env.d` folder set the `ingressRedirects` as you want, it can be a array of redirections, redirecting with a special code (_default code 301_):

example: `./src/helm/env.d/feature/values.impress.yaml.gotmpl`
```yaml
ingressRedirects:
  enabled: true
  defaultHost: {{ .Values.feature }}-docs.{{ .Values.domain }}
  rules:
    - name: home        
      from: /home
      to: https://lasuite.numerique.gouv.fr/produits/docs
    - name: visio        
      from: /visio
      to: https://lasuite.numerique.gouv.fr/produits/visio
    - name: error-404        
      from: /404
      to: https://lasuite.numerique.gouv.fr/404
      code: 404
```


src/helm/impress/templates/ingress-redirects.yaml
```diff
@@ -0,0 +1,49 @@
...
+kind: Ingress
+metadata:
+  name: {{ $name }}
+  namespace: {{ $ns }}
+  annotations:
+    {{- if or (not $r.code) (eq (toString $r.code) "301") }}
+    nginx.ingress.kubernetes.io/permanent-redirect: "{{ $to }}"
+    {{- else }}
+    nginx.ingress.kubernetes.io/configuration-snippet: |
+      return {{ $r.code }} {{ $to }};
```
lunika: Why not using `nginx.ingress.kubernetes.io/temporal-redirect` annotation ? We would like to avoid usage of `configuration-snippet` annotation

rouja: I agree with @lunika 

Image

You should manage the return code with the specific annotation instead of using configuration-snippet

src/helm/impress/templates/ingress-redirects.yaml
```diff
@@ -0,0 +1,49 @@
...
+            port:
+              number: {{ $.Values.frontend.service.port }}
+      - path: {{ printf "%s/" (trimSuffix "/" $from) }}
+        pathType: Exact
+        backend:
+          service:
+            name: {{ include "impress.frontend.fullname" $ }}
+            port:
+              number: {{ $.Values.frontend.service.port }}
+---
```
rouja: I prefer when the TLS configuration is explicit, because right now HTTPS is working thanks to other ingress.

AntoLC: Have a look in the fixup commits if it seems ok. Feel free to improve if it needs. 🙏

src/helm/impress/values.yaml
```diff
@@ -85,6 +85,16 @@ ingressCollaborationWS:
     nginx.ingress.kubernetes.io/proxy-send-timeout: "86400"
     nginx.ingress.kubernetes.io/upstream-hash-by: $arg_room
 
+ingressRedirects:
```
lunika, line 91: You have to add the documentation before thisblock (take a look at the other) and then run 
```
$ cd src/helm/impress
$ ./generate-readme.sh
```

AntoLC:
The  commit (`🐛(helm) fix OIDC authentication with standard scopes`) is necessary to test the feature but I will not merge it in this PR, but from this one instead #1740.
image

lunika (approved):

PR #1707, opened 2025-12-08
✨(backend) allow to create a new user in a marketing system

### LINKED ISSUE #1464: Add optional marketing automation capabilities
## Feature Request

As Docs instance manager I might need to add users into a mailing list so I can let them know when there is a new version or run onboarding sequences so that it's easier for them to get to know the tool 

Much like @lebaudantoine  did on Meet, we could add a [marketing.py](https://github.com/suitenumerique/meet/blob/main/src/backend/core/services/marketing.py) to our backend services. Which could be [called upon signup](https://github.com/suitenumerique/meet/blob/469e824167799a7ba6f1620398288fc65aa9d26f/src/backend/core/authentication/backends.py#L63) so that we add the new user to the contact list. 
Of course not all Docs instance admin will use the same mailing service or want to use this so we can create an env variable and a [Django Setting](https://github.com/suitenumerique/meet/blob/469e824167799a7ba6f1620398288fc65aa9d26f/src/backend/core/authentication/backends.py#L63) to configure the mailing service.

If you can, explain how users will be able to use this and possibly write out some documentation (if applicable).
Maybe add a screenshot or design?

lunika:
## Purpose

We want to create a new user in a marketing system to create a dedicated
onboarding for each of them. The marketing service is implemented in the
django-lasuite library and it is possible to pick the backend we want
or implement a new one following the documentation on this library.

## Proposal

- [x] ⬆️(backend) upgrade django-lasuite to version 0.0.22
- [x] ✨(backend) allow to create a new user in a marketing system

Fixes #1464


src/backend/impress/settings.py
```diff
@@ -808,6 +809,30 @@ class Base(Configuration):
...
+            "lasuite.marketing.backends.dummy.DummyBackend",
+            environ_name="LASUITE_MARKETING_BACKEND",
+            environ_prefix=None,
+        ),
+        "PARAMETERS": values.DictValue(
+            default={},
+            environ_name="LASUITE_MARKETING_PARAMETERS",
+            environ_prefix=None,
+        ),
+    }
```
AntoLC, line 834: I don't see any references in the tests about `LASUITE_MARKETING`, should we add a doc about how it is working   ?

src/backend/impress/settings.py
```diff
@@ -808,6 +809,30 @@ class Base(Configuration):
         ),
     }
 
+    # Marketing and communication settings
+    SIGNUP_NEW_USER_TO_MARKETING_EMAIL = values.BooleanValue(
```
AntoLC, line 813: We should add them in https://github.com/suitenumerique/docs/blob/main/docs/env.md.

lunika:
> Don't you need a management command to add existing users from admin?

Good question, @virgile-dev :arrow_up: ?

PR #1708, opened 2025-12-08
✨(backend) manage reconciliation requests for user accounts

### LINKED ISSUE #1616: Handle users with multiple accounts
## Feature Request

On our instance docs.numerique.gouv.fr people with multiple email adresses have created multiple accounts on Docs.
This leads to not knowing which user to invite on the docs has we have multiple results for one individual and also we for the user not knowing on which account a document has been shared and having to ask "can you invite me on this address instead?

While this is an issue created by the user and the identity provider it would be nice to have an internal tool that help handles this situation in the short term. 

When a user contacts the admin telling him he'd like to have his account(s) deleted / disabled in favor of another account he's actually actively using we should be able to run a task in the db where : 
- **owner** the document he created change ownership to the account he's using
- **member** his memberships and associated roles to documents he hasn't created get attributed to the account he's actively using -
- **access request** : a user can request access to a doc, we should update this access request with the account the user is actively using

When need to think about the emails the application sends, we don't want to flood the user with invitation emails.
So I'd be in favour of sending a digest instead. Recaping all the documents the user access after running the procedure.

We also need to think of who can run this task, only a sysadmin or someone with a django admin access which would mean we need to do UI for this. I'm in favour of doing the first option at first.

If you can, explain how users will be able to use this and possibly write out some documentation (if applicable).
Maybe add a screenshot or design?

### LINKED ISSUE #1791: Add a front route for user reconciliations
## Feature Request

Required for #1708 : when a `UserReconciliation` is created, two confirmation emails are sent per email pair (active and inactive), each with a link to validate that the user controls the address.

When a user clicks a link with the following format: `https://<root_url>/user_reconciliations/<user_type>/<confirmation_id>`, the equivalent backend URL in the API is called (`https://<root_url>/api/<api_version>/user_reconciliations/<user_type>/<confirmation_id>`), which processes the request in the background.

The user is then redirected to a simple confirmation page telling if everything is alright.

The value of `<user_type>` can be either `active` or `inactive`.
The value of `<confirmation_id>` is an UUID.

The confirmation email:

Image

Ash-Crow:
## Purpose
Fix #1616.

For now, the reconciliation requests are imported through CSV in the Django admin, which sends confirmation email to both addresses. When both are checked, the actual reconciliation is processed, in a threefold process (update document acess, update invitations, update user status.)

## Proposal
- [x] New `UserReconciliationCsvImport` model to manage the import of reconciliation requests through a task (`user_reconciliation_csv_import_job`)
- [x] New `UserReconciliation` model to store the user reconciliation requests themselves (a row = a `active_user`/`inactive_user` pair)
  - [x] On save, a confirmation email is sent to the users
- [x] A `process_reconciliation` admin action process the action on the requested entries, if both emails have been checked.
  - [x] Bulk update the `DocumentAccess` items, while managing the case where both users have access to the document (keeping the higher role)
  - [x] Bulk update the `LinkTrace` items, while managing the case where both users have link traces to the document
  - [x] Bulk update the `DocumentFavorite` items, while managing the case where both users have put the document in their favorites
  - [x] Bulk update the comment system items (`Thread`, `Comment` and `Reaction` items)
  - [x] Bulk update the `is_active` status on both users
- [x] New `USER_RECONCILIATION_FORM_URL` env variable for the "make a new request" URL in an email.
- [x] Write unit tests

## Demo page reconciliation success

In waiting for @rl-83 to provide the design:
image


src/backend/core/tasks/user_reconciliation.py
```diff
@@ -0,0 +1,98 @@
...
+                        rec_entry = UserReconciliation.objects.create(
+                            active_email=active_email,
+                            inactive_email=inactive_email,
+                            active_email_checked=active_email_checked,
+                            inactive_email_checked=inactive_email_checked,
+                            active_confirmation_id=uuid.uuid4(),
+                            inactive_confirmation_id=uuid.uuid4(),
+                            status="pending",
+                        )
+                        rec_entry.save()
```
lunika:
```suggestion
```
I think it is not needed. If it's to take advantage of the `created_at` and `updated_at` auto fill thent you should probably not use the `create` function but instead instantiate the model and then save it later.

src/backend/core/tasks/user_reconciliation.py
```diff
@@ -0,0 +1,98 @@
...
+    job = UserReconciliationCsvImport.objects.get(id=job_id)
+    job.status = "running"
+    job.save()
+
+    try:
+        with job.file.open(mode="r") as f:
+            reader = csv.DictReader(f)
+            rec_entries_created = 0
+            for row in reader:
+                status = row["status"]
```
lunika: Maybe I'm wrong, but this column `status` is useless, don't you think? You can accept only one value. Or perhaps I'm not understanding its purpose.

src/backend/core/models.py
```diff
@@ -265,6 +297,246 @@ def teams(self):
...
+                    existing_access.role = max_role
+                    updated_accesses.append(existing_access)
+                removed_accesses.append(entry)
+            else:
+                entry.user = self.active_user
+                updated_accesses.append(entry)
+
+        self.logs += f"""Requested update for {len(updated_accesses)} DocumentAccess items
+            and deletion for {len(removed_accesses)} DocumentAccess items.\n"""
+        self.status = "done"
```
lunika: Changing the status can lead to an unwanted state if something wrong happens between when this function is executed and when the bulk update is made.

src/backend/core/models.py
```diff
@@ -265,6 +297,246 @@ def teams(self):
...
+        related_name="active_user",
+    )
+    inactive_user = models.ForeignKey(
+        User,
+        on_delete=models.CASCADE,
+        null=True,
+        blank=True,
+        related_name="inactive_user",
+    )
+    active_confirmation_id = models.UUIDField(
```
lunika: It's a little bit long, but `active_email_confirmation_id` gives probably more information.

src/backend/core/models.py
```diff
@@ -265,6 +297,246 @@ def teams(self):
...
+        default=uuid.uuid4, unique=True, editable=False, null=True
+    )
+    inactive_confirmation_id = models.UUIDField(
```
lunika: Same here

src/backend/locale/br_FR/LC_MESSAGES/django.po
AntoLC, line 1: I can see you get lot of conflicts, translations push and pull are usually managed automatically, during release.

src/backend/core/models.py
```diff
@@ -265,6 +297,246 @@ def teams(self):
...
+                            - {email2}
+
+                            Please check for typos.
+
+                            You can submit another request with the valid email addresses.
+                             """
+                ).format(email1=email_1, email2=email_2),
+                "link": f"{domain}/",
+                "link_label": str(_("Click here")),
+                "button_label": str(_("Make a new request")),
```
AntoLC: I can see there is a button to trigger a new request, is there a route for that ?

src/backend/core/api/viewsets.py
```diff
@@ -247,6 +249,59 @@ def get_me(self, request):
         )
 
 
+class ReconciliationConfirmView(APIView):
+    """API endpoint to confirm user reconciliation emails.
+
+    GET /user_reconciliations/{user_type}/{confirmation_id}/
+    Marks `active_email_checked` or `inactive_email_checked` to True.
+    """
```
AntoLC, line 259: Could be nice to have a few tests about this part, to have like a blueprint on how to use it.

src/backend/core/models.py
```diff
@@ -265,6 +297,246 @@ def teams(self):
...
+            ("done", _("Done")),
+            ("error", _("Error")),
+        ],
+        default="pending",
+    )
+    logs = models.TextField(blank=True)
+
+    class Meta:
+        verbose_name = _("user reconciliation")
+        verbose_name_plural = _("user reconciliations")
```
AntoLC, line 349:
```suggestion
    class Meta:
        db_table = "impress_user_reconciliation"
        verbose_name = _("user reconciliation")
        verbose_name_plural = _("user reconciliations")
```

By default the table will be created with `core` prefix otherwise.

Image

src/backend/core/models.py
```diff
@@ -265,6 +297,246 @@ def teams(self):
...
+            context = {
+                "title": subject,
+                "message_bold": _(
+                    "You have requested a reconciliation of your user accounts on Docs."
+                ),
+                "message": _(
+                    """To confirm that you are the one who initiated the request
+                    and that this email belongs to you:"""
+                ),
+                "link": f"{domain}/user_reconciliations/{user_type}/{confirmation_id}/",
```
AntoLC: Nitpick, but could we change to hyphens ?

```suggestion
                "link": f"{domain}/user-reconciliations/{user_type}/{confirmation_id}/",
```

lunika:
After discussion with @AntoLC some data migration are missing.
Other relations with the user exists in the database and this data will be lost if not transfered.
When a user visit a docs it doesn't have access (a public or connected one), an entry is added to the `LinkTrace` model. These docs will be shown to the user on its homepage. This table should also be migrated.
Same for the `DocumentFavorite` model. Should be migrated.

There are also relation with `Thread`, `Comment` and `Reaction`. But I don't know if they should be migrated. WDYT @virgile-dev @rl-83 ? A thread (the comment container) and comments should be migrated from the inactive user to the new one ?

src/backend/core/tasks/user_reconciliation.py
Ash-Crow: This is the status on the source Grist form, as the CSV generation exports the whole table, so the lines that were already processed are ignored.

Capture d’écran du 2026-01-20 15-14-29

That said, I just saw how I can export the unique ID Grist gives to each row, so I'll update the model to use that instead

src/backend/locale/br_FR/LC_MESSAGES/django.po
Ash-Crow: Thanks for the info, I removed the translations from the commits

src/backend/core/tasks/user_reconciliation.py
Ash-Crow: I made the change.

src/backend/core/models.py
Ash-Crow: It is an URL to a third party form (Grist in our case). I used an ENV variable to store it.

src/backend/core/models.py
Ash-Crow: Done

src/backend/core/models.py
Ash-Crow: I put the reconciliation requests in an atomic transaction, this should fix the issue.

Ash-Crow:
> After discussion with @AntoLC some data migration are missing. Other relations with the user exists in the database and this data will be lost if not transfered. When a user visit a docs it doesn't have access (a public or connected one), an entry is added to the `LinkTrace` model. These docs will be shown to the user on its homepage. This table should also be migrated. Same for the `DocumentFavorite` model. Should be migrated.

Done.
> 
> There are also relation with `Thread`, `Comment` and `Reaction`. But I don't know if they should be migrated. WDYT @virgile-dev @rl-83 ? A thread (the comment container) and comments should be migrated from the inactive user to the new one ?

Done as well, if not needed we can put it behind a toggle controlled by an env variable.

src/backend/core/api/viewsets.py
```diff
@@ -40,6 +40,8 @@
 from rest_framework import filters, status, viewsets
 from rest_framework import response as drf_response
 from rest_framework.permissions import AllowAny
+from rest_framework.response import Response
```
lunika, line 43: line 41 the drf reponse module is imported, you can probably use it instead of making an other import ?

src/backend/core/models.py
```diff
@@ -251,6 +251,37 @@ def _convert_valid_invitations(self):
 
         valid_invitations.delete()
 
+    def send_email(self, subject, context=None, language=None):
+        """Generate and send email to the user from a template."""
+        emails = [self.email]
+        context = context or {}
+        domain = Site.objects.get_current().domain
```
lunika, line 258: You should do like in this PR : https://github.com/suitenumerique/docs/pull/1825

lunika (approved):
Few nitepicking otherwise it's good to me

CHANGELOG.md
```diff
@@ -71,6 +71,7 @@ and this project adheres to
 - ✨(helm) redirecting system #1697
 - 📱(frontend) add comments for smaller device #1737
 - ✨(project) add custom js support via config #1759
+- ✨(backend) manage reconciliation requests for user accounts #1708
```
AntoLC, line 74: Should be placed under "Unreleased"

lunika (changes requested):
Didn't saw you have merge commits in your commits history.
You have to remove them

PR #1714, opened 2025-12-09
💄(export) add style file to html export

AntoLC:
## Purpose

We now include a CSS style file in the exported ZIP archive. This file contains styles that enhance the appearance of the exported HTML document when viewed in a web browser to look more like the original document.

## Proposal

[test-html-export.zip](https://github.com/user-attachments/files/24051894/test-html-export.zip)

https://github.com/user-attachments/assets/06cc8b22-1642-4679-8819-5d9b9ae2a979


Ovgodd (approved):
really nice improvement

PR #1730, opened 2025-12-12
🛂(backend) stop throttling collaboration servers

AntoLC:
## Purpose

We observe some throttling pick here and there.
We observed that when the collaboration has a problem, it is retrying to connect, leading to more requests to the django backend. At one point, the throttling is reached and the user would not be able to use the application anymore.

## Proposal

Now when the request comes from a collaboration server, we do not throttle it anymore.

To do so we add the `Y_PROVIDER_API_KEY` in the request coming from the collaboration server.
```ts
const response = await axios.get<T>(
    `${COLLABORATION_BACKEND_BASE_URL}${path}`,
    {
      headers: {
        cookie: requestHeaders['cookie'],
        origin: requestHeaders['origin'],
        'X-Y-Provider-Key': Y_PROVIDER_API_KEY,
      },
    },
  );
  ```

### DIFF
--- a/src/backend/core/tests/test_api_throttling_document_throttle.py
+++ b/src/backend/core/tests/test_api_throttling_document_throttle.py
@@ -1,0 +1,107 @@
+"""
+Test DocumentThrottle for regular throttling and y-provider bypass.
+"""
+
+import pytest
+from rest_framework.test import APIClient
+
+from core import factories
+
+pytestmark = pytest.mark.django_db
+
+
+def test_api_throttling_document_throttle_regular_requests(settings):
+    """Test that regular requests are throttled normally."""
+
+    current_rate = settings.REST_FRAMEWORK["DEFAULT_THROTTLE_RATES"]["document"]
+    settings.REST_FRAMEWORK["DEFAULT_THROTTLE_RATES"]["document"] = "3/minute"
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: src/backend/core/tests/test_api_throttling_document_throttle.py [hunk 1]
```diff
@@ -0,0 +1,100 @@
...
+
+    user = factories.UserFactory()
+    client = APIClient()
+    client.force_login(user)
+
+    document = factories.DocumentFactory()
+    factories.UserDocumentAccessFactory(document=document, user=user)
+
+    # Make many requests with the y-provider API key
+    for _i in range(100):
```
lunika: Maybe you can test with 10 ? To save time in the test suite

src/backend/core/tests/test_api_throttling_document_throttle.py
```diff
@@ -0,0 +1,100 @@
...
+            f"/api/v1.0/documents/{document.id!s}/",
+        )
+        assert response.status_code == 200
+
+    # 4th request should be throttled
+    response = client.get(
+        f"/api/v1.0/documents/{document.id!s}/",
+    )
+    assert response.status_code == 429
+
```
lunika, line 39: You can maybe here make a new request using the special header to test that this one will not be throttled ?

Issue #1680, opened 2025-12-02
⚡️Comments feature improvements


AntoLC:
## Bug Report

- [x] When we add a comments we have to scroll to the bottom of the thread to see it, add a automatic scroll down [#1687](https://github.com/suitenumerique/docs/pull/1687)
- [x] When we add a comments, add automatically the focus in the input [#1687](https://github.com/suitenumerique/docs/pull/1687)
- [x] Accesibility color issue on overline text to comment: [#1687](https://github.com/suitenumerique/docs/pull/1687)
<img width="207" height="134" alt="Image" src="https://github.com/user-attachments/assets/a73ef532-e811-469c-96ed-7eb23b2ebb2c" />

- [x] The comment button is displayed in the toolbar in mobile even though it's not possible to comment on mobile [#1687](https://github.com/suitenumerique/docs/pull/1687)
- [x] Deleting a comment doesn't remove background highlight [#1687](https://github.com/suitenumerique/docs/pull/1687)

https://github.com/user-attachments/assets/be2c5dd1-2300-4f10-95d8-4c38dd438cff

- [ ] When we write a comment, "enter" add a break line, add "ctrl+enter" to trigger the comment submit 
       _(Doing "Tab then enter" works fine, if it is not enough we have to open a issue BN side)_
- [ ] We have to do a double click to open the emoji picker (under the text comment)
- [ ] During collaboration, if someone write/type when we try to do a initial comment (first modal) , the modal close, only if the typing is previous of the message
       _(Next BN release has maybe a patch about it)_
- [x] Color of the submit button doesn't seem right it's not coherent with the rest of the UI #1732

<img width="592" height="162" alt="Image" src="https://github.com/user-attachments/assets/6bf5948c-ed45-49ca-a1d9-9da117d23e2f" />

- [x] when you copy paste a comment from a doc to another Docs, if you click on the copied text you get an application error #1732
This PR fixes partially the problem as well, no crash anymore: https://github.com/suitenumerique/docs/pull/1733

https://github.com/user-attachments/assets/903cd056-5ed2-4735-ae5d-93b65e352f9a

- [ ] when you click on a comments that was made on a link you can't open it as it opens the link in a new tab

bzg:
+1 on "add "ctrl+enter" to trigger the comment submit", it's more expected than "TAB then RET".

Wqrld:
Tangentially related: Someone on matrix requested the ability to @mention someone in comments and notify them by email.

[...]

PR #1732, opened 2025-12-12
🐛(frontend) paste content with comments from another document

AntoLC:
## Purpose

See: #1680

When pasting comments, the `data-bn-thread-id` attribute is present in the clipboard data.
This indicates that the pasted content contains comments. But if the content with comments comes from another document, it will create orphaned comments that are not linked to this document and create errors. 

## Proposal

To avoid this, when a comment is pasted, we refresh the threads to ensure that only comments relevant to the current document are displayed.


AntoLC:
This #1733 save partially the problem as well, no crash anymore.

Ovgodd (approved):
Looks good !

PR #1733, opened 2025-12-12
⬆️(dependencies) bump blocknote to 0.44.2

AntoLC:
## Purpose

We bump the blocknote dependencies to version 0.44.2 to incorporate the latest features and bug fixes.
It seems to fix an issue with Titap, when the text was selected and the user clicked on the Go Back button of the browser, the application was crashing.
```
"[tiptap error]: The editor view is not available. Cannot access view['dom']. The editor may not be
mounted yet."
```

### DIFF
--- a/src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-blocks/CalloutBlock.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-blocks/CalloutBlock.tsx
@@ -97,34 +97,41 @@ const CalloutComponent = ({
       `}
     >
       <CalloutBlockStyle />
-      <BoxButton
-        contentEditable={false}
-        onClick={toggleEmojiPicker}
-        $css={css`
-          font-size: 1.125rem;
-          cursor: ${isEditable ? 'pointer' : 'default'};
-          ${isEditable &&
-          `
+      <Box $position="relative">
+        <BoxButton
+          contentEditable={false}
+          onClick={toggleEmojiPicker}
+          onKeyDown={(e) => {
+            if (e.key === 'Escape' && openEmojiPicker) {
+              setOpenEmojiPicker(false);
+            }
+          }}
+          $css={css`
+            font-size: 1.125rem;
+            cursor: ${isEditable ? 'pointer' : 'default'};
+            ${isEditable &&
+            `
           &:hover {
             background-color: rgba(0, 0, 0, 0.1);
           }
           `}
-        `}
-        $align="center"
-        $width="28px"
-        $radius="4px"
-      >
-        {block.props.emoji}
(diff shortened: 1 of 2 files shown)


src/frontend/apps/impress/src/features/docs/doc-editor/components/comments/DocsThreadStore.tsx
```diff
@@ -200,7 +200,7 @@ export class DocsThreadStore extends ThreadStore {
     const { editor } = useEditorStore.getState();
 
     // Should not happen
-    if (!editor) {
+    if (!editor || !editor._tiptapEditor?.view?.dom) {
```
Ovgodd, line 203: nit : 
```suggestion
const isEditorReady =  Boolean(editor) &&
  Boolean(editor._tiptapEditor?.view?.dom);
    if (!isEditorReady ) {
```

src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-blocks/CalloutBlock.tsx
```diff
@@ -97,34 +97,41 @@ const CalloutComponent = ({
...
-          `
+      <Box $position="relative">
+        <BoxButton
+          contentEditable={false}
+          onClick={toggleEmojiPicker}
+          onKeyDown={(e) => {
+            if (e.key === 'Escape' && openEmojiPicker) {
+              setOpenEmojiPicker(false);
+            }
+          }}
```
Ovgodd, line 108: Might be worth extracting this inline handler to keep the JSX more readable.maybe like this : 
`const handleEmojiButtonKeyDown = (
  event: React.KeyboardEvent<HTMLElement>
) => {
  if (event.key === 'Escape' && openEmojiPicker) {
    event.stopPropagation();
    setOpenEmojiPicker(false);
  }
};
`

Ovgodd (approved):
good for me ( just suggested 2 changes )

src/frontend/apps/impress/src/features/docs/doc-editor/components/comments/DocsThreadStore.tsx
AntoLC: It helps the linter to understand that editor is set, if we do as suggested, you get lint error:
image

src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-blocks/CalloutBlock.tsx
AntoLC: For big chunk of code I agree it is better, but for small snippet I find it is easier to have it directly on the component, I don't have to search where the function is, I can see directly what it is doing like that.

PR #1754, opened 2025-12-22
🥅(frontend) intercept 401 error on GET threads

AntoLC:
## Purpose

We intercept 401 errors on GET /threads to avoid spamming Sentry with authentication errors when users are not logged in.


Ovgodd (approved):
👍

PR #1755, opened 2025-12-22
✈️(frontend) pause Posthog when offline

AntoLC:
## Purpose

Posthog keeps trying to send events when the user is offline, causing the network request queue to fill up and slowing down the app. We now pauses Posthog when the user is offline and resumes it when back online.


Ovgodd (approved):
💯

PR #1768, opened 2026-01-05
🔒️(backend) validate more strictly url used by cors-proxy endpoint

lunika:
## Purpose

The cors-proxy endpoint allow to download images host externally without
being blocked by cors headers. The response is filter on the return
content-type to avoid disclosure and the usage of this endpoint as the
proxy used by attacker. We want to restrict the usage of this endpoint
by filtering on non legit ips used. This filter avoid exploitation of
Server Side Request Forgery (SSRF).

## Proposal

- [x] 🔒️(backend) validate more strictly url used by cors-proxy endpoint
- [x] ♻️(backend) stop returning a 415 on cors-proxy endpoint
- [x] ♻️(backend) stop allowing redirect in cors-proxy endpoint

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -1715,7 +1715,6 @@ class DocumentViewSet(
         if not hostname:
             raise drf.exceptions.ValidationError("Invalid hostname")
 
-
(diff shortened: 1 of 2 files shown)


AntoLC (approved):
LGTM

PR #1776, opened 2026-01-08
✨(backend) use langfuse to monitor AI actions

### LINKED ISSUE #1672: Add LLM observability and tracing
### **Summary**

Docs includes an integrated LLM to assist users while editing. Some users report that responses are slow, inaccurate, or occasionally include out-of-context elements. We currently lack detailed visibility into how the LLM behaves in production, which limits our ability to diagnose and improve the experience. I propose adding LLM observability (e.g., via Langfuse) to trace, analyze, and debug all LLM interactions.

### **Problem**

We have no structured observability on LLM usage in Docs.
As a result, we cannot easily identify:

* why some responses are slow,
* which prompts lead to low-quality or hallucinated output,
* how tools are being called internally,
* or whether issues correlate with specific users, documents, or workflows.

This makes debugging slow and prevents data-driven improvements.

### **Proposed Solution**

Integrate an LLM observability tool—such as **Langfuse**—to trace all LLM interactions.
This should include:

* Logging the **input prompt** and any system/internal transformations
* Logging the **model output**, latency, and token usage
* Logging all **tool calls** with inputs/outputs
* Linking each trace to the **User ID** for cross-referencing

This will allow us to diagnose issues, understand failure patterns, and iterate on the LLM feature with proper data.

lunika:
## Purpose

We want to monitor AI actions. For this we choose to use langfuse. As this usage is optional, we load langfuse sdk only if settings are configured. Also, the openai client from langfuse is a dropin replacement of openai client, so we only have to change how openai is imported.

## Proposal

- [x] ✨(backend) use langfuse to monitor AI actions

Fixes #1672


src/backend/impress/settings.py
```diff
@@ -699,6 +699,16 @@ class Base(Configuration):
         "day": 200,
     }
 
+    LANGFUSE_SECRET_KEY = SecretFileValue(
+        None, environ_name="LANGFUSE_SECRET_KEY", environ_prefix=None
+    )
+    LANGFUSE_PUBLIC_KEY = values.Value(
+        None, environ_name="LANGFUSE_PUBLIC_KEY", environ_prefix=None
+    )
+    LANGFUSE_BASE_URL = values.Value(
```
AntoLC, line 708: The env should be added in https://github.com/suitenumerique/docs/blob/main/docs/env.md.

PR #1787, opened 2026-01-11
🚸(frontend) remove blocking modal on save in Firefox

### LINKED ISSUE #1783: Improve document saving behaviour with Firefox
## Feature Request

If you type in a doc and try to close the tab you'll see a browser modal like this

Image

This is not what we're selling to users. They to collaborate in real time and not worry about saving at all.

We should improve document save so users don't see this modal. On Google Docs or Notion you close without worrying ever.
Ideally we dont want users to click on a save icon, that would be lame for a product like Docs.

You can do 2 things. 
1. Wait a min until doc is saved
2. Click on the parent document in the left panel to trigger a save. Hacky but works. 

Please tech friends, could you find a clever solution to this 🙏 

AntoLC:
## Purpose

In Firefox the request to save documents did not complete when the user where closing is tab or browser. To let the save request complete we added a blocking modal, but it was not user friendly.
Now we removed the blocking modal and add "keepalive" option to the fetch request to let the request complete in background when the user close the tab or browser.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,19 +14,20 @@ and this project adheres to
 - 🔒(helm) Set default security context #1750
 - ✨(backend) use langfuse to monitor AI actions
 
-### Fixed 
-
-- ✅(backend) reduce flakiness on backend test #1769
-- 🐛(backend) fix TRASHBIN_CUTOFF_DAYS type error #1778
-
-### Security
-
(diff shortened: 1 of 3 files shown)


AntoLC:
Not working as expected on big documents.

Plus yhub server should solve this problem: https://github.com/suitenumerique/docs/pull/2556

PR #1792, opened 2026-01-13
🔒️(frontend) fix props vulnerability in Interlinking

AntoLC:
## Purpose

We were not properly sanitizing props passed to the `InterlinkingLinkInlineContent` component, which could lead to XSS attacks. This commit remove most of the props and only keep the necessary ones.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,13 @@ and this project adheres to
 - ✅(export) add PDF regression tests #1762
 - 📝(docs) Add language configuration documentation #1757
 - 🔒(helm) Set default security context #1750
-- ✨(backend) use langfuse to monitor AI actions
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -24,12 +30,7 @@ and this project adheres to
 ### Security
 
 - 🔒️(backend) validate more strictly url used by cors-proxy endpoint #1768
-
-### Changed
-
-- ♿(frontend) improve accessibility:
-  - ♿(frontend) make html export accessible to screen reader users #1743
-  - ♿(frontend) add missing label and fix Axes errors to improve a11y #1693
+- 🔒️(frontend) fix props vulnerability in Interlinking #1792
```
lunika, line 33: You should move it in the `Security` section

AntoLC: It is

PR #1801, opened 2026-01-14
📌(backend) pin celery to version<5.6.0

lunika:
## Purpose

Since celery version 5.6.0 we have trouble with retrying tasks and it is impactig the malware_detection workflow. We have to use version 5.5.3 while we found the issue.

## Proposal

- [x] 📌(backend) pin celery to version<5.6.0


lunika:
I don't think.

Issue #1521, opened 2025-10-24
Sharing - User search - Order by proximity to current user


olaurendeau:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
On the docs.numerique.gouv.fr instance, there is now a lot of users available in the search view. 

When searching for a user to share a doc with, the first results brought by a generic request, for example "Pierre", retrieves **people that are not related to the current user**, by far ! So you then have to enter almost the full name of the user you're looking for, which make the autocomplete feature less convenient.

**Describe the solution you'd like**
Ordering the results of the user autocomplete by proximity to the current user would solve the issue for me. 

For example : 
```
Given my user email is olivier.laurendeau@anct.gouv.fr 
And I have recently shared a document with paul.dupond@beta.gouv.fr
And I have less recently shared a document with paul.durand@impots.gouv.fr
And the users paul.destre@tutu.fr, paul.dru@anct.gouv.fr & paul.dro@culture.gouv.fr also exists
When I search a user in the sharing box with "paul.d", I should see this list in this order :
# Shared with first 
- paul.dupond@beta.gouv.fr # Most recent first
- paul.durand@impots.gouv.fr
# Same full domain second
- paul.dru@anct.gouv.fr
# Same partial domain third
- paul.dro@culture.gouv.fr
- paul.destre@tutu.fr
```

**Do you want to work on it through a Pull Request?**
Maybe with some help :)

PR #1802, opened 2026-01-14
🚸(backend) sort user search results by proximity with the active user

Ash-Crow:
## Purpose
Allows a user to find more easily the other users they search, with the following order of priority:
- users they already share documents with (more recent first)
- users that share the same full email domain
- ~~users that share the same partial email domain (last two parts)~~
- ~~other users~~

Edit: We need to ilter out other users in order to not reveal email addresses from members of other organisations. It's still possible to invite them by email.

Solves #1521

## Proposal
- [x] Add a new function in `core/utils.py`: `users_sharing_documents_with()`
- [x] Use it as a key to sort the results of a basic user search
- [x] Filter user results to avoid reveal of users (and email addresses) of other orgs or that have not been interacted with.
- [x] User research through "full" email address (contains the '@') is left unaffected.

### DIFF
--- a/src/backend/core/tests/test_api_users.py
+++ b/src/backend/core/tests/test_api_users.py
@@ -2,7 +2,6 @@
 Test users API endpoints in the impress core app.
 """
 
-from django.test import override_settings
 from django.utils import timezone
 
 import pytest
@@ -380,7 +379,7 @@ def test_api_users_list_query_email_exclude_doc_user():
     assert user_ids == [str(nicole_fool.id)]
 
 
-def test_api_users_list_query_short_queries_default_length():
+def test_api_users_list_query_short_queries():
     """
     If API_USERS_SEARCH_QUERY_MIN_LENGTH is not set, the default minimum length should be 3.
     """
@@ -392,36 +391,6 @@ def test_api_users_list_query_short_queries_default_length():
     factories.UserFactory(email="john.lennon@example.com", full_name="John Lennon")
 
     response = client.get("/api/v1.0/users/?q=joh")
-    assert response.status_code == 200
-    assert len(response.json()) == 2
-
-
-@override_settings(API_USERS_SEARCH_QUERY_MIN_LENGTH=5)
-def test_api_users_list_query_short_queries_custom_length():
-    """
-    Queries shorter than API_USERS_SEARCH_QUERY_MIN_LENGTH
-    should return a 400 error with a validation message.
-    """
-    user = factories.UserFactory(email="paul@example.com", full_name="Paul")
-    client = APIClient()
-    client.force_login(user)
-
-    factories.UserFactory(email="john.doe@example.com", full_name="John Doe")
-    factories.UserFactory(email="john.lennon@example.com", full_name="John Lennon")
-
-    response = client.get("/api/v1.0/users/?q=jo")
-    assert response.status_code == 400
-    assert response.json() == {
-        "q": ["Ensure this value has at least 5 characters (it has 2)."]
-    }
-
-    response = client.get("/api/v1.0/users/?q=john")
-    assert response.status_code == 400
-    assert response.json() == {
-        "q": ["Ensure this value has at least 5 characters (it has 4)."]
-    }
-
-    response = client.get("/api/v1.0/users/?q=john.")
     assert response.status_code == 200
     assert len(response.json()) == 2
 


lunika (approved):
Great implementation

src/backend/core/utils.py
```diff
@@ -96,3 +100,39 @@ def extract_attachments(content):
 
     xml_content = base64_yjs_to_xml(content)
     return re.findall(enums.MEDIA_STORAGE_URL_EXTRACT, xml_content)
+
+
+def users_sharing_documents_with(user):
```
qbey, line 114: I'm a bit afraid of this, because on each search query, you will load all user accesses, I don't have a better proposition and maybe we should let it like that but we should at least add a "timer" log to be able to see if this takes too much time. Maybe a cache could also help?

src/backend/core/utils.py
```diff
@@ -96,3 +100,39 @@ def extract_attachments(content):
...
+    user_docs_qs = models.DocumentAccess.objects.filter(user=user).values_list(
+        "document_id", flat=True
+    )
+    shared_qs = (
+        models.DocumentAccess.objects.filter(document_id__in=user_docs_qs)
+        .exclude(user=user)
+        .values("user")
+        .annotate(last_shared=db.Max("created_at"))
+    )
+    return {item["user"]: item["last_shared"] for item in shared_qs}
```
qbey:
```suggestion
    return {item["user"]: item["last_shared"] for item in shared_qs.iterator()}
```

I would suggest to use an iterator here.

src/backend/core/utils.py
```diff
@@ -96,3 +100,39 @@ def extract_attachments(content):
...
+    The partial domain consists of the last two segments of the domain, eg. "gouv.fr".
+
+    If the email is invalid (eg, is an empty string), returns empty strings.
+    """
+    try:
+        validate_email(email)
+    except ValidationError:
+        return "", ""
+
+    domain = email.split("@", 1)[1].lower()
```
qbey: Such split is not "safe", the safe way is using `email.headerregistry.Address` but for performances sake, I guess we should keep it like this: maybe add a comment and rename the function `unsafe_extract_email_domain_parts` to say you did it this way on purpose could help, and might prevent the use of this method for another context :)

Note: django-lasuite already provides a `get_domain_from_email`

src/backend/core/utils.py
```diff
@@ -96,3 +100,39 @@ def extract_attachments(content):
...
+    """
+    Returns a map of users sharing documents with the given user,
+    sorted by last shared date.
+    """
+
+    user_docs_qs = models.DocumentAccess.objects.filter(user=user).values_list(
+        "document_id", flat=True
+    )
+    shared_qs = (
+        models.DocumentAccess.objects.filter(document_id__in=user_docs_qs)
```
qbey: Would using a Subquery be better here? (I mean, to prevent data from being passed to Django for nothing)

qbey (commented):
Interesting:) I like the python instead of SQL approach but I wonder if it will not bring too much overhead. I guess it could be ok for a first version but:
 - is the "closer" compute not too much? I mean if you put the user you share something with, wouldn't that be lighter to compute and enough? Could be a first step, before bringing the heavy machine gun.
 - we need to monitor the function execution time to see if at some point or for some users it starts to be long (either logger.info or prometheus)

Anyway, this is just a review to help (I know I only ask more questions and did not bring any solution ^^)

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ and this project adheres to
 
 - ✅(e2e) fix e2e test for other browsers #1799
 
+### Changed
+
+- 🚸(backend) sort user search results by proximity with the active user #1802
```
sampaccoud, line 16: Maybe create an issue now in django-lasuite so we don't forget to upstream it as it will have to be shared with all our apps once it has proven its efficiency?

Ash-Crow:
Per a talk with @virgile-dev, we will only keep results for the first two categories (shared doc with and same full email domain) and drop the others to avoid exposing email addresses, for GDPR reason.

src/backend/core/utils.py
Ash-Crow: I added both caching and a timer

src/backend/core/utils.py
Ash-Crow: I didn't know about django-lasuite's `get_domain_from_email()`! As we don't need to compare subdomains anymore, I switched to it and removed that method.

src/backend/core/utils.py
Ash-Crow: It seems to be about the same, but I added it nonetheless.

```
app-dev-1  | 2026-02-10 14:32:55,702 core.utils INFO users_sharing_documents_with cache miss for user 46ed053c-370c-445d-b405-cefebf7afb25 (took 0.007s)
```

CHANGELOG.md
Ash-Crow: Done

Ash-Crow:
@AntoLC → @virgile-dev asked that we lower the char limit for user search to 3. I made the change on the backend side, can you make it on the frontend as well?

src/backend/core/utils.py
```diff
@@ -96,3 +104,41 @@ def extract_attachments(content):
...
     return re.findall(enums.MEDIA_STORAGE_URL_EXTRACT, xml_content)
+
+
+def users_sharing_documents_with(user):
+    """
+    Returns a map of users sharing documents with the given user,
+    sorted by last shared date.
+    """
+    start_time = time.time()
+    cache_key = f"users_sharing_documents_with_{user.id}"
```
lunika: You should probaly create a function `get_users_sharing_documents_with_cache_key(user)` to have only one place where the cache key string is created to avoid error.

Ash-Crow: Done.

PR #1814, opened 2026-01-19
Mark mosa.cloud demo instance as unavailable

laem:
Updated the status of the mosa.cloud demo instance to indicate it is unavailable.

## Purpose

Describe the purpose of this pull request.


mosa-riel:
The instance is back again!

PR #1832, opened 2026-01-26
✨(frontend) Can print a doc

AntoLC:
## Purpose

If we want to print our docs, they are now displayed correctly.

## Demo

https://github.com/user-attachments/assets/363deced-99e1-4673-b4c8-a770ef0ed114


StephanMeijer:
Seems like a sensible approach. You lose a bit of control versus first converting to Typst and then to PDF, but it gives you a bit more control on the front-end

PR #1834, opened 2026-01-27
Integrate Find

mascarpon3:
## Purpose

integrate Find to Docs

## Proposal

- [x] ✨ add a `useSeachDocs` hook in charged of calling the search endpoint. 
- [x] ✨ add a optional `path` param to the `search` route. This param represents the parent document path in case of a sub-documents (descendants) search. 
- [x] ⚡️return Indexer results directly without DB calls to retrieve the Document objects. All informations necessary for display are indexed in Find. We can skip the DB calls and improve performance. 
- [x] ♻️ refactor react `DocSearchContent` components. `DocSearchContent` and `DocSearchSubContent` are now merged a unique component handling all search scenarios and relying on the unique `search` route.
- [x] 🔥remove pagination logic in the Indexer. Removing the DB calls also removes the DRF queryset object which handles the pagination. Also we consider pagination not to be necessary for search v1.  
- [x] 🔥remove the `document/<document_id>/descendants` route. This route is not used anymore. The logic of finding the descendants are moved to the internal `_list_descendants` method. This method is based on the parent `path` instead of the parent `id` which has some consequence about the user access management. Relying on the path prevents the use of the `self.get_object()` method which used to handle the user access logic. 
- [x] ✨handle fallback logic on DRF based title search in case of non-configured, badly configured or failing at run time indexer. 
- [x] ✨handle language extension in `title` field. Find returns titles with a language extension (ex: `{ title.fr: "rapport d'activité" }` instead of `{ "title": "rapport d'activité" }`.
- [x] 🔧 add a `common.test` file to allow running the tests without docker
- [x] ♻️ rename `SearchIndexer` -> `FindDocumentIndexer`. This class has to do with Find in particular and the convention is more coherent with `BaseDocumentIndexer`
- [x] ♻️ rename `SEARCH_INDEXER_URL` -> `INDEXING_URL` and `SEARCH_INDEXER_QUERY_URL` -> `SEARCH_URL`. I found the original names very confusing. 
- [x] 🔧 update the environment variables to activate the FindDocumentIndexer.
- [x] ✨automate the generation of encryption key during bootstrap. OIDC_STORE_REFRESH_TOKEN_KEY is a mandatory secret key. We can not push it on Github and we want any contributor to be able to run the app by only running the `make bootstrap`. We chose to generate and wright it into the `common.local` during bootstrap. 

## External contributions


mascarpon3:
@AntoLC 

Effectivement tu dois ajouter une variable d'env `OIDC_STORE_REFRESH_TOKEN_KEY` dans ton `common.local` qui soit une clé Fernet. C'est utilisé pour crypter les OIDC refresh et access token. C'est déjà documenté mais j'ai amélioré un peu le `search.md`. L'utilitaire `bin/fernetkey` te génère une clé. 

J'ai aussi ajouté une gestion des erreur de l'api de Find se qui inclue le cas ou il n'est pas sur le réseau. J'avais géré le cas ou il est pas ou mal configuré mais pas celui ou on a une erreur au run time.

bin/generate-oidc-store-refresh-token-key.sh
```diff
@@ -0,0 +1,12 @@
...
+
+# Generate the secret OIDC_STORE_REFRESH_TOKEN_KEY and store it to common.local
+
+set -eo pipefail
+
+COMMON_LOCAL="env.d/development/common.local"
+
+OIDC_STORE_REFRESH_TOKEN_KEY=$(openssl rand -base64 32)
+
+echo "OIDC_STORE_REFRESH_TOKEN_KEY=${OIDC_STORE_REFRESH_TOKEN_KEY}" > "${COMMON_LOCAL}"
```
lunika:
```suggestion
echo "OIDC_STORE_REFRESH_TOKEN_KEY=${OIDC_STORE_REFRESH_TOKEN_KEY}" >> "${COMMON_LOCAL}"
```

Otherwise you will erase all the file content to put the output of the `echo`

env.d/development/common
```diff
@@ -85,8 +78,9 @@ DOCSPEC_API_URL=http://docspec:4000/conversion
...
-# Indexer (disabled)
-# SEARCH_INDEXER_CLASS="core.services.search_indexers.SearchIndexer"
+# Indexer
+SEARCH_INDEXER_CLASS=core.services.search_indexers.FindDocumentIndexer
 SEARCH_INDEXER_SECRET=find-api-key-for-docs-with-exactly-50-chars-length  # Key generated by create_demo in Find app.
-SEARCH_INDEXER_URL="http://find:8000/api/v1.0/documents/index/"
-SEARCH_INDEXER_QUERY_URL="http://find:8000/api/v1.0/documents/search/"
+INDEXING_URL=http://find:8000/api/v1.0/documents/index/
+SEARCH_URL=http://find:8000/api/v1.0/documents/search/
+SEARCH_INDEXER_QUERY_LIMIT=50
```
lunika: I don't think we want to enable it by default. If doing this, everybody will need to have find installed on its machine and this is not what we want.

src/backend/core/api/permissions.py
```diff
@@ -12,6 +12,7 @@
 ACTION_FOR_METHOD_TO_PERMISSION = {
     "versions_detail": {"DELETE": "versions_destroy", "GET": "versions_retrieve"},
     "children": {"GET": "children_list", "POST": "children_create"},
+    "search": {"GET": "retrieve"},
```
lunika: This should be added in the abilities (method `get_abilities` in the Document model)

src/backend/core/api/viewsets.py
lunika, line 1002: Removing an action should be made in multiple step.
First add a deprecation warning.
In the next major version it will be removed.
Otherwise integrators using this endpoint will have a Breaking Change and it is not something we want as we follow the [semantic versionning](https://semver.org/)

src/backend/core/api/viewsets.py
```diff
@@ -1304,82 +1165,109 @@ def duplicate(self, request, *args, **kwargs):
...
+        """
+        request.GET = request.GET.copy()
+        request.GET["title"] = validated_data["q"]
 
-        Applies filtering based on request parameter 'q' from `SearchDocumentSerializer`.
-        Depending of the configuration it can be:
-         - A fulltext search through the opensearch indexation app "find" if the backend is
-           enabled (see SEARCH_INDEXER_CLASS)
-         - A filtering by the model field 'title'.
+        if "path" not in validated_data or not validated_data["path"]:
```
```suggestion
        if not validated_data.get("path"):
```

src/backend/core/api/viewsets.py
```diff
@@ -1304,82 +1165,109 @@ def duplicate(self, request, *args, **kwargs):
...
-    @method_decorator(refresh_oidc_access_token)
-    def search(self, request, *args, **kwargs):
+    def _title_search(self, request, validated_data, *args, **kwargs):
         """
-        Returns a DRF response containing the filtered, annotated and ordered document list.
+        Fallback search method when no indexer is configured.
+        Only searches in the title field of documents.
+        """
+        request.GET = request.GET.copy()
+        request.GET["title"] = validated_data["q"]
```
lunika: Doing this you are breaking the `GET` immutable property, so you should not do that.
Also, you should use `request.query_params` instead of `request.GET`

I understand that you are doing this to then use the `ListDocumentFilter`. Why not adding the `q` in this filter to avoid doing this ?

src/backend/core/api/viewsets.py
```diff
@@ -1304,82 +1165,109 @@ def duplicate(self, request, *args, **kwargs):
...
+    def _list_descendants(self, request):
         """
-        params = serializers.SearchDocumentSerializer(data=request.query_params)
-        params.is_valid(raise_exception=True)
+        List all documents whose path starts with the provided path parameter.
+        Includes the parent document itself.
+        Used internally by the search endpoint when path filtering is requested.
+        """
+        # Get parent document without access filtering
+        parent_path = request.GET["path"]
```
lunika: you should not use `request.GET["path"] but rely on the validated_data available in the `_title_search` method.

src/backend/core/api/viewsets.py
```diff
@@ -1304,82 +1165,109 @@ def duplicate(self, request, *args, **kwargs):
...
+        try:
+            parent = models.Document.objects.get(path=parent_path)
+        except models.Document.DoesNotExist as exc:
+            raise drf.exceptions.NotFound("Document not found from path.") from exc
 
-        indexer = get_document_indexer()
+        # Check object-level permissions using DocumentPermission logic
+        self.check_object_permissions(request, parent)
```
```suggestion
        abilities = parent.get_abilities(request.user)
        if not abilities.get("search):
            raise...
```

This is why you should also add the `search` in the abilities and not in the mapping in the permission module

env.d/development/common
mascarpon3: https://github.com/suitenumerique/docs/pull/1834/changes/9f82371b81078ea669d7b39d2b98c9c4e4bf656a

bin/generate-oidc-store-refresh-token-key.sh
mascarpon3: https://github.com/suitenumerique/docs/pull/1834/changes/f035a4017e86a170ea67a90e39122288138088a4

src/backend/core/api/permissions.py
mascarpon3: https://github.com/suitenumerique/docs/pull/1834/changes/a4b511ea063c237443809d8d874adf6f05d31329

src/backend/core/api/viewsets.py
mascarpon3: https://github.com/suitenumerique/docs/pull/1834/changes/8c48a472e4bc0cc630eef7bdcb399e4c6bc97a8a

src/backend/core/api/viewsets.py
mascarpon3: https://github.com/suitenumerique/docs/pull/1834/changes/8c48a472e4bc0cc630eef7bdcb399e4c6bc97a8a

src/backend/core/api/viewsets.py
mascarpon3: https://github.com/suitenumerique/docs/pull/1834/changes/8c48a472e4bc0cc630eef7bdcb399e4c6bc97a8a

src/backend/core/api/viewsets.py
mascarpon3: https://github.com/suitenumerique/docs/pull/1834/changes/8c48a472e4bc0cc630eef7bdcb399e4c6bc97a8a

src/backend/core/api/viewsets.py
```diff
@@ -1087,19 +964,10 @@ def all(self, request, *args, **kwargs):
...
-        filterset = DocumentFilter(request.GET, queryset=queryset)
-        if not filterset.is_valid():
-            raise drf.exceptions.ValidationError(filterset.errors)
-
-        queryset = filterset.qs
-
-        return self.get_response_for_queryset(queryset)
+        """deprecated enpoint, returns warning"""
+        return drf.response.Response(
+            {"detail": "This endpoint is deprecated, please use the search endpopint."}
```
```suggestion
            {"detail": "This endpoint is deprecated, please use the search endpoint."}
```

lunika: I don't really understand why you are deprecating it ? Both endpoint have not the same behavior.

IMO you should not do that. 
A deprecation does not remove what the endpoint/function/whatever is doing. It keeps the same behavior and add a deprecation warning, not for the final user but for the person who is maintaining the application.

And when we will do a major upgrade, then the endpoint can be deleted.

src/backend/core/api/viewsets.py
```diff
@@ -1276,183 +1113,164 @@ def _duplicate_document(
...
-        params = serializers.SearchDocumentSerializer(data=request.query_params)
-        params.is_valid(raise_exception=True)
+        List all documents whose path starts with the provided path parameter.
+        Includes the parent document itself.
+        Used internally by the search endpoint when path filtering is requested.
+        """
+        # Get parent document without access filtering
+        parent_path = validated_data["path"]
+        try:
+            parent = models.Document.objects.get(path=parent_path)
```
```suggestion
            parent = models.Document.objects.annotate_user_roles(request.user).get(path=parent_path)
```

This will save you a `+1 query`, you need the role later in the `get_abilities` method

src/backend/core/api/viewsets.py
mascarpon3: [c05269391e110e3772dbbe136dacb846b4c6f9d0](https://github.com/suitenumerique/docs/pull/1834/changes/f3be93242fae356599038f258b10606a48412e5e)

src/backend/core/api/viewsets.py
mascarpon3: https://github.com/suitenumerique/docs/pull/1834/changes/f3be93242fae356599038f258b10606a48412e5e

src/backend/core/api/viewsets.py
mascarpon3: I am deprecated it because the search endpoint is now in charged of listing sub-docs. Both endpoint do have the same behaviour when searching in sub-docs.

src/backend/core/tests/documents/test_api_documents_descendants_filters.py
```diff
@@ -1,95 +0,0 @@
-"""
-Tests for Documents API endpoint in impress's core app: list
```
lunika, line 2: While the endpoint is not removed, we should not removed the related tests.
Thanks

bin/generate-oidc-store-refresh-token-key.sh
```diff
@@ -0,0 +1,12 @@
...
+
+# Generate the secret OIDC_STORE_REFRESH_TOKEN_KEY and store it to common.local
+
+set -eo pipefail
+
+COMMON_LOCAL="env.d/development/common.local"
+
+OIDC_STORE_REFRESH_TOKEN_KEY=$(openssl rand -base64 32)
+
+echo "OIDC_STORE_REFRESH_TOKEN_KEY=${OIDC_STORE_REFRESH_TOKEN_KEY}" >> "${COMMON_LOCAL}"
```
AntoLC, line 12: If keys are already existing in `common.local`, it inserts new key on the same line as the last one:
```
USER_ONBOARDING_SANDBOX_DOCUMENT="56acd797-0621-4add-bb7e-970fcbd71758"OIDC_STORE_REFRESH_TOKEN_KEY=APa+Y01d2Ix43foNlg81suJ1d6z8ab2KbEH4ntyhxuM=
```

That will fix this issue:
```suggestion
echo "" >> "${COMMON_LOCAL}"
echo "OIDC_STORE_REFRESH_TOKEN_KEY=${OIDC_STORE_REFRESH_TOKEN_KEY}" >> "${COMMON_LOCAL}"
```

You will then have:
```
USER_ONBOARDING_SANDBOX_DOCUMENT="56acd797-0621-4add-bb7e-970fcbd71758"

OIDC_STORE_REFRESH_TOKEN_KEY=APa+Y01d2Ix43foNlg81suJ1d6z8ab2KbEH4ntyhxuM=
```

src/frontend/apps/impress/src/features/docs/doc-management/api/useSearchDocs.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  return response.json() as Promise<DocsResponse>;
+};
+
+export const useSearchDocs = (
+  { q, page, target, parentPath }: SearchDocsParams,
+  queryConfig?: { enabled?: boolean },
+) => {
+  return useQuery<DocsResponse, APIError, DocsResponse>({
+    queryKey: [KEY_LIST_DOC, 'search', { q, page, target, parentPath }],
```
AntoLC: It should have its own key, it is a different endpoint.
```suggestion
    queryKey: [KEY_LIST_SEARCH_DOC, { q, page, target, parentPath }],
```

src/frontend/apps/impress/src/features/docs/doc-management/api/useSearchDocs.tsx
```diff
@@ -0,0 +1,69 @@
...
+    queryFn: () => searchDocs({ q, page, target, parentPath }),
+    ...queryConfig,
+  });
+};
+
+export const useInfiniteSearchDocs = (
+  params: SearchDocsParams,
+  queryConfig?: { enabled?: boolean },
+) => {
+  return useAPIInfiniteQuery(KEY_LIST_DOC, searchDocs, params, queryConfig);
```
AntoLC: Same here:

```suggestion
  return useAPIInfiniteQuery(KEY_LIST_SEARCH_DOC, searchDocs, params, queryConfig);
```

PR #1850, opened 2026-01-31
🐛(backend) add AWS_S3_SIGNATURE_VERSION environment variable support

### LINKED ISSUE #1788: Unable to save pages due to S3 Signature Version incompatibility
## Bug Report

**Problematic behavior**
I created a Doc and was able to type some content into it. But when I go back to the homepage and go back into the Doc, I see no content at all. I see a blank Doc.

Upon further inspection, I found that the frontend made a `PATCH /api/v1.0/documents/:id/` call to save the document content. However, that endpoint returned a 500 error, and the front end does not display any feedback at all. So I am not aware that saving failed until I come back again and see that my content was not saved at all.

Image

I dug deeper, and looked at backend logs. I saw this:

```
botocore.exceptions.ConnectionClosedError: Connection was closed before we received a valid response from endpoint URL: "https://ap-south-1.linodeobjects.com/…".
```

I should note that I am using **Linode Object Storage** as the S3-compatible service. Somehow, retrieving files from the storage service work; only when uploading file it fails with this weird error message.

It took me quite a bit of further digging until I found that this is due to the **signature version incompatibility**: Right now, Docs always use the `s3v4` signature version, which is the default. However, Linode Object Storage only works with the `s3` signature version.

The fix, I thought, would be to set this in my env file:

```
AWS_S3_SIGNATURE_VERSION=s3
```

However, it did not work, because there is no mapping from `os.environ` to Django's settings in here:

https://github.com/suitenumerique/docs/blob/7823303d030fa885e0cbe9a782b5514a9e3e9f92/src/backend/impress/settings.py#L150-L167

Therefore, the environment variable did not get picked up.

**Expected behavior/code**

Docs should accept the `AWS_S3_SIGNATURE_VERSION` environment variable (default is `s3v4`, but for some object storage platforms, this has to be set to `s3` instead).

It would also be great if the frontend notifies me when there was a problem saving the doc because I lost some work in the process 🥺.

**Steps to Reproduce**
1. Configure S3 environment variables to point to Linode Object Storage and deploy.
2. Create a Doc
3. Add some contents
4. Navigate away from the Doc, triggering a save
5. Observe that the save failed (should pass or should have some visual feedback if failed)
6. Come back to the doc, observe that the doc is now a blank page

**Environment**
- Docs version: v4.3.0
- Instance url: (private)

**Possible Solution**
1. Add mapping from environ to django settings as noted above.
2. Add docs to document this issue

dtinth:
## Purpose

Add support for the `AWS_S3_SIGNATURE_VERSION` environment variable to allow configuring S3 signature version for compatibility with S3-compatible storage services like Linode Object Storage.

Fixes #1788

## Proposal

- [x] Add `AWS_S3_SIGNATURE_VERSION` Django setting mapped to environment variable
- [x] Set default value to `s3v4` (current AWS standard)
- [x] Allow users to override with `s3` for older S3-compatible services

## Verification

I deployed the new backend from this branch to my VPS, and updated the S3 credentials to point to Linode Object Storage. I've verified that the backend is now compatible with Linode Object Storage.

image

## External contributions

- [ ] I have added corresponding tests for new features or bug fixes (if applicable) - N/A for settings change


lunika:
Hi,

Thanks for your contribution.
Can you update the `docs/env.md` file accordingly please ?

Thank you

dtinth:
@lunika Thank you for the review, I added the new row in `docs/env.md` in 72db738.

lunika:
Thanks I will manage the error in the CI and then merge it.

PR #1816, opened 2026-01-19
♿️(frontend) Keyboard focus Fixes for docs Tree/Editor

Ovgodd:
## Purpose

Improve keyboard focus management in the docs tree/editor flow: auto-focus title on open, Escape returns focus to the correct tree item, and EmojiPicker Escape returns focus to the originating tree item, including sub-doc cases.

https://github.com/user-attachments/assets/6530ee96-2ae6-4d08-b4db-c3f553cb9a5d

## Proposal

- [x] Focus title only when appropriate (avoid focus traps when re-opening a doc)
- [x] F6 from editor/title returns focus to the right tree item (selected/focused/root)
- [x] EmojiPicker Escape returns focus to the correct tree row (including sub-docs)
- [x] Centralize CSS selectors in a single `cssSelectors` object for reuse

PR #1854, opened 2026-02-02
♿️(frontend) Focus main container after navigation

Ovgodd:
## Purpose

Improve keyboard accessibility when switching document filters by moving focus to the
docs list title after navigation

https://github.com/user-attachments/assets/8d73fa97-6229-4b39-b338-bae991f08ee1

## Proposal

- [x] Focus the docs list title after filter navigation
- [x] Ensure the focus target is a semantic heading for better screen-reader output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Changed
+
+- ♿️(frontend) Focus main container after navigation #1854
+
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


AntoLC:
Let's see if we can leverage `useRouteChangeCompleteFocus` or something similar instead for these PR:
- https://github.com/suitenumerique/docs/pull/1854
- https://github.com/suitenumerique/docs/pull/1816
- https://github.com/suitenumerique/docs/pull/1857

PR #1863, opened 2026-02-05
♿️(frontend) restore focus to triggers after closing menus and modals

Ovgodd:
## Purpose

Restore consistent focus behavior after closing menus, modals, and emoji pickers to improve keyboard accessibility and avoid focus loss across the UI.

## Proposal

- [x] Centralize focus restoration within `useRestoreFocus` store
- [x] Ensure triggers regain focus after closing overlays (menus/modals/emoji)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to
 ### Changed
 
 - 📝(docs) improve README and add documentation hub #1870
+- ♿️(frontend) restore focus to triggers after closing menus and modals #1863
 
 ## [v4.6.0] - 2026-03-03
 
--- a/src/frontend/apps/impress/src/features/docs/doc-management/components/DocIcon.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-management/components/DocIcon.tsx
@@ -13,6 +13,7 @@ import {
   TextType,
   emojidata,
 } from '@/components';
+import { useFocusStore } from '@/stores';
 
 import { useDocTitleUpdate } from '../hooks/useDocTitleUpdate';
 
@@ -41,6 +42,7 @@ export const DocIcon = ({
 }: DocIconProps) => {
   const { updateDocEmoji } = useDocTitleUpdate();
   const { t } = useTranslation();
+  const { addLastFocus, restoreFocus } = useFocusStore();
 
   const iconRef = useRef<HTMLDivElement>(null);
 
@@ -90,12 +92,16 @@ export const DocIcon = ({
         });
       }
 
+      if (!openEmojiPicker) {
+        addLastFocus(iconRef.current);
+      }
       setOpenEmojiPicker(!openEmojiPicker);
     }
   };
 
   const handleEmojiSelect = ({ native }: { native: string }) => {
     setOpenEmojiPicker(false);
+    restoreFocus();
 
     // Update document emoji if docId is provided
     if (docId && title !== undefined) {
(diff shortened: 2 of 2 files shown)


src/frontend/apps/impress/src/hooks/useFocusOnMount.ts
AntoLC, line 1: It seems to be used for all the modal to target the `Cancel` button, why not using the prop `autoFocus` instead, it seems to works as fine and it is much simpler ?

src/frontend/apps/impress/src/features/docs/doc-header/components/DocToolBox.tsx
```diff
@@ -59,6 +64,14 @@ export const DocToolBox = ({ doc }: DocToolBoxProps) => {
...
   const selectHistoryModal = useModal();
   const modalShare = useModal();
+  const optionsButtonRef = useRef<HTMLButtonElement | null>(null);
+  const shareButtonRef = useRef<ButtonElement | null>(null);
+  const exportButtonRef = useRef<ButtonElement | null>(null);
+  const shareTriggerRef = useRef<HTMLElement | null>(null);
+  const exportTriggerRef = useRef<HTMLElement | null>(null);
+  const historyTriggerRef = useRef<HTMLElement | null>(null);
+  const removeTriggerRef = useRef<HTMLElement | null>(null);
+  const restoreFocus = useRestoreFocus();
```
AntoLC: I am not sure we need all that, when a modal open the dropdown close, so we cannot give back the focus on these elements.

src/frontend/apps/impress/src/hooks/useFocusOnMount.ts
Ovgodd: yes I agree !!

Ovgodd:
> I think we can simplify a bit all this code, can you maybe separate in different commits when it is about the docgrid / docTree / docToolbox? It will be easier to validate part

Of course I will do that

src/frontend/apps/impress/src/components/modal/AlertModal.tsx
```diff
@@ -35,21 +35,6 @@ export const AlertModal = ({
...
-  useEffect(() => {
-    const timeout = setTimeout(() => {
-      const contents = document.querySelectorAll('.c__modal__content');
-      contents.forEach((content) => {
-        content.setAttribute('tabindex', '-1');
-      });
-    }, 100);
-    return () => clearTimeout(timeout);
-  }, []);
-
```
AntoLC, line 52: Why did you removed this part ?

src/frontend/apps/impress/src/components/modal/ButtonCloseModal.tsx
AntoLC, line 1: Is it still necessary to forward the ref ?

src/frontend/apps/impress/src/components/DropButton.tsx
```diff
@@ -66,12 +68,14 @@ export const DropButton = ({
...
   label,
   testId,
+  buttonRef,
 }: PropsWithChildren<DropButtonProps>) => {
   const { themeTokens } = useCunninghamTheme();
   const font = themeTokens['font']?.['families']['base'];
   const [isLocalOpen, setIsLocalOpen] = useState(isOpen);
 
-  const triggerRef = useRef(null);
+  const internalTriggerRef = useRef<HTMLButtonElement | null>(null);
```
AntoLC: Does not seem to be ever set.

AntoLC (commented):
As talk together, let's see with a store for useRestoreFocus to avoid props drilling.

PR #1876, opened 2026-02-09
✨(frontend) add floating bar with leftpanel collapse button

Ovgodd:
## Purpose

Add a desktop floating bar and left panel collapse/expand button .

https://github.com/user-attachments/assets/830af933-0a97-4252-bfda-f33e34a70de5

## Proposal

- [x] Add `FloatingBar` on desktop with a dedicated `CollapsePanel` button.
- [x] Connect the collapse button to the left panel store to toggle panel open/close.
- [x] When panel is collapsed, display document title in the button (with emoji when available).
- [x] Keep floating bar disabled on mobile for now (temporary choice to avoid overlap/layout regressions).
- [x] Update E2E tests.
- [x] Ensure left-panel filters do not collapse panel on desktop (close only on mobile).


src/frontend/apps/impress/src/features/floating-bar/components/CollapsePanel.tsx
```diff
@@ -0,0 +1,65 @@
...
+  );
+  const docTitle = titleWithoutEmoji || untitledDocument;
+  const buttonTitle = emoji ? `${emoji} ${docTitle}` : docTitle;
+  const shouldShowButtonTitle = !isPanelOpen && !isDocHeaderVisible;
+  const ariaLabel = t(
+    isPanelOpen
+      ? 'Hide the side panel for {{title}}'
+      : 'Show the side panel for {{title}}',
+    { title: docTitle },
+  );
```
AntoLC: The parser cannot extract the translation when it is like that.


```suggestion
  const ariaLabel = isPanelOpen ? t('Hide the side panel for {{title}}', { title: docTitle }) : t('Show the side panel for {{title}}', { title: docTitle });
```

src/frontend/apps/impress/src/features/left-panel/stores/useLeftPanelStore.tsx
```diff
@@ -7,7 +7,7 @@ interface LeftPanelState {
 }
 
 export const useLeftPanelStore = create<LeftPanelState>((set, get) => ({
-  isPanelOpen: false,
+  isPanelOpen: true,
```
AntoLC, line 11: If the panel is by default at true it creates a flickering when we are on mobiles:

https://github.com/user-attachments/assets/8eb03eea-3702-4564-a498-17c54c77e6af

Maybe not the best solution but we could have 2 states, togglePanel will toggle them both, close panel close them both: 
```ts
isPanelOpen: true,
isPanelOpenMobile: false,
```

src/frontend/apps/impress/src/features/floating-bar/index.ts
AntoLC, line 1: Big part of what is inside the feature `doc-header` will switch inside the floating bar, so I think we should not create a feature `floating-bar`, but add the floating bar components (and others) directly in the feature `doc-header`.

src/frontend/apps/e2e/__tests__/app-impress/doc-editor.spec.ts
```diff
@@ -24,6 +24,45 @@ test.beforeEach(async ({ page }) => {
 });
 
 test.describe('Doc Editor', () => {
+  test('shows floating bar and collapse button on desktop', async ({
```
AntoLC: I would rather add this test in `doc-header.spec.ts`.

We need to assert the title behavior, that appear when we scroll down.

src/frontend/apps/impress/src/features/docs/doc-header/components/DocHeader.tsx
```diff
@@ -23,19 +26,49 @@ interface DocHeaderProps {
...
+      },
+    );
+
+    observer.observe(header);
+
+    return () => {
+      observer.disconnect();
+      setIsDocHeaderVisible(true);
+    };
+  }, [doc.id, setIsDocHeaderVisible]);
```
AntoLC: What is the purpose of this part ? I dont see any diff when I remove it.

src/frontend/apps/impress/src/features/floating-bar/components/FloatingBarLeft.tsx
```diff
@@ -0,0 +1,5 @@
+import { CollapsePanel } from './CollapsePanel';
+
+export const FloatingBarLeft = () => {
+  return <CollapsePanel />;
+};
```
AntoLC: Seems overkill to have a component just wrapping another component, wdyt ?

src/frontend/apps/impress/src/features/floating-bar/stores/useFloatingBarStore.ts
```diff
@@ -0,0 +1,13 @@
...
+  isDocHeaderVisible: boolean;
+  setIsDocHeaderVisible: (visible: boolean) => void;
+}
+
+export const useFloatingBarStore = create<FloatingBarState>((set) => ({
+  isDocHeaderVisible: true,
+  setIsDocHeaderVisible: (visible) => {
+    set({ isDocHeaderVisible: visible });
+  },
+}));
```
AntoLC: I am not totally sure why you need a store for this component, you seems to use it only in `CollapsePanel.tsx`. Could we add directly the logic inside `CollapsePanel.tsx` component ? If the classname `--docs--doc-title` is not in the viewport anymore, we display the title.

src/frontend/apps/impress/src/features/floating-bar/components/FloatingBar.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { useResponsiveStore } from '@/stores';
+
+import { FloatingBarLeft } from './FloatingBarLeft';
+
+export const FLOATING_BAR_HEIGHT = '64px';
+export const FLOATING_BAR_Z_INDEX = 1000;
+const FLOATING_BAR_BLUR_RADIUS = '1px';
+const FLOATING_BAR_GRADIENT =
+  'linear-gradient(180deg, #FFF 0%, rgba(255, 255, 255, 0) 100%)';
+
```
AntoLC: Could we add this part directly inside getFloatingBarStyles ? The css is already meaningfull `backdrop-filter: blur`.

src/frontend/apps/impress/src/features/floating-bar/components/FloatingBar.tsx
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <Box
+      data-testid="floating-bar"
+      $css={getFloatingBarStyles(
+        mainContentPadding,
+        barSpacing,
+        FLOATING_BAR_BLUR_RADIUS,
+      )}
+    >
```
AntoLC: Same for that, I think we can avoid the props, wdyt ?


```suggestion

  if (!isDesktop) {
    return null;
  }

  return (
    <Box
      data-testid="floating-bar"
      $css={FLOATING_STYLE}
    >
```

src/frontend/apps/impress/src/hooks/useRouteChangeCompleteFocus.tsx
AntoLC, line 1: I am not sure what you want to do here, can you describe ?

src/frontend/apps/impress/src/layouts/MainLayout.tsx
```diff
@@ -120,9 +134,13 @@ const MainContent = ({
       $css={css`
         overflow-y: auto;
         overflow-x: clip;
-        &:focus-visible {
-          outline: 3px solid ${colorsTokens['brand-400']};
-          outline-offset: -3px;
+        &:focus-visible::after {
```
AntoLC, line 123: Will it not create a double kind of outline ?

src/frontend/apps/impress/src/features/floating-bar/components/FloatingBar.tsx
```diff
@@ -0,0 +1,77 @@
...
+  margin-right: calc(-${mainContentPadding});
+  margin-top: calc(-${mainContentPadding});
+  z-index: ${FLOATING_BAR_Z_INDEX};
+  display: flex;
+  align-items: flex-start;
+  justify-content: flex-start;
+  background: ${FLOATING_BAR_GRADIENT};
+  backdrop-filter: blur(${blurRadius});
+  -webkit-backdrop-filter: blur(${blurRadius});
+
```
AntoLC: If you want to have a gradual blurry (soft at the beggining): 
```suggestion
  backdrop-filter: blur(${blurRadius});
  -webkit-backdrop-filter: blur(${blurRadius});
  mask-image: linear-gradient(180deg, black 0%, transparent 100%);
  -webkit-mask-image: linear-gradient(180deg, black 50%, transparent 100%);
```

src/frontend/apps/impress/src/features/floating-bar/index.ts
Ovgodd: Ok that makes sense !

src/frontend/apps/impress/src/features/left-panel/stores/useLeftPanelStore.tsx
Ovgodd: oho good catch

src/frontend/apps/impress/src/features/docs/doc-header/components/DocHeader.tsx
Ovgodd: > What is the purpose of this part ? I dont see any diff when I remove it.

It shows the doc title in the floating bar button when the main title scrolls out of view as suggested by robin.

src/frontend/apps/impress/src/features/floating-bar/components/FloatingBar.tsx
Ovgodd: ah yes nice ! I had it and removed for unknown reasons, thank you

src/frontend/apps/impress/src/layouts/MainLayout.tsx
Ovgodd: I think it won’t. 

There’s only the ::after for the focus ring, no outline on the main element. The ::after is used so the focus ring renders above the floating bar (z-index 2001).

src/frontend/apps/impress/src/hooks/useRouteChangeCompleteFocus.tsx
Ovgodd: > I am not sure what you want to do here, can you describe ?

The problem was : On every route change, focus always moved to main, whether by mouse or keyboard. With mouse clicks, this caused unexpected scrolling and focus stealing, and the focus ring sometimes appeared behind the floating bar due to z-index.

Solution: Focus is now moved to main only when navigation is triggered from the keyboard (Tab, Enter, or Space). Mouse users are unaffected; keyboard and screen reader users still get proper focus managemen

PR #1880, opened 2026-02-10
🚸(oidc) ignore case when fallback on email

qbey:
## Purpose

Some identity providers might change the case, but in our products we don't consider case variation to be consider as different email addresses.

Next step would be to normalize the DB value of email to be lower-case.

## Proposal

- [x] ignore email case in OIDC login fallback
- [x]  ignore email case in invitations

### DIFF
--- a/src/backend/core/tests/documents/test_api_document_invitations.py
+++ b/src/backend/core/tests/documents/test_api_document_invitations.py
@@ -596,6 +596,38 @@ def test_api_document_invitations_create_cannot_invite_existing_users():
     }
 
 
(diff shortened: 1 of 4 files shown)


src/backend/core/tests/documents/test_api_document_invitations.py
```diff
@@ -596,6 +596,38 @@ def test_api_document_invitations_create_cannot_invite_existing_users():
     }
 
 
+def test_api_item_invitations_create_cannot_invite_existing_users_case_insensitive():
```
lunika, line 599: Not sure to understand why you are adding this test. It is not related to the code you are modifying.

PR #1882, opened 2026-02-11
✨(backend) add management command to reset a Document

lunika:
## Purpose

We need a management command to reset a Document to an initial state and deletes everything related to it. This command can be usefull to reset a demo for example.

## Proposal

- [x] ✨(backend) add management command to reset a Document

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to
 - ✨(y-provider) preserve callouts, PDFs, page breaks, interlinking
   links and commented text on HTML/markdown export #2296
 - ✨(frontend) add a user menu #2463
+- ✨(backend) add management command to reset a Document #1882
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


AntoLC (commented):
I noticed a few problems:
- if you reset the root, all the members are removed from the documents, even the owner, so you cannot access the document or sub documents anymore
- I guess a demo doc is often public, the command seems to switch its visibility to private, if the command is automatized, you will have to manually change the visibility to the doc again
- nit there is still the data in the table thread

AntoLC (approved):
Nice, just need a doc now 😜

virgile-dev:
The doc is this one : https://docs.la-suite.eu/docs/486d0d17-a1e7-423d-935a-810e8a2b40f9/ @lunika @AntoLC

AntoLC:
> The doc is this one : https://docs.la-suite.eu/docs/486d0d17-a1e7-423d-935a-810e8a2b40f9/ @lunika @AntoLC

I was speaking about a doc in GitHub repo about how to use this feature, users will not be aware this command is now existing and if so, to understand how to use it by looking at the code is not trivial.

PR #1886, opened 2026-02-12
✨(frontend) Move doc modal

AntoLC:
## Purpose

It is not every time easy to move document from the grid with the drag and drop.
We added a search modal to move document.

## Proposal

- [x] search modal to move documents
- [x] confirmation modal if accesses
- [x] access request modal when not enough rights

## Demo

https://github.com/user-attachments/assets/0e76c9fe-8f79-41f2-8363-4eceb57b8665

### DIFF
--- a/src/frontend/apps/impress/src/components/modal/AlertModal.tsx
+++ b/src/frontend/apps/impress/src/components/modal/AlertModal.tsx
@@ -1,4 +1,10 @@
-import { Button, Modal, ModalSize } from '@gouvfr-lasuite/cunningham-react';
+import {
+  Button,
+  ButtonProps,
+  Modal,
+  ModalProps,
+  ModalSize,
+} from '@gouvfr-lasuite/cunningham-react';
 import { ReactNode, useEffect } from 'react';
 import { useTranslation } from 'react-i18next';
 
@@ -10,10 +16,11 @@ export type AlertModalProps = {
   isOpen: boolean;
   onClose: () => void;
   onConfirm: () => void;
+  themeCTA?: ButtonProps['color'];
   title: string;
   cancelLabel?: string;
   confirmLabel?: string;
-};
+} & Partial<ModalProps>;
 
 export const AlertModal = ({
   cancelLabel,
@@ -23,6 +30,8 @@ export const AlertModal = ({
   onClose,
   onConfirm,
   title,
+  themeCTA,
+  ...props
 }: AlertModalProps) => {
   const { t } = useTranslation();
 
@@ -71,13 +80,14 @@ export const AlertModal = ({
           </Button>
           <Button
             aria-label={confirmLabel ?? t('Confirm')}
-            color="error"
+            color={themeCTA ?? 'error'}
             onClick={onConfirm}
           >
             {confirmLabel ?? t('Confirm')}
           </Button>
         </Box>
       }
+      {...props}
     >
       <Box className="--docs--alert-modal">
         <Box>
--- a/src/frontend/apps/impress/src/features/docs/doc-share/components/AlertModalRequestAccess.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-share/components/AlertModalRequestAccess.tsx
@@ -1,0 +1,89 @@
+import { Button } from '@gouvfr-lasuite/cunningham-react';
+import { Trans, useTranslation } from 'react-i18next';
+
+import { AlertModal, Box, Icon, Text } from '@/components';
+
+import { useDocAccessRequests } from '../api/useDocAccessRequest';
+
+import { ButtonAccessRequest } from './DocShareAccessRequest';
+
+interface AlertModalRequestAccessProps {
+  docId: string;
+  isOpen: boolean;
+  onClose: () => void;
+  onConfirm: () => void;
+  targetDocumentTitle: string;
+  title: string;
+}
+
+export const AlertModalRequestAccess = ({
+  docId,
+  isOpen,
+  onClose,
+  onConfirm,
+  targetDocumentTitle,
+  title,
+}: AlertModalRequestAccessProps) => {
+  const { t } = useTranslation();
+  const { data: requests } = useDocAccessRequests({
+    docId,
+    page: 1,
+  });
+
+  const hasRequested = !!(
+    requests && requests?.results.find((request) => request.document === docId)
+  );
+
+  return (
+    <AlertModal
+      onClose={onClose}
+      isOpen={isOpen}
+      title={title}
+      aria-label={t('Request access modal')}
+      description={
+        <>
+          <Text $display="inline">
+            <Trans
+              i18nKey="You don't have permission to move this document to <strong>{{targetDocumentTitle}}</strong>. You need edit access to the destination. Request access, then try again."
+              values={{
+                targetDocumentTitle,
+              }}
+              components={{ strong: <strong /> }}
+            />
+          </Text>
+          {hasRequested && (
+            <Text
+              $weight="bold"
+              $margin={{ top: 'sm' }}
+              $direction="row"
+              $align="center"
+            >
+              <Icon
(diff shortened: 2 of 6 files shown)


src/frontend/apps/impress/src/features/docs/docs-grid/components/DocMoveModal.tsx
```diff
@@ -0,0 +1,297 @@
...
+    });
+  };
+
+  return (
+    <>
+      <DocMoveModalStyle />
+      <Modal
+        isOpen={isOpen}
+        onClose={onClose}
+        closeOnClickOutside
```
Ovgodd, line 108: same here

```suggestion
        closeOnClickOutside
        aria-label={t(move-modal)}
```

src/frontend/apps/impress/src/features/docs/doc-share/components/AlertModalRequestAccess.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  const hasRequested = !!(
+    requests && requests?.results.find((request) => request.document === docId)
+  );
+
+  return (
+    <AlertModal
+      onClose={onClose}
+      isOpen={isOpen}
+      title={title}
```
Ovgodd, line 41: Actually the SR announces [object Object],
This is because the title prop is a ReactNode, and Cunningham internally calls title.toString() to set the dialog's accessible name. Adding an explicit aria-label on the <Modal> fixes it since Cunningham prioritizes it over title.toString().

```suggestion
      title={title}
      aria-label={t('Request access modal')}
```

src/frontend/apps/impress/src/features/docs/docs-grid/components/DocMoveModal.tsx
Ovgodd, line 1: When you press Enter on a doc, it only looks selected visually, the screen reader isn’t notified.

To fix it we can :

- Add an isSelected prop to QuickSearchItem
- Pass it to Command.Item as aria-selected={isSelected}
- Propagate the selected state from DocMoveModal (using docSelected?.id)
- Let QuickSearchGroup pass isSelected to the correct item (via selectedId or a callback)
- This way, the selected item gets aria-selected="true" and the screen reader announces it properly.  

wdyt ?

src/frontend/apps/impress/src/features/docs/doc-share/components/AlertModalRequestAccess.tsx
```diff
@@ -0,0 +1,87 @@
...
+      }
+      confirmLabel={t('Request access')}
+      onConfirm={onConfirm}
+      rightActions={
+        <Box $direction="row-reverse" $gap="small">
+          <ButtonAccessRequest docId={docId} onClick={onConfirm} />
+          <Button
+            aria-label={t('Cancel')}
+            variant="secondary"
+            fullWidth
```
Ovgodd, line 78: It’s good practice to move focus to an interactive element when a modal opens, here, the Cancel button.

- Focusing Cancel is intentional because:
- It provides a safe, non-destructive default action
- It lets keyboard users immediately close the modal with Enter or Space
- It avoids accidentally triggering a primary or destructive action
- It gives screen reader users a clear, actionable starting point inside the dialog

```suggestion
            fullWidth
            autoFocus
```

src/frontend/apps/impress/src/components/modal/AlertModal.tsx
```diff
@@ -43,7 +52,14 @@ export const AlertModal = ({
...
+          <Button
+            aria-label={confirmLabel ?? t('Confirm')}
+            color={themeCTA ?? 'error'}
+            onClick={onConfirm}
+          >
+            {confirmLabel ?? t('Confirm')}
+          </Button>
           <Button
             aria-label={`${t('Cancel')} - ${title}`}
             variant="secondary"
```
Ovgodd, line 74:
```suggestion
            variant="secondary"
            autoFocus
```

src/frontend/apps/impress/src/components/modal/AlertModal.tsx
```diff
@@ -43,7 +52,14 @@ export const AlertModal = ({
         </Text>
       }
       rightActions={
```
Ovgodd, line 70: row-reverse causes the DOM order to differ from the visual order, which breaks tab order for keyboard/screen reader users. 
Actually we have "Cancel > Close > Text > Confirm"

```suggestion
       rightActions={
        <Box $direction="row" $gap="small">
          <Button
            aria-label={`${t('Cancel')} - ${title}`}
            variant="secondary"
            fullWidth
            autoFocus
            onClick={() => onClose()}
          >
            {cancelLabel ?? t('Cancel')}
          </Button>
          <Button
            aria-label={confirmLabel ?? t('Confirm')}
            color={themeCTA ?? 'error'}
            onClick={onConfirm}
          >
            {confirmLabel ?? t('Confirm')}
          </Button>
        </Box>
      }
```

src/frontend/apps/impress/src/features/docs/doc-share/components/AlertModalRequestAccess.tsx
```diff
@@ -0,0 +1,87 @@
...
+                variant="symbols-outlined"
+              />
+              {t('You have already requested access to this document.')}
+            </Text>
+          )}
+        </>
+      }
+      confirmLabel={t('Request access')}
+      onConfirm={onConfirm}
+      rightActions={
```
Ovgodd, line 73: same here : 
broken tab order
```suggestion
      rightActions={
        <Box $direction="row" $gap="small">
          <Button
            aria-label={t('Cancel')}
            variant="secondary"
            fullWidth
            autoFocus
            onClick={onClose}
          >
            {t('Cancel')}
          </Button>
          <ButtonAccessRequest docId={docId} onClick={onConfirm} />
        </Box>
      }
```

src/frontend/apps/impress/src/components/modal/AlertModal.tsx
Ovgodd, line 1: <Modal> sets tabindex="0" on .c__modal__content, making it a tab stop. 

SR users hit the description text when pressing Tab, but only interactive elements should be tabbable. The text is still readable via the SR browse mode (arrow keys). Changing to tabindex="-1" would fix this. 
we could do this : 
```

useEffect(() => {
  if (!isOpen) return;
  const content = document.querySelector('.c__modal__content');
  if (content) {
    content.setAttribute('tabindex', '-1');
  }
}, [isOpen]);

```
Image

Ovgodd (changes requested):
Nice feature! Just a few accessibility improvements to consider.

src/frontend/apps/impress/src/components/modal/AlertModal.tsx
AntoLC: Actually it was by purpose, to have "Confirm" being focused before Cancel. But if Cancel is `autofocus` yes it does not make sense anymore.

src/frontend/apps/impress/src/features/docs/docs-grid/components/DocMoveModal.tsx
AntoLC: The secret 🪄 

https://github.com/suitenumerique/docs/blob/c0d3ac7d4bc2618085df462686d39e55cc63156c/src/frontend/apps/impress/src/components/quick-search/QuickSearch.tsx#L76

src/frontend/apps/impress/src/components/modal/AlertModal.tsx
AntoLC: I made a patch as well to Cunningham as it is the root problem: https://github.com/suitenumerique/cunningham/pull/377/

https://github.com/suitenumerique/docs/blob/f5a6888ca3678feca63637ef873a265d7804ff32/src/frontend/apps/impress/src/components/modal/AlertModal.tsx#L38-L51

PR #1891, opened 2026-02-16
🚸(backend) add onboarding docs for new users

### LINKED ISSUE #393: Onboarding Docs for new users
## Feature Request

Some users are not super familiar with block based structured editors, slash commands, markdown etc.
Docs is super confusing for them.

For new users instead of having no docs. Why not display a doc which content can serve as tutorial. Where there would be examples of all text blocs. Instructions to play around etc.

Adding a tooltip on the docs editiing page. When you click on it you get a modal that explain the features of the editor. Like in the pad. But I feel it's less interactive.
Image

Notion isn't empty when you join. You can play with template to get started.
Image

Ash-Crow:
## Purpose
Some users are not super familiar with block based structured editors, slash commands, markdown etc.
Docs is super confusing for them. This adds both read-only tutorials and a sandbox owned by the user.

Fix #393 

## Proposal

- [x] Adds two methods to allow new users to start with some docs.
  - User._handle_onboarding_documents_access() gives READER access to each document listed in settings.USER_ONBOARDING_DOCUMENTS.
  - User._duplicate_onboarding_sandbox_document() creates a local copy of the sandbox document specified in
settings.USER_ONBOARDING_SANDBOX_DOCUMENT.
- [x] Fix some formating issues with the `docs/env.md` file with the help of a [Markdown table formatter](https://finitefield.org/fr/tools/markdown/markdown-table-formatter/)

### DIFF
--- a/src/backend/core/models.py
+++ b/src/backend/core/models.py
@@ -260,24 +260,25 @@ class User(AbstractBaseUser, BaseModel, auth_models.PermissionsMixin):
             try:
                 template_document = Document.objects.get(id=sandbox_id)
 
-                sandbox_document = template_document.add_sibling(
-                    "right",
-                    title=template_document.title,
-                    content=template_document.content,
-                    attachments=template_document.attachments,
-                    duplicated_from=template_document,
-                    creator=self,
-                )
-
-                DocumentAccess.objects.create(
-                    user=self, document=sandbox_document, role=RoleChoices.OWNER
-                )
-
             except Document.DoesNotExist:
                 logger.warning(
                     "Onboarding sandbox document with id %s does not exist. Skipping.",
                     sandbox_id,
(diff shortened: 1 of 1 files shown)


src/backend/core/models.py
```diff
@@ -209,14 +209,76 @@ def __str__(self):
...
+
+                DocumentAccess.objects.create(
+                    user=self, document=sandbox_document, role=RoleChoices.OWNER
+                )
+
+            except Document.DoesNotExist:
+                logger.warning(
+                    "Onboarding sandbox document with id %s does not exist. Skipping.",
+                    sandbox_id,
+                )
```
lunika, line 267: I would move the except block right after `template_document = Document.objects.get(id=sandbox_id)`

PR #1892, opened 2026-02-18
🔧(compose) replace minio by rustfs

lunika:
## Purpose

Minio is no longer maintained, its git repository is archived. We need to replace it. We decided to choose rustfs
https://github.com/minio/minio/commit/7aac2a2c5b7c882e68c1ce017d8256be2feea27f
https://github.com/rustfs/rustfs

## Proposal

- [x] 🔧(compose) replace minio by rustfs
- [ ] replace minio in github actions
- [ ] replace minio in the helm run in dev mode

### DIFF
--- a/.github/workflows/impress.yml
+++ b/.github/workflows/impress.yml
@@ -189,7 +189,7 @@ jobs:
 
       - name: Configure Rustfs
         run: |
-          docker exec rustfs sh -c \
+          docker run --rm -it minio/mc sh -c \
             "mc alias set impress http://localhost:9000 impress password && \
             mc alias ls && \
             mc mb impress/impress-media-storage && \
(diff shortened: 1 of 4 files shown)


z3ntu:
Good to see something being done about Minio being deprecated/stopped. Is there anything missing, or did this PR just fly under the radar?

lunika:
I was waiting to have feedback about this solution. If you have feedback, already using it, you can share it here.

z3ntu:
It's mostly interest from me running La Suite Docs on a server, checking what's done upstream about Minio. If it's useful, we can update this on the infra and see how it's performing. But we're just a few users, so apart from working vs. not working I don't think we can provide too much feedback :) I also don't really have any other experience with Minio or related software

PR #1898, opened 2026-02-19
🐛 (backend) prevent privileged users from requesting access

maboukerfa:
## Purpose
Prevent privileged users (owners and admins) of a document from submitting an access request on a document they already control.

Without this guardrail, two problems arise:

- Queue pollution — privileged users could inadvertently inflate the access request queue with redundant entries.
- Silent privilege downgrade — more critically, if an owner submits an access request on their own document, another admin could accept it and grant them a lower role (e.g. reader), which would silently strip them of their ownership.

**Example:**
We have a document shared between 2 users:
- John Doe (owner)
- Amine BOUKERFA (admin)
Screenshot 2026-02-19 at 13 32 28

If Amine approves John’s request for read-only access, John will lose ownership of the document. If that happens, neither of them will be able to delete it.

## Proposal
 - Add a validation check in DocumentAskForAccessViewSet.create() that returns a 400 Bad Request if the requesting user already holds a privileged role (OWNER or ADMIN) on the document.
 - Add parametrized tests covering both

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,7 +53,7 @@ and this project adheres to
 - ♿️(frontend) add focus on open to modals #1948
 
 ### Fixed
-
+- 🐛 (backend) prevent privileged users from requesting access #1898
 - 🐛(frontend) fix broadcast store sync #1846
 - 🐛(helm) use celery resources instead of backend resources #1887
 - 🐛(helm) reverse liveness and readiness for backend deployment #1887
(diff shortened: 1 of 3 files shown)


maboukerfa:
Thanks for the approval! my last push was to sign the commit. Just checking in, is there any additional action required from me to get this merged, or is it just pending in the queue? Let me know if anything else is needed.

lunika:
Hi,
Thanks for your contribution.
We have some trouble with the CI, once fixed we will be able to merge it.

maboukerfa:
Alright! Thank you

PR #1910, opened 2026-02-23
⚡️(backend) remove content from Document serializer when asked

### LINKED ISSUE #1229: ⚡️Request `/api/v1.0/documents/[DOC_ID/` without `content` property
## Feature Request

We need to do request on `/api/v1.0/documents/[DOC_ID/` to get information about a doc. 
Sometime we just need the title of a doc, but the request does not give the ability to filter the property needed so we get all the properties.
The problem with that is `content` is a base64 of the doc content, and it can be relatively heavy.

It could be nice to provide a way to have the response without `content` property, or a special routes to get the content.

```
Request URL  `/api/v1.0/documents/[DOC_ID/?without_content=true`
Request Method GET
```
or
```
Request URL  `/api/v1.0/documents/[DOC_ID/content/`
Request Method GET POST
```

By example here we just need the abilities:
https://github.com/suitenumerique/docs/blob/05aa225aed0986168c35abb3ef10c60ad9a6bd4d/src/frontend/servers/y-provider/src/servers/hocuspocusServer.ts#L42

Here only the title:
https://github.com/suitenumerique/docs/blob/05aa225aed0986168c35abb3ef10c60ad9a6bd4d/src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-inline-content/Interlinking/InterlinkingLinkInlineContent.tsx#L30

Just adding a few links like that, because we need to use `/api/v1.0/documents/[DOC_ID/` will pull the `content` of each of them:
Image

lunika:
## Purpose

The frontend can fetch the retrieve endpoint just for having the title. Everytime, the content is added and to get the content, a request is made to the s3 bucket. A query string `without_content` can be used, if the value is `true` then the content is not added (and not also not fetch).

## Proposal

- [x] ⚡️(backend) remove content from Document serializer when asked

Fix #1229

### DIFF
--- a/src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-inline-content/Interlinking/InterlinkingLinkInlineContent.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-inline-content/Interlinking/InterlinkingLinkInlineContent.tsx
@@ -95,7 +95,7 @@ export const LinkSelected = ({
   isEditable,
   updateInlineContent,
 }: LinkSelectedProps) => {
-  const { data: doc } = useDoc({ id: docId });
+  const { data: doc } = useDoc({ id: docId, withoutContent: true });
 
   /**
    * Update the content title if the referenced doc title changes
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to
 - ✨(frontend) Can print a doc #1832
 - ✨(backend) manage reconciliation requests for user accounts #1878
 - 👷(CI) add GHCR workflow for forked repo testing #1851
+- ⚡️(backend) remove content from Document serializer when asked
```
Ash-Crow: Shouldn't `CHANGELOG.md` entries always include the PR number?

PR #1914, opened 2026-02-24
🐛(frontend) fix callout block spacing for old browsers

AntoLC:
## Purpose

On the old Firefox versions, the spacing of the callout block was not applied correctly.
We changed the "white-space" property to "pre-wrap" to ensure that the spacing is applied correctly on all browsers.

## Demo

https://github.com/user-attachments/assets/34b6e800-fe90-4911-bbd9-54fecdb20103


Ovgodd (approved):
💯

PR #1919, opened 2026-02-25
⚡️(CI) optimize Docker Hub workflow

AntoLC:
## Purpose

We use the Docker Hub Workflow to build and push our images to Docker Hub, but to check if we have vulnerabilities in our images as well.
When we are just checking for vulnerabilities, we don't need to do all the builing steps.
This commit optimizes the workflow by only doing the necessary steps when we are just checking for vulnerabilities, so during pull requests without label "preview" we skip the build steps, and we do not activate QEMU.

### DIFF
--- a/.github/workflows/docker-hub.yml
+++ b/.github/workflows/docker-hub.yml
@@ -11,10 +11,10 @@ on:
   pull_request:
     branches:
       - "main"
-      - "ci/trivy-fails"
 
 env:
   DOCKER_USER: 1001:127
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-hub.yml
```diff
@@ -26,6 +26,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
       - name: Set up QEMU
+        if: env.SHOULD_PUSH == 'true'
         uses: docker/setup-qemu-action@v3
       - name: Set up Docker Buildx
```
lunika, line 31: Here you can also add the `if` condition. And in all other jobs

lunika (approved):

PR #1923, opened 2026-02-26
✨(backend) add resource server api

Ash-Crow:
## Purpose
Adds a resource server API similar to the one that already exists for Drive.

## Proposal

- [x] Add a new `resource_server` API with four endpoints : `documents`, `document_accesses`, `document_invitation`, `users`, along with unit tests that check that :
  - anonymous users can't use these endpoints
  - logged-in users can't use the endpoints that are not active by default
  - for each action that is allowed by default, that it works
  - for each action that is not allowed by default, that is is forbidden, and works if explicitely allowed (see detailed list below)
 - [x] Add a documentation for the new API at `docs/resource_server.md`. 
- [x] Fix some intermittent errors in other unit tests
- [x] Fix some missing entries in the DocumentViewSet docstring.

## List of actions for each endpoint
- For documents:
  - [x] list (allowed by default)
  - [x] retrieve (allowed by default)
  - [x] create (allowed by default)
  - [x] update
  - [x] delete
  - [x] trashbin
  - [x] restore → added to the documentation, it was missing
  - [x] move → added to the documentation, it was missing
  - [x] duplicate → bypasses restriction checks, so always allowed for logged-in users
  - [x] children (allowed by default)
  - [x] versions_list
  - [x] versions_detail
  - [x] favorite_list → bypasses restriction checks, so always allowed for logged-in users
  - [x] favorite
  - [x] create-for-owner → not allowed.
  - [x] link_configuration
  - [x] attachment_upload
  - [x] media_auth
  - [x] ai_transform
  - [x] ai_translate
  - [x] ai_proxy

- For document accesses:
  - [x] list
  - [x] retrieve
  - [x] create
  - [x] update
  - [x] partial_update
  - [x] destroy

- For invitations:
  - [x] list
  - [x] retrieve
  - [x] create
  - [x] partial_update
  - [x] destroy

- For users:
  - [x] get_me (allowed by default)
  - [x] check that other actions are not allowed

### DIFF
--- a/docs/resource_server.md
+++ b/docs/resource_server.md
@@ -1,0 +1,106 @@
+# Use Docs as a Resource Server
+
+Docs implements resource server, so it means it can be used from an external app to perform some operation using the dedicated API.
+
+> **Note:** This feature might be subject to future evolutions. The API endpoints, configuration options, and behavior may change in future versions.
+
+## Prerequisites
+
+In order to activate the resource server on Docs you need to setup the following environment variables
+
+```python
+OIDC_RESOURCE_SERVER_ENABLED=True
+OIDC_OP_URL=
+OIDC_OP_INTROSPECTION_ENDPOINT=
+OIDC_RS_CLIENT_ID=
+OIDC_RS_CLIENT_SECRET=
+OIDC_RS_AUDIENCE_CLAIM=
+OIDC_RS_ALLOWED_AUDIENCES=
+```
+
+It implements the resource server using `django-lasuite`, see the [documentation](https://github.com/suitenumerique/django-lasuite/blob/main/documentation/how-to-use-oidc-resource-server-backend.md)
+
+## Customise allowed routes
+
+Configure the `EXTERNAL_API` setting to control which routes and actions are available in the external API. Set it via the `EXTERNAL_API` environment variable (as JSON) or in Django settings.
+
+Default configuration:
+
+```python
+EXTERNAL_API = {
+    "documents": {
+        "enabled": True,
+        "actions": ["list", "retrieve", "create", "children"],
+    },
+    "document_access": {
+        "enabled": False,
+        "actions": [],
+    },
+    "document_invitation": {
+        "enabled": False,
+        "actions": [],
+    },
+    "users": {
+        "enabled": True,
+        "actions": ["get_me"],
+    },
+}
+```
+
+**Endpoints:**
+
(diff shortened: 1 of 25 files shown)


docs/resource_server.md
```diff
@@ -0,0 +1,106 @@
...
+            {
+                "file": file,
+            },
+            format="multipart",
+        )
+
+        response.raise_for_status()
+        data = response.json()
+        return {"id": data["id"]}
+```
```
lunika, line 95: This example does not work for Docs, only for Drive.

src/backend/core/external_api/permissions.py
```diff
@@ -0,0 +1,41 @@
...
+        """
+        if not isinstance(
+            request.successful_authenticator, ResourceServerAuthentication
+        ):
+            # Not a resource server request
+            return False
+
+        # Check if the user is authenticated
+        if not request.user.is_authenticated:
+            return False
```
lunika, line 31:
```suggestion
            return False

```

gustavotrott:
Hey, I'm testing the API and ran into an error that I'm not sure is expected.
When I create a document using a service token (or user token either), I can't delete that document afterward.

The response returned is:
`{"detail":"You do not have permission to perform this action."}`

Script used for testing:

```bash
TOKEN=$(curl -s -X POST \
  "https://$KEYCLOAK_HOST/realms/$REALM/protocol/openid-connect/token" \
  -H "Content-Type: application/x-www-form-urlencoded" \
  -d "client_id=$CLIENT_ID" \
  -d "client_secret=$CLIENT_SECRET" \
  -d "grant_type=client_credentials" \
  -d "scope=openid email" \
  | jq -r '.access_token')

# Create document
json_return=$(curl -s -X POST "$HOST/external_api/v1.0/documents/" \
  -H "Authorization: Bearer $TOKEN" \
  -F "file=@/home/gustavo/README.md;type=text/markdown")

DOC_ID=$(echo $json_return | jq -r '.id')
CONTENT=$(echo $json_return | jq -r '.content')

# Remove document
curl -X DELETE "$HOST/external_api/v1.0/documents/$DOC_ID/" \
  -H "Authorization: Bearer $TOKEN"
```

Am I missing something?

lunika:
Hi,

In the settings there is a list of allowed actions for each viewset. And `destroy` is not an allowed one by default. You have to add it, it is explained in the documentation present in this PR.

gustavotrott:
> Hi,
> 
> In the settings there is a list of allowed actions for each viewset. And `destroy` is not an allowed one by default. You have to add it, it is explained in the documentation present in this PR.

I see it now, thank you!

Issue #1796, opened 2026-01-13
Onboarding modal for Docs


rl-83:
User research revealed that new users often misunderstood Docs. We propose a user-triggered onboarding modal that can be reactivated at any time and scaled across the entire suite.

[Here is the proposal validated](https://www.figma.com/design/qdCWR4tTUr7vQSecEjCyqO/Docs?node-id=12047-16316&t=QxWx4pA5X9yePywC-1) with @virgile-dev. Ready for dev!

virgile-dev:
Closing  https://github.com/suitenumerique/docs/issues/1835 in favor of this one

virgile-dev:
@AntoLC this was implemented for Drive in their latest release. Maybe you can reuse some stuff.

AntoLC:
We did the dev already. 

We are waiting for:
- [ ] Backend - We must have a flag telling that the user has done is first login

----

We didn't merge what we have done; because the Help button is not validated anymore apparently, I don't find the design about it anymore:

<img width="558" height="555" alt="Image" src="https://github.com/user-attachments/assets/ee3a0376-beea-4bdc-bf28-c46f6617b788" />

virgile-dev:
This was done on Drive. We should be able to do a simple backport.

AntoLC:
> This was done on Drive. We should be able to do a simple backport.

I don't think it is made on Drive, they made a release modal, that explain the last features, a flag is per user with the last release they have seen. 
The onboarding modal is different, it explains de app globally, it is display only at the first user connection, and when they click on the information button, but this button seems to have disappear from the design.

If you think it is made on Drive, could you share a screenshot or/and the pull request to be sure ?

PR #1938, opened 2026-03-02
✨(backend) add a is_first_connection flag to the User model

Ash-Crow:
## Purpose
Backend part of #1796.

This changes allows to display an onboarding modal the first time that the get_me() API view is called.
I originally tried to check if `User.last_login` was `None`, but it is updated as soon as the user is logged, so I chose to create a flag on the model.

## Proposal
- [x] Add a `is_first_connection` boolean field on the user, that is updated ~~the first time the `get_me()` API view is called~~ through the new `onboarding_done` on the `UserViewSet`.


Ash-Crow:
@lunika and @AntoLC: As per our discussion, I updated the PR so that the `is_first_connection` field is updated through an API call from the front-end (a `POST` request on `/api/v1.0/users/onboarding-done/`)

src/backend/core/migrations/0030_user_is_first_connection.py
AntoLC, line 1: I am wondering if we should not update it to false for the existing users, because everybody is going to get the onboarding modal at the next release otherwise.

src/backend/core/api/viewsets.py
```diff
@@ -317,6 +317,25 @@ def get_me(self, request):
...
+        permission_classes=[permissions.IsAuthenticated],
+    )
+    def onboarding_done(self, request):
+        """
+        Allows the frontend to mark the first connection as done for the current user,
+        e.g. after showing an onboarding message.
+        """
+        if request.user.is_first_connection:
+            request.user.is_first_connection = False
+            request.user.save(update_fields=["is_first_connection"])
```
```suggestion
            request.user.save(update_fields=["is_first_connection", "updated_at"])
```

src/backend/core/migrations/0030_user_is_first_connection.py
lunika: Agree. We should run a migration for the existing users.

PR #1949, opened 2026-03-05
🐛(frontend) fix zIndex table of content

AntoLC:
## Purpose

When we open a modal, the table of content is above the modal, which is not the expected behavior. This commit fixes the zIndex of the table of content to be below the modal.

## Demo problem

image


Ovgodd (approved):
good 👍

PR #1952, opened 2026-03-06
✨(frontend) add top parent on sub docs search

### LINKED ISSUE #677: Display sub-docs in search results
For the multipage update, the search will evolve to not only (1) be able to search for documents (as it does today), but also **(2) be able to search for sub-pages within their context**. This means that under the sub-page, **the parent document is displayed** (see Figure below). The displayed parent document is the highest page accessible to the user in the tree. The backend must therefore be able to return it.

Image

[Figma Prototype (FR)](https://www.figma.com/design/qdCWR4tTUr7vQSecEjCyqO/Docs?node-id=9124-23749&t=KnchDogt64iF3Uie-0)

AntoLC:
## Purpose

In the search modal, under the sub docs, we now display the top parent of the doc, to give more context to the user about the doc they are looking for.
We refactorize the filters to get more flexibility and avoid too much props drilling.

## Demo

https://github.com/user-attachments/assets/a506e51a-3a9f-4a2c-978f-f413d48a177d


src/frontend/apps/impress/src/features/docs/doc-search/components/DocSearchFilters.tsx
```diff
@@ -1,73 +1,59 @@
...
-      )}
+      <Switch
+        labelSide="right"
+        label={t('All docs')}
+        checked={filter === 'all'}
+        onChange={() => setFilter(filter === 'all' ? 'current' : 'all')}
+        aria-label={t(
+          'Toggle to search in all documents or only in current document',
+        )}
+      />
```
Ovgodd, line 62:
```suggestion
  const { setFilter, filter } = useDocSearchFilterStore();
  
  const isAll = filter === 'all';
  const toggle = () => setFilter(isAll ? 'current' : 'all');
  
  const handleKeyDown = (e: React.KeyboardEvent) => {
  if (e.key === ' ' || e.key === 'Enter') {
    e.preventDefault();
    toggle();
  }
};

return (
    /**
     * The switch is not focusable, so we wrap it in a div that can be focused
     * and handle the keydown event to toggle the switch with
     * space key for accessibility reasons
     */
<Box
      role="switch"
      aria-checked={isAll}
      aria-label={t('Search in all documents')}
      tabIndex={0}
      onKeyDown={handleKeyDown}
      onClick={toggle}
      $css={css`
      cursor: pointer;
        &:focus-visible {
          outline: none;

          .c__switch__rail {
            outline: none;
            box-shadow: 0 0 0 2px
              var(--c--contextuals--border--semantic--brand--primary);
          }
        }
        // Remove the default focus style of the switch component
        .c__checkbox:focus-within {
          border: none;
          box-shadow: none;
          outline: 0;
        }
        & *:focus-visible {
          outline: none;
        }
      `}
>
      <Switch
        labelSide="right"
        label={t('All docs')}
        checked={isAll}
        onChange={toggle}
        aria-hidden={true}
        tabIndex={-1}
      />
```

Right now, the screen reader announces: "Search results controls, group, clickable, All docs, Toggle to search in all documents or only in current document, checkbox, unchecked",
that's quite verbose, it uses the wrong role (checkbox instead of switch), and both label and aria-label are read out which creates double labeling.

With this suggestion , the SR  announces: "Search in all documents, switch, on/off", one label, the correct role, and the right state.
For this we add` role="switch" `+ `aria-checked` +` aria-label` on the focusable wrapper, and hide the inner Cunningham Switch from screen readers with aria-hidden + tabIndex={-1}.

src/frontend/apps/impress/src/features/docs/doc-search/components/DocSearchModal.tsx
```diff
@@ -44,77 +52,110 @@ const DocSearchModalGlobal = ({
...
-                  values={filters}
-                  onValuesChange={setFilters}
-                  onReset={handleResetFilters}
+            <Box
+              $margin={{ vertical: 'sm', horizontal: 'base' }}
+              $justify="space-between"
+              $direction="row"
+              $align="center"
+              role="group"
+              aria-label={t('Search results controls')}
```
Ovgodd:
```suggestion
              $justify="space-between"
              $direction="row"
              $align="center"
```

The `role="group"` +` aria-label="Search results controls" `here makes the screen reader say "Search results controls, group" before announcing the switch. Since the switch already has its own role and aria-label, this extra announcement is a bit noisy

src/frontend/apps/impress/src/features/docs/doc-search/components/DocSearchModal.tsx
```diff
@@ -44,77 +52,110 @@ const DocSearchModalGlobal = ({
...
+  const { filter, setFilter } = useDocSearchFilterStore();
+
+  useEffect(() => {
+    if (!search) {
+      setResults([]);
+    }
+  }, [search]);
+
+  useEffect(() => {
+    setFilter(defaultFilters);
```
Ovgodd, line 69:
```suggestion
    setFilter(defaultFilters);
      return () => setFilter('all');
```

Should we cleanup the filter ? if the user opens the modal, toggle the filter, then closes and reposens on the same doc, `defaultFilter `don't change so the store will keep the old value ?

src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-inline-content/Interlinking/SearchPage.tsx
```diff
@@ -55,6 +56,16 @@ export const SearchPage = ({
...
+  const { setFilter } = useDocSearchFilterStore();
+
+  /**
+   * When the search page is opened, we set the search
+   * target to 'current' to limit the search to the current
+   * document and its sub-documents.
+   */
+  useEffect(() => {
+    setFilter('current');
+  }, [setFilter]);
```
Ovgodd, line 68:
```suggestion
  useEffect(() => {
    setFilter('current');
    return () => setFilter('all')
  }, [setFilter]);
```

Should we cleanup here the filter ? . If the user closes the interlinking and then opens the search modal, the filter will still be not reset ?

src/frontend/apps/impress/src/features/docs/docs-grid/components/DocMoveModal.tsx
```diff
@@ -158,41 +173,17 @@ export const DocMoveModal = ({
...
-          <Box>
+          <>
+            <Text as="h2" $margin="0" $size="s" $align="flex-start">
+              {t('Choose a new parent doc')}
+            </Text>
             <Box $position="absolute" $css="top: 4px; right: 4px;">
               <ButtonCloseModal
-                aria-label={t('Close the move modal')}
-                onClick={() => onClose()}
+                aria-label={t('Close the search modal')}
```
Ovgodd:
```suggestion
                aria-label={t('Close the move modal')}
```

we are in `DocMoveModal`, it is intended ?

Ovgodd (changes requested):
It works well on my side, just a few suggestions in terms of accessibility and readability.

src/frontend/apps/impress/src/features/docs/doc-search/components/DocSearchModal.tsx
AntoLC: I don't think it keeps the old filter, try again maybe ? 
I don't clean it because depends some states of the doc, the default filter is different: 
https://github.com/suitenumerique/docs/blob/e2caf4fff72bb8d5cc57ff6ef780d8d44f2a8eba/src/frontend/apps/impress/src/features/docs/doc-search/components/DocSearchModal.tsx#L206

src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-inline-content/Interlinking/SearchPage.tsx
AntoLC: There is not a static default filter, depend the state of the doc, the default change, so we force onMount the default filter that we want to use.

src/frontend/apps/impress/src/features/docs/doc-search/components/DocSearchModal.tsx
Ovgodd: OK I see ! thanks for the explanations

src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-inline-content/Interlinking/SearchPage.tsx
Ovgodd: OK that's understandable

Ovgodd (approved):

PR #1956, opened 2026-03-08
✨(frontend) Optionally include title in export

### LINKED ISSUE #1837: Export title document
## Feature Request

When exporting a document to PDF, only the content is exported, and not the title of the document. 

Being able to export the title (as an h1) before the content would be very useful. 

It should at least be an option (maybe we don't want to force it to every user, if some of them do not consider the title to be content ?)

Wqrld:
## Purpose

This commit introduces a checkbox in the export menu for inserting the document title.

On modal open, we check if the document contains an h1 and only precheck the checkbox if none is found. If the user overrides this checkbox, this is saved on a per-document basis.

Resolves https://github.com/suitenumerique/docs/issues/1837

image

## Functionality
| Format                  | `withTitle` | Title added?                              |
|-------------------------|-------------|-------------------------------------------|
| PDF / DOCX / ODT / HTML | `true`      | Yes   |
| PDF / DOCX / ODT / HTML | `false`     | No                                        |
| Print                   |  `true`/`false`       | No |

## Discussion
Print uses a slightly different exporter which does not have a neat way of prepending a title so we just hide the checkbox there.

DIsclaimer: AI was used in the creation of this code. The solution has been tested by a human.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to
 ### Added
 
 - ✨(helm) allow all keys in configMap as env var #1872
+- ✨(frontend) Optionally include title in export #1837
 
 ### Changed
 
--- a/src/frontend/apps/impress/src/features/docs/doc-export/components/ModalExport.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-export/components/ModalExport.tsx
@@ -1,8 +1,10 @@
+import { SpecificBlock } from '@blocknote/core';
 import { DOCXExporter } from '@blocknote/xl-docx-exporter';
 import { ODTExporter } from '@blocknote/xl-odt-exporter';
 import { PDFExporter } from '@blocknote/xl-pdf-exporter';
 import {
   Button,
+  Checkbox,
   Loader,
   Modal,
   ModalSize,
@@ -20,9 +22,15 @@ import { css } from 'styled-components';
 
 import { Box, ButtonCloseModal, Text } from '@/components';
 import { useMediaUrl } from '@/core';
-import { useEditorStore } from '@/docs/doc-editor';
+import {
+  DocsBlockSchema,
+  DocsInlineContentSchema,
+  DocsStyleSchema,
+  useEditorStore,
+} from '@/docs/doc-editor';
 import { Doc, useTrans } from '@/docs/doc-management';
 import { fallbackLng } from '@/i18n/config';
+import { safeLocalStorage } from '@/utils/storages';
 
 import { exportCorsResolveFileUrl } from '../api/exportResolveFileUrl';
 import { docxDocsSchemaMappings } from '../mappingDocx';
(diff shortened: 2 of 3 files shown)


virgile-dev:
Thanks for your contribution @Wqrld !
I like that checks if the fist block is a H1.
Nevertheless, I have doubts about including an element in the export that's not actually in the document content.
@rl-83 what do you think ?

Wqrld:
> Thanks for your contribution @Wqrld ! I like that checks if the fist block is a H1. Nevertheless, I have doubts about including an element in the export that's not actually in the document content. @rl-83 what do you think ?

I had the same doubts for a while, but given the current editor UI (with the title being right ontop of the editor, resulting in having the title twice right above eachother in many documents) this seems like what most people would expect. That is also what i got from the linked issue and some discussion on matrix. Given the automatic title detection it should not be intrusive to any workflows and only be an improvement.

If this path is not taken, I would choose to refactor the UI so the title is further away (a la google docs); but I obviously prefer this PR :).

robin-lecomte:
My UX suggestion is not to include a checkbox and to always display the page title, exactly as Notion does.
- Title is then displayed as an h1 title in the export
- h1, h2, h3 inside the doc become respectively h2, h3, h4 in the export

Wqrld:
It sounds like this would remove the option of customizing the title, which some people still want to do. 

Ex: We often have the date and the logo at the top of the document, and then a centered title. But other documents are more basic and would just benefit from having a simple title in the export

Wqrld:
@robin-lecomte 
To visualise the issue, here's two documents of mine.

Custom title (Do not want automatic insert):
image

Double title (Do want automatic insert)
image

virgile-dev:
@robin-lecomte just checked Notion doesn't covert H1 into h2 etc.
I also noticed that their document outline isn't available at export.
I think it's sketchy to restructure the document titles at export.
I'm still hesistant to add in the export content that's not in the document.
Google Docs and Word don't do that for example.
It's a recurring ask from user, but they have an easy way to add their title, I'm unsure it's worth the added complexity.

Wqrld:
I get the hesitance. I think the comparison with GDocs and Word is unfair in this case though, as in their UI the Document title is much further away from the content itself, so you don't have the same double title issue.

Unless there is a redesign planned that moves the title further away I think it's a necessary evil. Anyways multiple that i've shown docs to have called this out as "weird". 

> I think it's sketchy to restructure the document titles at export.

This commit doesn't. It just puts the title literally into the document with the same accessibility settings.

> I'm still hesistant to add in the export content that's not in the document.

Don't think that i agree, but the hesitation is fair.

> but they have an easy way to add their title,

Resulting in a double title, which is quite bad UI/UX

> I'm unsure it's worth the added complexity.

From a code perspective the change is relatively simple. 

From a UX perspective it is definitely something to think about. I am of the belief that you cannot get away with not putting any export options in (I have had people ask for page numbers aswell). I think the current feature of automatically toggling the checkbox based on if a title is detected makes for a good mid-way, but i'd be happy to discuss this further.

virgile-dev:
> My UX suggestion is not to include a checkbox and to always display the page title, exactly as Notion does.
> 
>     * Title is then displayed as an h1 title in the export
> 
>     * h1, h2, h3 inside the doc become respectively h2, h3, h4 in the export

Hey @robin-lecomte 
I gave some more thoughts to this. I think that @Wqrld is a better approach. Some people do want to use a custom title to their page. We even have a template to create an  [official header](https://docs.numerique.gouv.fr/docs/9bcfc396-a570-4c62-a2f8-1b543522edae/). So I'd be okay with a tick box that pops if it detects no titles at the top (h1 or h2 btw).
If you agree, we should validate the tickbox designed proposed here. 
Also should we consider only PDF exports or Docx and ODT. 
Cheers,

PR #1965, opened 2026-03-10
🐛(backend) enforce emoji validation for reactions

maboukerfa:
## Purpose

This pull request enforces emoji validation in the `ReactionSerializer` to prevent invalid emoji submissions.

- Currently we can do such reactions
Screenshot 2026-03-10 at 07 09 24

## Proposal

- Validate emojis in `ReactionSerializer` (previously accepted any string)
- Prevents multiple emojis or text uploads in a single reaction

### DIFF
--- a/src/backend/core/api/serializers.py
+++ b/src/backend/core/api/serializers.py
@@ -13,6 +13,7 @@ from django.utils.functional import lazy
 from django.utils.text import slugify
 from django.utils.translation import gettext_lazy as _
 
+import emoji
 import magic
 from rest_framework import serializers
 
@@ -874,6 +875,12 @@ class ReactionSerializer(serializers.ModelSerializer):
             "users",
         ]
         read_only_fields = ["id", "created_at", "users"]
+
+    def validate_emoji(self, value):
+        """Ensure the reaction is a single emoji."""
+        if not emoji.is_emoji(value):
+            raise serializers.ValidationError("Reaction must be a single valid emoji.")
+        return value
 
 
 class CommentSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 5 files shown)


src/backend/core/api/serializers.py
```diff
@@ -882,6 +883,12 @@ class Meta:
         ]
         read_only_fields = ["id", "created_at", "users"]
 
+    def validate_emoji(self, value):
+        """Ensure the reaction is a single emoji."""
+        if not emoji.is_emoji(value):
```
lunika, line 881: Can you please add a warning log here? To ensure that valid emoji are not rejected?

lunika (commented):
Thanks a lot for your contribution.

Just one comment, and it will be good.

src/backend/core/api/serializers.py
maboukerfa: I added a warning log to for the rejected emojis and their corresponding Unicode

Image

maboukerfa:
Hey @lunika, any updates on this?

lunika:
> Hey @lunika, any updates on this?

Thank you for your patience.
We are in a stabilization period; we try not to add new features before version 4.9 or 5.0
We will consider merging this kind of PR once this period ended, probably at the middle of april.

About emoji, we would like to localize them. Will your validation still work once it's made?

Thanks.

maboukerfa:
Alright, thanks for the update!

Yes, the validation should still work as long as the localization is handled on the front end. On the backend, I’m using the Unicode encoding for emojis, so it remains consistent regardless of how they’re implemented in the front.

I'll check it again once the localization implemented

lunika (approved):

PR #1967, opened 2026-03-10
💚(CI) immutable error not blocking

AntoLC:
## Purpose

We have cases where a workflow can fails in a middle step, if it happens, we were not able to retry the workflow, because of immutable images error. 
This commit makes the immutable error not blocking, if the workflow fails, we can retry it without the need to change the image tag.


AntoLC:
Let's close it for the moment, if the problem happens again we should maybe considerer it again.
The problem flow: https://github.com/suitenumerique/docs/actions/runs/22903876757/job/66457147885

PR #1971, opened 2026-03-10
Fix errors related to users onboarding feature

lunika:
## Purpose

In version 4.6.0 we introduce an onboarding feature when a new user is created. 
A sandbox document is created for the user
And for onboarding documents accesses were created.

On the sandbox creation we have a race condition when more than one user is created at the same time. This an issue we already had on the document creation, so we fixed it using the same solution : a transaction and a lock on the document table.

For the onboarding documents, we don't create accesses anymore; we use the link_trace table instead. So the documents must not be restricted.

## Proposal

- [x] 🐛(backend) create a link_trace record for onboarded documents
- [x] 🐛(backend) manage race condition when creating sandbox document


src/backend/core/tests/test_models_invitations.py
```diff
@@ -79,7 +79,7 @@ def test_models_invitations_is_expired():
         assert expired_invitation.is_expired is True
 
 
-def test_models_invitationd_new_userd_convert_invitations_to_accesses():
+def test_models_invitations_new_userd_convert_invitations_to_accesses():
```
jmaupetit: 😛

src/backend/core/tests/commands/test_clean_onboarding_documents_accesses.py
```diff
@@ -0,0 +1,129 @@
...
+
+import pytest
+
+from core import factories, models
+
+pytestmark = pytest.mark.django_db
+
+
+def test_clean_onboarding_documents_accesses(settings):
+    """test running the command remove the accesses for non privileged roles."""
```
jmaupetit:
```suggestion
    """Test running the command removes the accesses for non privileged roles."""
```

src/backend/core/tests/commands/test_clean_onboarding_documents_accesses.py
```diff
@@ -0,0 +1,129 @@
...
+    assert (
+        models.DocumentAccess.objects.filter(
+            document__in=non_onboarding_documents, role__in=models.PRIVILEGED_ROLES
+        ).count()
+        == 6
+    )
+
+
+def test_clean_onboarding_documents_accesses_no_onboarding_documents(settings):
+    """test running the command without onboarding documents should stop it."""
```
jmaupetit:
```suggestion
    """Test running the command without on-boarding documents is stopping it."""
```

CHANGELOG.md
```diff
@@ -6,6 +6,15 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
+- ✨(backend) add management command cleaning onboarding document accesses
+
+### Fixed
+
+- 🐛(backend) create a link_trace record for onboarded documents
```
jmaupetit:
```suggestion
- 🐛(backend) create a link_trace record for on-boarding documents
```

src/backend/core/migrations/0031_clean_onboarding_accesses.py
```diff
@@ -0,0 +1,34 @@
+# Generated by Django 5.2.12 on 2026-03-11 17:16
+
+from django.conf import settings
+from django.db import migrations
+
+from core.models import PRIVILEGED_ROLES
+
+
+def clean_onboarding_accesses(apps, schema_editor):
+    """clean accesses on onboarding documents."""
```
jmaupetit:
```suggestion
    """Clean accesses on on-boarding documents."""
```

src/backend/core/models.py
jmaupetit: Maybe it would be more explicit as we don't create accesses anymore?

```suggestion
            onboarding_link_traces = []
```

jmaupetit:
```suggestion
        create link traces to these documents and pin them as favorites for the user.
```

src/backend/core/models.py
```diff
@@ -261,29 +265,39 @@ def _duplicate_onboarding_sandbox_document(self):
...
+                    )
 
-            except Document.DoesNotExist:
-                logger.warning(
-                    "Onboarding sandbox document with id %s does not exist. Skipping.",
-                    sandbox_id,
+                sandbox_id = settings.USER_ONBOARDING_SANDBOX_DOCUMENT
+                try:
+                    template_document = Document.objects.get(id=sandbox_id)
+
```
jmaupetit: Remove unnecessary empty row:

```suggestion
                try:
                    template_document = Document.objects.get(id=sandbox_id)
```

src/backend/core/tests/test_models_users.py
```diff
@@ -166,24 +165,34 @@ def test_models_users_handle_onboarding_documents_access_with_invalid_document_i
...
 
-    user_accesses = models.DocumentAccess.objects.filter(user=user, document=document)
+    link_traces = models.LinkTrace.objects.filter(user=user, document=document)
+
+    assert link_traces.count() == 1
 
-    assert user_accesses.count() >= 1
+
+def test_models_users_handle_onboarding_documents_on_restricted_document_is_not_allowed():
+    """Onboarding document can be used when restricted"""
```
jmaupetit:
```suggestion
    """On-boarding document can be used when restricted."""
```

src/backend/core/tests/test_models_users.py
```diff
@@ -284,11 +295,37 @@ def test_models_users_duplicate_onboarding_sandbox_document_integration_with_oth
...
 
-    sandbox_access = user_accesses.get(document=sandbox_doc)
-    onboarding_access = user_accesses.get(document=onboarding_doc)
+        user1 = future1.result()
+        user2 = future2.result()
 
-    assert sandbox_access.role == models.RoleChoices.OWNER
-    assert onboarding_access.role == models.RoleChoices.READER
+        assert isinstance(user1, models.User)
+        assert isinstance(user2, models.User)
```
jmaupetit, line 331: Not sure you can really test this race condition with only 2 workers and a function that takes IMO less than 1ms. Have you tried increasing it to let say 100?

src/backend/core/tests/test_models_invitations.py
lunika: userd is a new daemon managed by systemd

src/backend/core/tests/test_models_users.py
lunika: 100 what ? workers ?

lunika: and if the number of worker is related t the number of cpu available, I think on the CI we can't use more than 2.

jmaupetit: As discussed IRL, since the test fails without this patch, we can consider it as relevant. Increasing the number of documents would be useless.

PR #1978, opened 2026-03-10
✨(backend) add limit on distinct reactions per comment

maboukerfa:
## Purpose

The current implementation allows users to add an unlimited number of reactions per comment.
Because the `/threads` endpoint is called on document load, having multiple comments with
thousands of reactions each could significantly impact page load performance.

Screenshot 2026-03-11 at 00 16 56

## Proposal

Implement a configurable limit (default: 15) on the number of distinct emoji reactions per comment.

- Backend validation ensures the limit cannot be exceeded via API
- Frontend disables reaction buttons when limit is reached

https://github.com/user-attachments/assets/321f2ed4-a2b9-44aa-b129-fcf71a182b2b

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -3234,9 +3234,11 @@ class CommentViewSet(
     permission_classes = [permissions.CommentPermission]
     pagination_class = Pagination
     serializer_class = serializers.CommentSerializer
-    queryset = models.Comment.objects.select_related("user").prefetch_related(
-        "reactions__users"
-    ).all()
+    queryset = (
+        models.Comment.objects.select_related("user")
+        .prefetch_related("reactions__users")
+        .all()
+    )
 
     def get_queryset(self):
         """Override to filter on related resource."""
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -3088,6 +3088,7 @@ class ConfigView(drf.views.APIView):
             "POSTHOG_HOST",
             "LANGUAGES",
             "LANGUAGE_CODE",
+            "REACTIONS_MAX_PER_COMMENT",
             "SENTRY_DSN",
             "TRASHBIN_CUTOFF_DAYS",
         ]
@@ -3233,7 +3234,9 @@ class CommentViewSet(
     permission_classes = [permissions.CommentPermission]
     pagination_class = Pagination
     serializer_class = serializers.CommentSerializer
-    queryset = models.Comment.objects.select_related("user").all()
+    queryset = models.Comment.objects.select_related("user").prefetch_related(
+        "reactions__users"
+    ).all()
 
     def get_queryset(self):
         """Override to filter on related resource."""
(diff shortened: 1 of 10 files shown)


src/backend/core/api/viewsets.py
```diff
@@ -2756,9 +2757,29 @@ def reactions(self, request, *args, **kwargs):
         serializer.is_valid(raise_exception=True)
 
         if request.method == "POST":
+            emoji = serializer.validated_data["emoji"]
+
+            if (
+                not models.Reaction.objects.filter(
+                    comment=comment, emoji=emoji
+                ).exists()
+                and comment.reactions.count() >= settings.REACTIONS_MAX_PER_COMMENT
```
lunika, line 3293: You should probably add `reactions` in the `select_related` objects in the `queryset` property to avoid `N+1` query.

src/backend/core/tests/documents/test_api_documents_comments.py
```diff
@@ -876,3 +876,56 @@ def test_delete_reaction_owned_by_the_current_user():
 
     reaction.refresh_from_db()
     assert reaction.users.exists()
+
+
+def test_create_reaction_exceeds_maximum():
+    """
+    Users should not be able to add more than REACTIONS_MAX_PER_COMMENT
+    (here we set it to 10) distinct emoji reactions to a comment.
```
lunika, line 965: I don't see where you set it to `10`.

lunika (commented):
I only focus on the backend part.

If possible, can you split in 2 commits ? A first one for the backend and a second one for the frontend ?

Thank you so much for your contribution

src/backend/core/api/viewsets.py
maboukerfa: I changed it to  

```python
queryset = (
    models.Comment.objects.select_related("user")
    .prefetch_related("reactions__users")
    .all()
)
```  

This loads each comment’s `user`, its related `reactions`, and the `users` who reacted.

src/backend/core/tests/documents/test_api_documents_comments.py
maboukerfa: It was hardcoded. I’ve imported `settings` in the test files and used the `REACTIONS_MAX_PER_COMMENT` var:

```python
  max_reactions = settings.REACTIONS_MAX_PER_COMMENT
```

src/backend/core/tests/documents/test_api_documents_comments.py
```diff
@@ -7,6 +7,8 @@
 import pytest
 from rest_framework.test import APIClient
 
+from django.conf import settings
```
lunika: We try to not import settings in the tests. We instead use the [setting fixture](https://pytest-django.readthedocs.io/en/latest/helpers.html#settings) in the tests where we want to control the settings.

src/backend/core/tests/documents/test_api_documents_comments.py
```diff
@@ -876,3 +878,56 @@ def test_delete_reaction_owned_by_the_current_user():
 
     reaction.refresh_from_db()
     assert reaction.users.exists()
+
+
+def test_create_reaction_exceeds_maximum():
```
lunika:
```suggestion
def test_create_reaction_exceeds_maximum(settings):
```

according to my previous comment

maboukerfa: fixed in [0356763](https://github.com/suitenumerique/docs/commit/03567637a8652b706b0188bc6d18e52c534d8cee)

src/backend/core/api/viewsets.py
```diff
@@ -2776,9 +2779,29 @@ def reactions(self, request, *args, **kwargs):
...
+                        ]
+                    },
+                    status=status.HTTP_400_BAD_REQUEST,
+                )
+
             reaction, created = models.Reaction.objects.get_or_create(
                 comment=comment,
-                emoji=serializer.validated_data["emoji"],
+                emoji=emoji,
             )
```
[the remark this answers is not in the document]
maboukerfa: @lunika, this is prone to race condition, a user can create a MAX+1 emoji if he spams emojis, is it worth fixing or it's an overkill ?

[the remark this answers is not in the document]
lunika: I think it's overkill and I'm not ok to add a lock on the table.

lunika (approved):

Issue #1973, opened 2026-03-10
Waffle label and decorative svg


cyberbaloo:
## Bug Report

**Problematic behavior**
The waffle label is not explicit. It contains the following text in the attribute `aria-label:"Digital LaSuite Services"`. It is not translated into French.
The `svg` element of this button should not be rendered.

**Expected behavior/code**
Add a more explicit label such as: "LaSuite Numérique services".
Add an `aria-hidden="true"` attribute to the `svg` element of the waffle.

**Additional context/Screenshots**
<img width="213" height="240" alt="Image" src="https://github.com/user-attachments/assets/bbda2517-3ebd-4fb8-b327-78df9e0de025" />

**Critères RGAA**

[1.2 Chaque image de décoration est-elle correctement ignorée par les technologies d’assistance ?](https://accessibilite.numerique.gouv.fr/methode/criteres-et-tests/#1.2)

[6.1 Chaque lien est-il explicite (hors cas particuliers) ?](https://accessibilite.numerique.gouv.fr/methode/criteres-et-tests/#6.1)

**Impact**
Decorative images should not be rendered to screen readers. This generates "noise" and can also slow down navigation by rendering useless information.
This can also add cognitive overload for people who use speech synthesis.

The importance of explicit links is to facilitate navigation for the user so that they can immediately know what the link corresponds to and the action associated with it. Some will navigate from link to link and others from heading to heading. Making it explicit saves the user time, but also reduces the cognitive load for others.

PR #1979, opened 2026-03-11
♿(frontend) localize LaGaufre label fallback in Docs

Ovgodd:
## Purpose

Fix LaGaufre (waffle) label localization in Docs. The label stayed in English regardless of the selected language because Docs uses `CunninghamProvider` instead of the ui-kit provider, so `LaGaufreV2` internal translations were not applied. We now pass an explicit, translated label from Docs i18n.

[1973](https://github.com/suitenumerique/docs/issues/1973)

## Proposal

- [x] Pass `label={t('LaSuite digital services')}` in `Waffle.tsx` when `waffleConfig.label` is not provided
- [x] Add translations for "LaSuite digital services" in `translations.json` (en, fr, de, es)

### DIFF
--- a/src/frontend/apps/impress/src/features/header/components/Waffle.tsx
+++ b/src/frontend/apps/impress/src/features/header/components/Waffle.tsx
@@ -1,5 +1,6 @@
 import { LaGaufreV2, LaGaufreV2Props } from '@gouvfr-lasuite/ui-kit';
 import React from 'react';
+import { useTranslation } from 'react-i18next';
 import { css } from 'styled-components';
 
 import { Box } from '@/components';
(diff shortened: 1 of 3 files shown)


src/frontend/apps/impress/src/features/header/components/Waffle.tsx
```diff
@@ -42,7 +44,12 @@ export const Waffle = () => {
         }
       `}
     >
-      <LaGaufreV2Fixed {...waffleConfig} />
+      <div style={{ display: 'flex' }}>
+        <LaGaufreV2Fixed
+          {...waffleConfig}
+          label={waffleConfig.label ?? t('LaSuite digital services')}
+        />
+      </div>
```
AntoLC:
```suggestion
      <LaGaufreV2Fixed
          label={t('LaSuite digital services')}
          {...waffleConfig}
      />
```

PR #1983, opened 2026-03-11
♿(frontend) focus skip link on headings and skip grid dropzone

Ovgodd:
## Purpose

Improve keyboard and screen reader navigation by moving skip link focus to page
headings instead of the whole main region, and by keeping the docs grid
dropzone out of the keyboard focus order.

## Proposal

- [x] focus the skip link and route-change target on the first available page
      heading (`h1`, then `h2`) instead of `main`
- [x] remove the header logo heading semantics so each page keeps a single,
      meaningful main heading
- [x] update main page titles to use proper heading levels for the new focus
      target
- [x] prevent the docs grid dropzone container from receiving keyboard focus
- [x] Change Skiplink semantic from `button `to `a`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to
 - 💫(frontend) fix the help button to the bottom in tree #2073
 - ♿️(frontend) improve version history list accessibility #2033
 - ♿️(frontend) fix more options menu feedback for screen readers #2071
+- ♿(frontend) focus skip link on headings and skip grid dropzone #1983
 
 ## [v4.8.2] - 2026-03-19
 
@@ -68,7 +69,6 @@ and this project adheres to
 - ⬆️(frontend) upgrade Next.js to v16 #1980
 - ♿️(frontend) fix aria-label and landmark on document banner state #1986
 - 🌐(i18n) add "new window" translation key for waffle aria-label #1984
-- ♿(frontend) focus skip link on headings and skip grid dropzone #1983
 
 ### Fixed
 
--- a/src/frontend/apps/impress/src/components/SkipToContent.tsx
+++ b/src/frontend/apps/impress/src/components/SkipToContent.tsx
@@ -1,37 +1,16 @@
 import { Button } from '@gouvfr-lasuite/cunningham-react';
-import { useRouter } from 'next/router';
-import { useEffect, useState } from 'react';
+import { useState } from 'react';
 import { useTranslation } from 'react-i18next';
-import { css } from 'styled-components';
 
 import { Box } from '@/components';
 import { useCunninghamTheme } from '@/cunningham';
+import { MAIN_LAYOUT_ID } from '@/layouts/conf';
 import { focusMainContentStart } from '@/layouts/utils';
 
 export const SkipToContent = () => {
   const { t } = useTranslation();
-  const router = useRouter();
   const { spacingsTokens } = useCunninghamTheme();
   const [isVisible, setIsVisible] = useState(false);
-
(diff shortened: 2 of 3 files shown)


src/frontend/apps/impress/src/layouts/conf.ts
```diff
@@ -1 +1,37 @@
...
+  }
+
+  if (!focusTarget.hasAttribute('tabindex')) {
+    focusTarget.setAttribute('tabindex', '-1');
+  }
+
+  focusTarget.focus(options);
+
+  return focusTarget;
+};
```
AntoLC: Could we add that in a utils file instead (`src/layouts/utils.ts`) ?

Ovgodd: yes of course !

src/frontend/apps/impress/src/components/SkipToContent.tsx
```diff
@@ -33,46 +50,38 @@ export const SkipToContent = () => {
...
-            0 0 0 1px var(--c--globals--colors--white-000),
-            0 0 0 4px var(--c--contextuals--border--semantic--brand--primary);
-          border-radius: var(--c--globals--spacings--st);
+        & a {
+          outline: none;
         }
       `}
     >
-      <Button
+      <StyledSkipLink
```
AntoLC: The Cunningham Button has a property `href`, so you don't have to design again a button.
The onclick seems to be used only to smooth the scroll, could we leverage the css property `scroll-behavior: smooth;` instead ?

src/frontend/apps/impress/src/components/SkipToContent.tsx
```diff
@@ -1,20 +1,37 @@
...
+  }
+`;
 
 export const SkipToContent = () => {
   const { t } = useTranslation();
   const router = useRouter();
   const { spacingsTokens } = useCunninghamTheme();
   const [isVisible, setIsVisible] = useState(false);
 
-  // Reset focus after route change so first TAB goes to skip link
```
AntoLC, line 17: I think you can remove this `useEffect`, it seems to be in collision with: https://github.com/suitenumerique/docs/blob/0e5c9ed8347f4b96e02c7e3192563f8fc64cc15d/src/frontend/apps/impress/src/hooks/useRouteChangeCompleteFocus.tsx#L6

Ovgodd: oh yes !

src/frontend/apps/impress/src/components/SkipToContent.tsx
Ovgodd: Hey I'm now using the Cunningham Button with href, it is perfect. 

But, the onClick can't be fully removed,scroll-behavior: smooth only handles the scroll, but onClick is needed to focus the heading (h1/h2) instead of the container `<div>`. Without it, screen reader users won't land on the page title.

CHANGELOG.md
```diff
@@ -18,13 +18,15 @@ and this project adheres to
...
 - 🌐(i18n) add "new window" translation key for waffle aria-label #1984
+- ♿(frontend) focus skip link on headings and skip grid dropzone #1983
 
 ### Fixed
 
 - 🐛(backend) create a link_trace record for on-boarding documents
 - 🐛(backend) manage race condition when creating sandbox document
 - 🐛(frontend) fix flickering left panel #1989
 - ♿️(frontend) improve doc tree keyboard navigation #1981
+- ✨(backend) add a is_first_connection flag to the User model#1938
```
AntoLC: Does not concern this PR

PR #1985, opened 2026-03-12
🔧(backend) enable psycopg-pool allowing configuring min and max size

lunika:
## Purpose

We want the possibility to configure specific environment variables on
backend and celery deployment. Most of them are common but in the case
of the newly added settings DB_PSYCOPG_POOL_MIN_SIZE we want to
configure ot only on the backend deployment, not on the celery or with a
different value.

## Proposal

- [x] ➕(backend) install psycopg_pool
- [x] 🔧(backend) enable psycopg-pool allowing configuring min and max size
- [x] 🔧(helm) allow specific env var for the backend and celery deploy
- [x] 🔧(helm) update values.yaml annotations


src/helm/env.d/dev/values.impress.yaml.gotmpl
```diff
@@ -74,6 +74,12 @@ backend:
     Y_PROVIDER_API_BASE_URL: http://impress-docs-y-provider:443/api/
     Y_PROVIDER_API_KEY: my-secret
     CACHES_KEY_PREFIX: "{{ now | unixEpoch }}"
+  django:
+    envVars:
+      WEB_CONCURRENCY: 1
+      DB_PSYCOPG_POOL_MIN_SIZE: 10
```
jmaupetit, line 80: Isn't it a bit high for a default value? I would suggest 2 or 4 instead. WDYT?

lunika: This is not a default value. It is the value I want to set on the development environment used locally with tilt.
The default value is the default one from the psycopg-pool library and it's 4.

PR #2006, opened 2026-03-16
♿️(frontend) ensure doc title is h1 for accessibility

Ovgodd:
## Purpose

In production, the theme config has `withTitle: false`, which prevents the header from rendering the "Docs" `h1`. This leaves the page with no `h1` at all, breaking heading hierarchy for screen readers and accessibility audit tools (HeadingsMap).

## Proposal

- [x] Always render the `<Title headingLevel="h1">` in the header, using `sr-only` when `withTitle` is `false` so it remains accessible to screen readers without visual impact
- [x] Remove `aria-hidden="true"` from the `Title` component so the `h1` is exposed to assistive technologies
- [x] Add `className` prop support to the `Title` component

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to
 
 - 🙈(docker) add **/.next to .dockerignore #2034
 
+### Changed
+
+- ♿️(frontend) ensure doc title is h1 for accessibility #2006
 
 ## [v4.8.1] - 2026-03-17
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to
 - ⬆️(frontend) upgrade Next.js to v16 #1980
 - ♿️(frontend) fix aria-label and landmark on document banner state #1986
 - 🌐(i18n) add "new window" translation key for waffle aria-label #1984
+- ♿️(frontend) ensure doc title is h1 for accessibility #2006
```
AntoLC: It will have to be moved under "Unreleased".

PR #2007, opened 2026-03-16
♿️(frontend) fix share modal heading hierarchy and title

Ovgodd:
## Purpose

Fix two accessibility issues in the share modal:
1. The modal's accessible name was `[object Object]` and there was a duplicate `role="dialog"` causing confusion for screen readers.
2. The heading hierarchy was inconsistent: group names and "Link settings" were not structured as h2 under the h1 "Share the document", and some were hidden from assistive technologies via `aria-hidden`.

hierarchy

sharemodalname

## Proposal

- [x] Fix modal `aria-label` being `[object Object]` by passing an explicit string instead of relying on `title.toString()` on a ReactNode
- [x] Remove duplicate `role="dialog"` and  `aria-label` from the inner presentational container
- [x] Replace the `heading` prop of `Command.Group` (which rendered a div with `aria-hidden="true"`) with an explicit `<Text as="h2">` for group names (e.g. Shared with X users, Document owner, Pending invitations)
- [x] Render "Link settings" as h2 in `DocVisibility`
- [x] Ensure consistent h1 → h2 heading hierarchy in the share modal

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,12 @@ and this project adheres to
 
 ### Fixed
 
-- 🙈(docker) add **/.next to .dockerignore #2034
+- 🙈(docker) add \*\*/.next to .dockerignore #2034
 
 ### Changed
 
 - ♿️(frontend) ensure doc title is h1 for accessibility #2006
+- ♿️(frontend) fix share modal heading hierarchy #2007
 
 ## [v4.8.1] - 2026-03-17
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to
 - ⬆️(frontend) upgrade Next.js to v16 #1980
 - ♿️(frontend) fix aria-label and landmark on document banner state #1986
 - 🌐(i18n) add "new window" translation key for waffle aria-label #1984
+- ♿️(frontend) fix share modal heading hierarchy #2007
```
AntoLC: Under Unreleased.

PR #2017, opened 2026-03-16
♿️(frontend) add nb accesses in share button aria-label

Ovgodd:
## Purpose

The share button shows the number of users with access, but screen readers don't announce it.

## Proposal

- [x] Add `nb_accesses_direct` to the share button `aria-label` when the count is visible

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,12 @@ and this project adheres to
 ### Changed
 
 - ♿️(frontend) ensure doc title is h1 for accessibility #2006
+- ♿️(frontend) add nb accesses in share button aria-label #2017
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to
 - ⬆️(frontend) upgrade Next.js to v16 #1980
 - ♿️(frontend) fix aria-label and landmark on document banner state #1986
 - 🌐(i18n) add "new window" translation key for waffle aria-label #1984
+- ♿️(frontend) add nb accesses in share button aria-label #2017
```
AntoLC: Under "Unreleased".

PR #2030, opened 2026-03-17
♿️(frontend) fix waffle aria-label spacing for new-window links

Ovgodd:
## Purpose

Fix the aria-label for waffle links that open in a new window. Screen readers previously announced "Visionouvelle fenêtre" instead of "Visio (nouvelle fenêtre)".

## Proposal

- [x] Include the space and parentheses in the key


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ and this project adheres to
 - ⬆️(frontend) upgrade Next.js to v16 #1980
 - ♿️(frontend) fix aria-label and landmark on document banner state #1986
 - 🌐(i18n) add "new window" translation key for waffle aria-label #1984
+- 🌐(i18n) fix waffle aria-label spacing for new-window links #2030
```
AntoLC: Under "Unreleased"

PR #2033, opened 2026-03-17
♿️(frontend) improve version history modal accessibility

Ovgodd:
## Purpose

Fix accessibility issues in the version history list: each version button was announced as "version item" by screen readers, the focus was not visible when tabbing, and the HTML structure was invalid (`<li>` inside `<button>`). Also ensure the modal's accessible name is translated (e.g. "Historique des versions" in French instead of "version history modal" in English).

## Proposal

- [x] Replace static `aria-label="version item"` with a dynamic translated label including the version date
- [x] Add visible focus background on version items for keyboard navigation
- [x] Add `aria-pressed` and a live region to announce the selected version
- [x] Change `role="listbox"` to `role="list"` so Tab/Enter works on all items
- [x] Wrap each button in a `<li>` for valid `<ul> > <li> > <button>` structure
- [x] Replace `aria-label="version history modal"` with `aria-labelledby` pointing to the translated h1
- [x] Translate `aria-label="version list"` with i18n

### DIFF
--- a/src/frontend/apps/e2e/__tests__/app-impress/doc-version.spec.ts
+++ b/src/frontend/apps/e2e/__tests__/app-impress/doc-version.spec.ts
@@ -163,7 +163,7 @@ test.describe('Doc Version', () => {
 
     await expect(modal.getByText('World')).toBeHidden();
 
-    await page.getByRole('button', { name: 'Restore' }).click();
+    await page.getByRole('button', { name: 'Restore', exact: true }).click();
     await expect(page.getByText('Your current document will')).toBeVisible();
     await page.getByText('If a member is editing, his').click();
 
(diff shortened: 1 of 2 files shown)


src/frontend/apps/impress/src/features/docs/doc-versioning/components/VersionList.tsx
```diff
@@ -49,23 +50,41 @@ const VersionListState = ({
...
+                &:focus-visible {
+                  background: var(--c--globals--colors--gray-100);
+                  border-radius: var(--c--globals--spacings--st);
+                }
+              `}
+              className="version-item"
+              onClick={() => {
+                onSelectVersion?.(version.version_id);
+              }}
+            >
```
AntoLC: Can you add this part inside `VersionItem` ?

AntoLC (approved):

PR #2035, opened 2026-03-17
🔧(backend) add DB_PSYCOPG_POOL_ENABLED settings

lunika:
## Purpose

The psycopg pool config was enabled by default forcing its usage. Using psycopg pool can be difficult, finding the good configuration take time. By default its usage should be disable and the maintainer of the instance should decide to enable it or not.

## Proposal

- [x] 🔧(backend) add DB_PSYCOPG_POOL_ENABLED settings


src/backend/impress/settings.py
```diff
@@ -99,24 +99,6 @@ class Base(Configuration):
                 "localhost", environ_name="DB_HOST", environ_prefix=None
             ),
             "PORT": values.Value(5432, environ_name="DB_PORT", environ_prefix=None),
```
jmaupetit, line 101: I would add a comment here mentioning that database options may be configured in the `post_setup` class method. WDYT?

lunika: Done

Issue #2044, opened 2026-03-17
Search modal: focus loss after activating the "Reset" button


cyberbaloo:
## Bug Report

**Problematic behavior**
When navigating with a screen reader and after activating the "Reset" button, focus is lost and returns to the entire window instead of staying within the modal.

When navigating with the keyboard and activating the "Reset" button, there is no action.

**Expected behavior/code**
After activating the "Reset" button, focus must be repositioned on the input field of the search modal.

**Steps to Reproduce**
1. Open the document search modal.
2. Navigate with the keyboard to the "Reset" button and activate it.
3. Notice that focus leaves the modal and returns to the global window.

**Critères RGAA**

[Critère 12.8 : Dans chaque page web, l'ordre de tabulation est-il cohérent ?](https://accessibilite.numerique.gouv.fr/methode/criteres-et-tests/#12.8)

**Impact**

A user navigating with the keyboard activates the "Reset" button and loses navigation focus.

They must restart from the beginning of the page to return to the search field, which makes using the modal very difficult.

Issue #2043, opened 2026-03-17
Search modal: absence of vocalization of results


cyberbaloo:
## Bug Report

**Problematic behavior**
When the user enters a word in the search field, results are displayed in the suggestions list.
There is no feedback about the search results as long as the focus remains on the input field.

**Expected behavior/code**
Search results must be announced automatically, without the user having to leave the input field. This can be implemented via an aria-live attribute on the results container, or via a role="status" element announcing the number of available results.

**Steps to Reproduce**
1. Open the document search modal.
2. Place focus on the input field with a screen reader.
3. Enter a search term.
4. Notice that no update is announced automatically when the results appear.

**Possible Solution**
```
<!-- Option A : aria-live sur le conteneur des résultats -->
<div role="listbox" aria-live="polite" aria-label="Suggestions" ...>
  ...
</div>

<!-- Option B : élément status annonçant le nombre de résultats -->
<div role="status" aria-live="polite">3 résultats disponibles</div>
```

## Critères RGAA

[Critère 7.1 : Chaque script est-il, si nécessaire, compatible avec les technologies d'assistance ?](https://accessibilite.numerique.gouv.fr/methode/criteres-et-tests/#7.1)

## Impact

A blind user navigating with a screen reader enters a term in the search field. The results appear visually but are not announced. They do not know if their search has produced results and must leave the field to explore the list, which disrupts their navigation.

Issue #2042, opened 2026-03-17
Search modal: "Reset" button label not explicit


cyberbaloo:
## Bug Report

**Problematic behavior**
The "Reset" button does not have a contextual accessible name. Its label alone does not make it possible to understand what it resets. We do not know whether we are resetting the search field or the search area.

**Expected behavior/code**
The button must have an explicit accessible name that precisely indicates its action.

**Steps to Reproduce**
1. Open the document search modal.
2. Navigate with the keyboard to the "Reset" button.
3. Notice that the accessible name announced is "Reset", with no specification of what is being reset.

**Possible Solution**
Complete the `aria-label` attribute label

**Before:**
`<button class="c__button ...">Réinitialiser</button>`

**After:**
`<button class="c__button ..." aria-label="Réinitialiser les filtres de recherche">Réinitialiser</button>`

**Critères RGAA**
[Critère 7.1 : Chaque script est-il, si nécessaire, compatible avec les technologies d'assistance ?](https://accessibilite.numerique.gouv.fr/methode/criteres-et-tests/#7.1)

**Impact**
A blind user navigating with a screen reader reaches the "Reset" button without being able to identify what it resets. They risk activating it by mistake or not understanding its role in the search form.

Issue #2041, opened 2026-03-17
Search modal: empty label on input field


cyberbaloo:
## Bug Report

**Problematic behavior**
The `<label>` element associated with the quick search input field (via for="radix-_r_h6_") is present but does not contain any text.

**Expected behavior/code**
The `label` must contain a visible label or, alternatively, text that is visually hidden but rendered by assistive technologies.

**Steps to Reproduce**
1. Open the document search modal.
2. Inspect the DOM and locate the `<label>` associated with the search `<input>` (`for="radix-_r_h6_"`).
3. Notice that the `<label>` element is.

**Possible Solution**
**Before :**
`<label for="radix-_r_h6_" id="radix-_r_h5_" style="..."></label>`

**After :**
`<label for="radix-_r_h6_" id="radix-_r_h5_" style="...">Rechercher des documents</label>`

**Additional context/Screenshots**

<img width="794" height="93" alt="Image" src="https://github.com/user-attachments/assets/70ea6fc7-10c3-4a6f-977c-0f094c04e99d" />

## Critères RGAA

[Critère 11.1 : Chaque champ de formulaire a-t-il une étiquette ?](https://accessibilite.numerique.gouv.fr/methode/criteres-et-tests/#11.1)

## Impact

A blind user navigating with a screen reader reaches the search input field. The empty label does not allow them to identify the function of the field, which may prevent them from using the document search.

PR #2054, opened 2026-03-17
♿️(frontend) fix search modal accessibility issues

Ovgodd:
## Purpose

Improve accessibility in the document search modal by fixing several issues
reported around labels, announcements, focus management and decorative content.

## Proposal

- [x] set an empty `alt` attribute on the decorative empty-state image (`#2038`)
- [x] update `aria-expanded` dynamically on the search combobox (`#2039`)
- [x] provide a non-empty label to the search input through `QuickSearch` (`#2041`)
- [x] add an explicit accessible name to the reset button (`#2042`)
- [x] announce result count updates through a polite live region (`#2043`)
- [x] move filters out of the suggestions listbox and restore focus to the input after reset (`#2044`)

### DIFF
--- a/src/frontend/apps/e2e/__tests__/app-impress/doc-search.spec.ts
+++ b/src/frontend/apps/e2e/__tests__/app-impress/doc-search.spec.ts
@@ -29,7 +29,7 @@ test.describe('Document search', () => {
     await page.getByTestId('search-docs-button').click();
 
     await expect(
-      page.getByRole('img', { name: 'No active search' }),
+      page.getByLabel('Search modal').locator('img[alt=""]'),
     ).toBeVisible();
 
     await expect(
@@ -107,7 +107,7 @@ test.describe('Document search', () => {
 
     await searchButton.click();
     await expect(
-      page.getByRole('combobox', { name: 'Quick search input' }),
+      page.getByRole('combobox', { name: 'Search documents' }),
     ).toBeVisible();
     await expect(filters).toBeHidden();
 
@@ -120,7 +120,7 @@ test.describe('Document search', () => {
 
     await searchButton.click();
     await expect(
-      page.getByRole('combobox', { name: 'Quick search input' }),
+      page.getByRole('combobox', { name: 'Search documents' }),
     ).toBeVisible();
     await expect(filters).toBeHidden();
 
@@ -164,9 +164,9 @@ test.describe('Document search', () => {
     const searchButton = page.getByTestId('search-docs-button');
 
     await searchButton.click();
-    await page.getByRole('combobox', { name: 'Quick search input' }).click();
+    await page.getByRole('combobox', { name: 'Search documents' }).click();
     await page
-      .getByRole('combobox', { name: 'Quick search input' })
+      .getByRole('combobox', { name: 'Search documents' })
       .fill('sub page search');
 
     // Expect to find the first and second docs in the results list
@@ -188,7 +188,7 @@ test.describe('Document search', () => {
     );
     await searchButton.click();
     await page
-      .getByRole('combobox', { name: 'Quick search input' })
+      .getByRole('combobox', { name: 'Search documents' })
       .fill('second');
 
     // Now there is a sub page - expect to have the focus on the current doc
--- a/src/frontend/apps/impress/src/features/docs/doc-search/components/DocSearchModal.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-search/components/DocSearchModal.tsx
@@ -64,6 +64,7 @@ const DocSearchModalGlobal = ({
       aria-describedby="doc-search-modal-title"
     >
       <Box
+        aria-label={t('Search modal')}
         $direction="column"
         $justify="space-between"
         className="--docs--doc-search-modal"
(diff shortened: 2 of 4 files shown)


src/frontend/apps/e2e/__tests__/app-impress/doc-search.spec.ts
```diff
@@ -29,7 +29,7 @@ test.describe('Document search', () => {
     await page.getByTestId('search-docs-button').click();
 
     await expect(
-      page.getByRole('img', { name: 'No active search' }),
+      page.locator('img[src*="illustration-docs-empty"]'),
```
AntoLC: I don't understand these changes on the selectors, seems more solid particulary in a accessibility perspective to catch element by their role and label.

src/frontend/apps/e2e/__tests__/app-impress/doc-search.spec.ts
```diff
@@ -106,9 +106,7 @@ test.describe('Document search', () => {
     const filters = page.getByTestId('doc-search-filters');
 
     await searchButton.click();
-    await expect(
-      page.getByRole('combobox', { name: 'Quick search input' }),
-    ).toBeVisible();
+    await expect(page.getByTestId('quick-search-input')).toBeVisible();
```
AntoLC: Same reflexion as before, why testId instead of role? Role and label assert that the accessibility is correct.

src/frontend/apps/impress/src/components/quick-search/QuickSearchInput.tsx
```diff
@@ -23,9 +24,20 @@ export const QuickSearchInput = ({
...
   const { spacingsTokens } = useCunninghamTheme();
+  const internalRef = useRef<HTMLInputElement>(null);
+  const inputRef = externalRef ?? internalRef;
+
+  useEffect(() => {
+    inputRef.current?.setAttribute(
+      'aria-expanded',
+      String(isExpanded ?? false),
+    );
+  }, [isExpanded, inputRef]);
```
AntoLC: In what I understand, `cmdk` is managing already the accessibility of the input. Even adding the role combobox is not necessary, aria-expanded describes whether the popup element is displayed, not whether it contains results. When there are no results, the Command.List (the listbox) is still rendered and visible in the DOM — it's just empty, or showing a "no results" message. So aria-expanded should remain true.

src/frontend/apps/impress/src/features/docs/doc-search/components/DocSearchModal.tsx
```diff
@@ -133,6 +140,12 @@ const DocSearchModalGlobal = ({
                 }
               />
             )}
+            <Box role="status" aria-live="polite" className="sr-only">
```
AntoLC: Not better to use `@react-aria/live-announcer` ?

src/frontend/apps/e2e/__tests__/app-impress/doc-search.spec.ts
Ovgodd: Yes, I changed it because the test was failing, but you're right, that wasn't the best approach. I've replaced it a search by label

src/frontend/apps/e2e/__tests__/app-impress/doc-search.spec.ts
Ovgodd: sometimes, I just overthink

src/frontend/apps/impress/src/components/quick-search/QuickSearchInput.tsx
Ovgodd: You're right, the Command.List is always rendered in the DOM even when empty, so aria-expanded="true" is correct here. I think the issue assumed the listbox was hidden when there are no results, but as you said it's not. I'll revert this and explain in the issue why.

src/frontend/apps/impress/src/features/docs/doc-search/components/DocSearchModal.tsx
Ovgodd: Totally ! I think I didn't have it yet when I started this fix

src/frontend/apps/impress/src/features/docs/doc-search/components/DocSearchModal.tsx
```diff
@@ -51,8 +53,21 @@ const DocSearchModalGlobal = ({
 
   const handleResetFilters = () => {
     setFilters({});
+    inputRef.current?.focus();
```
AntoLC: We should leverage `useFocusStore`, we will avoid the props drilling.

src/frontend/apps/impress/src/features/docs/doc-search/components/DocSearchModal.tsx
```diff
@@ -51,8 +53,21 @@ const DocSearchModalGlobal = ({
...
+      if (!search) return;
+
+      announce(
+        t('{{count}} result(s) available', { count }),
+        'polite',
+      );
+    },
+    [search, t],
+  );
+
```
AntoLC: Why do we announce that here ? 
Could we announce it directly here ?
https://github.com/suitenumerique/docs/blob/be5dcb858a211dffe812dcecc1457fbb24b4705f/src/frontend/apps/impress/src/features/docs/doc-search/components/DocSearchContent.tsx#L98

We will not have to add props like that.

src/frontend/apps/impress/src/features/docs/doc-search/components/DocSearchModal.tsx
Ovgodd: yes I like it !

PR #2064, opened 2026-03-18
🚸(frontend) hint min char search users

AntoLC:
## Purpose

We give a hint to the user about the minimum number of characters required to perform a search in the quick search input of the doc share modal.
This is to improve the user experience.

## Demo

https://github.com/user-attachments/assets/36d7c1a1-e0ea-4d75-aa4b-6ccfd2efd298

### DIFF
--- a/src/frontend/apps/impress/src/features/docs/doc-share/components/DocShareModal.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-share/components/DocShareModal.tsx
@@ -289,7 +289,7 @@ export const DocShareModal = ({ doc, onClose, isRootDoc = true }: Props) => {
                     />
                   )}
                   {showMemberSection && isRootDoc && (
-                    <Box $padding={{ horizontal: 'base' }}>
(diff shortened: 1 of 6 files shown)


src/frontend/apps/impress/src/features/docs/doc-share/components/DocShareModal.tsx
```diff
@@ -335,14 +336,31 @@ interface QuickSearchInviteInputSectionProps {
...
+    }
+    if (isValidEmail(userQuery)) {
+      return t('Choose the email');
+    }
+    if (!searchUsersRawData?.length) {
+      return t('No results. Type a full email address to invite someone.');
+    }
+
+    return t('Choose a user');
+  }, [minLength, searchUsersRawData?.length, t, userQuery]);
```
Ovgodd, line 350: Can you add announce() from @react-aria/live-announcer so the search hint ("Type at least X characters...", "Choose a user", etc.) is announced by screen readers when it changes.

```suggestion
 useEffect(() => {
    announce(hint, 'polite');
  }, [hint]);
```

Issue #1826, opened 2026-01-23
Search is not available on public docs that contain subdocs


virgile-dev:
## Bug Report

**Problematic behavior**
i'm getting feedback from users that they're sharing public docs with a few sub-docs. 

**Expected behavior/code**
they'd like for their viewers to be able to search the subdocs just like they do when they are connected.

**Steps to Reproduce**
1. Go on https://docs.numerique.gouv.fr/docs/b7518303-3e21-458f-a29e-7adca9a3619d/
2. Make sure you're loggued out
3. See that there is no search icon

**Environment**
- Docs version: 4.4.0
- Instance url:

**Possible Solution**

**Additional context/Screenshots**
Add any other context about the problem here. If applicable, add screenshots to help explain.

AntoLC:
It is a backend problem at the base, the front follow the back:

https://github.com/suitenumerique/docs/blob/1083aac9202ce0506ccb838ae69479ea208f46fd/src/backend/core/tests/documents/test_api_documents_search.py#L34-L38

virgile-dev:
@lunika any reason we prevent search for un authenticated users on public docs with subdocs ? Product side this is not the expected behaviour.

virgile-dev:
As mentionned in today's weekly. @lunika and @Ash-Crow mentioned that this scoped search is going to require a dedicated endpoint based on UUID. Unauthenticated users can only search items in the tree associated to the document. (they cannot search for other public docs for ex).

We also need to consider @mascarpon3 this issue is going to require your attention.
We'll need `find` to be able to take a query with a public doc UIDD where results will be looked for only in the document tree and not elsewhere. Is this clear enough or should have a specific meeting about this ?

lunika:
The `search` endpoint behavior has evolved since the `find` integration has been made.
A new ability `search` is available and is True for public document. Combined with the `path` query string, an anonymous user will be available to search for documents in all the current tree.
So this is now in the front application that it should be implemented @AntoLC

PR #2068, opened 2026-03-18
✨(backend) add a public_search API view to the Document viewset

Ash-Crow:
## Purpose
This allows searching (on document titles only) on a public document and related public documents on the same tree.

A new get_highest_public_ancestor() helper method is also added to the Document model.

Fixes the backend part of #1826 

## Proposal
- [x] Add a `public_search()` method on the `DocumentViewSet`
- [x] Add a `get_highest_public_ancestor()` method on the `Document` model.


src/backend/core/api/viewsets.py
```diff
@@ -1532,6 +1535,47 @@ def _list_descendants(self, request, validated_data):
...
+        The filtering is done on the model field 'title', there is no full text search.
+
+        The ordering is always by the most recent first.
+        """
+        params = serializers.SearchDocumentSerializer(data=request.query_params)
+        params.is_valid(raise_exception=True)
+
+        text = params.validated_data["q"]
+
+        document = self.get_object()
```
lunika: It should be moved to the top of the method, to check permissions first

src/backend/core/api/viewsets.py
```diff
@@ -1532,6 +1535,47 @@ def _list_descendants(self, request, validated_data):
...
+            raise drf.exceptions.ValidationError(filterset.errors)
+
+        queryset = filterset.filter_queryset(queryset)
+        queryset = queryset.filter(ancestors_deleted_at__isnull=True)
+
+        return self.get_response_for_queryset(
+            queryset.order_by("-updated_at"),
+            context={
+                "request": request,
+            },
```
lunika: The `get_response_for_queryset` is already adding a context that contains more info than the `request`. Is it intended to not use this mechanism and limit only to the request ?

src/backend/core/tests/documents/test_api_documents_public_search.py
```diff
@@ -0,0 +1,275 @@
...
+
+    return (
+        RequestFactory()
+        .get(
+            f"/api/v1.0/documents/{document_id}/public_search/",
+            dict(sorted(params.items())),
+        )
+        .build_absolute_uri()
+    )
+
```
lunika: This is not used and I think it is not a test.

src/backend/core/tests/documents/test_api_documents_public_search.py
```diff
@@ -0,0 +1,275 @@
...
+    client = APIClient()
+    client.force_login(user)
+    response = client.get(
+        f"/api/v1.0/documents/{document.id}/public_search/",
+        data={"q": "anything"},
+    )
+    assert response.status_code == 403
+    assert response.json() == {
+        "detail": "You do not have permission to perform this action."
+    }
```
lunika, line 118: Add a test for an authenticated user on a document with a link_reach authenticated ?

src/backend/core/tests/documents/test_api_documents_public_search.py
```diff
@@ -0,0 +1,275 @@
...
+    other_root = factories.DocumentFactory(link_reach="public", title="other root")
+    outside = factories.DocumentFactory(parent=other_root, title="alpha outside")
+
+    response = client.get(
+        f"/api/v1.0/documents/{public_doc.id}/public_search/",
+        data={"q": "alpha"},
+    )
+    assert response.status_code == 200
+
+    result_ids = {r["id"] for r in response.json()["results"]}
```
lunika, line 201: In all your tests you should assert the lenght of the response or explicitely check every documents present. Here you don't know exactly what is returned. This comment is also valid for all your tests.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
+- ✨(backend) add a public_search API view to the Document viewset#2068
```
lunika:
```suggestion
- ✨(backend) add a public_search API view to the Document viewset #2068
```

src/backend/core/api/viewsets.py
Ash-Crow: Not sure why I added it, I'll remove it.

src/backend/core/tests/documents/test_api_documents_public_search.py
Ash-Crow: I forgot to remove this unused helper, yes.

lunika:
The search endpoint refactored when find was implemented already manage it.

PR #2069, opened 2026-03-18
♿️(frontend) improve language picker accessibility

Ovgodd:
## Purpose

Improve accessibility of the language picker for screen reader users.

## Proposal

- [x] Use `menuitemradio` and `aria-checked` to announce selected state (checked/unchecked)
- [x] Add `lang` attribute on language labels for correct pronunciation
- [x] Add `aria-live` region to announce language change when confirming selection

### DIFF
--- a/src/frontend/apps/e2e/__tests__/app-impress/utils-common.ts
+++ b/src/frontend/apps/e2e/__tests__/app-impress/utils-common.ts
@@ -8,16 +8,6 @@ import theme_customization from '../../../../../backend/impress/configuration/th
 export type BrowserName = 'chromium' | 'firefox' | 'webkit';
 export const BROWSERS: BrowserName[] = ['chromium', 'webkit', 'firefox'];
 
-/** Returns a locator for a menu item (handles both menuitem and menuitemradio roles) */
-export const getMenuItem = (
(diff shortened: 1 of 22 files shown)


src/frontend/apps/e2e/__tests__/app-impress/utils-common.ts
```diff
@@ -3,15 +3,18 @@ import path from 'path';
...
   context: Page | Locator,
   name: string,
   options?: { exact?: boolean },
 ): Locator =>
   context
     .getByRole('menuitem', { name, exact: options?.exact })
-    .or(context.getByRole('menuitemradio', { name, exact: options?.exact }));
+    .or(context.getByRole('menuitemradio', { name, exact: options?.exact }))
+    .or(
+      context.getByRole('menuitemcheckbox', { name, exact: options?.exact }),
```
AntoLC: I am not fan to get selectors like that, because we are not totally in control of what we are trying to get; when you use it, if in your context you have 1 `menuitem` and 1 `menuitemradio`, you will never be able to catch the `menuitemradio` because it will get `menuitem` first.

You don't seem to use it moreover.

AntoLC (approved):

PR #2070, opened 2026-03-18
♿️(frontend) announce formatting shortcuts for screen readers

Ovgodd:
## Purpose

Ensure that formatting shortcuts  are announced to screen readers (NVDA, VoiceOver), so users know their shortcut was applied.

https://github.com/user-attachments/assets/94db1192-c43c-43ef-be2e-b45ba17c2583

## Proposal

- [x] Announce actions via `@react-aria/live-announcer` (assertive)

### DIFF
--- a/src/frontend/apps/impress/src/features/docs/doc-editor/hook/useShortcuts.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-editor/hook/useShortcuts.tsx
@@ -1,11 +1,83 @@
-import { useEffect } from 'react';
+import { announce } from '@react-aria/live-announcer';
+import { useCallback, useEffect } from 'react';
+import { useTranslation } from 'react-i18next';
 
 import { DocsBlockNoteEditor } from '../types';
+
+const getFormattingShortcutLabel = (
+  event: KeyboardEvent,
+  t: (key: string) => string,
+): string | null => {
+  const isMod = event.ctrlKey || event.metaKey;
+  if (!isMod) {
+    return null;
+  }
+
+  if (event.altKey) {
+    switch (event.code) {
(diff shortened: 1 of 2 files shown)


src/frontend/apps/impress/src/features/docs/doc-editor/hook/useShortcuts.tsx
```diff
@@ -1,13 +1,86 @@
...
+    }
+  }
+
+  return null;
+};
+
 export const useShortcuts = (
   editor: DocsBlockNoteEditor,
-  el: HTMLDivElement | null,
+  refEditorContainer: React.RefObject<HTMLDivElement | null>,
```
AntoLC: Why are we changing the type ? ref does not retrigger a `useEffect` I think.

src/frontend/apps/impress/src/features/docs/doc-editor/hook/useShortcuts.tsx
```diff
@@ -1,13 +1,86 @@
...
+    [editor, t],
+  );
+
+  useEffect(() => {
+    document.addEventListener('keydown', handleFormattingShortcut, true);
+
+    return () => {
+      document.removeEventListener('keydown', handleFormattingShortcut, true);
+    };
+  }, [handleFormattingShortcut]);
```
AntoLC: Can we use the same pattern ?

```suggestion
  useEffect(() => {
    el.addEventListener('keydown', handleFormattingShortcut, true);

    return () => {
      el.removeEventListener('keydown', handleFormattingShortcut, true);
    };
  }, [handleFormattingShortcut]);
```

PR #2084, opened 2026-03-19
🐛(backend) stop using add_sibling method to create sandbox document

lunika:
## Purpose

In a past release we added a feature to create a sandbox document to a newly created user. To create this sandbox document, we duplicate an existing document and this duplicate is using the add_sibling method with the "right" agument on this original document. Adding a sibling at the right to a document involve moving right every root document created after the original document, so the path of all this documents are recalculated and changed. This can lead to the lost of some leaf in a tree because to do this operation, multiple locks are created on the database, creating lot of connection to the database and if the max number connection to the database is reached or if the memory allocated by the database is too hight, the database can close all connections leading to inconsistent paths in the Document table.

## Proposal

- [x] 🐛(backend) stop using add_sibling method to create sandbox document


virgile-dev:
Wow good job debugging this.

Issue #2087, opened 2026-03-19
Search endpoint does not work without OIDC refresh token


lunika:
## Bug Report

**Problematic behavior**

In PR #1834 an hybrid search has been merged, but by default the normal search should be available.
In fact, the search endpoint is now using the decorator `refresh_oidc_access_token`, without a valid refresh token the search is not usable

**Expected behavior/code**

Should work without refresh token

virgile-dev:
Do you mean, if no refresh token, then I just have the regular search ?

lunika:
you should have. But it is not the case. Without it you have a status code 401 (or 403 I don't remember).

The problem is this line https://github.com/suitenumerique/docs/blob/9991820cb14fed19953ad4cf801a000dbbcf306d/src/backend/core/api/viewsets.py#L1418

virgile-dev:
Thanks. @mascarpon3 can you please fix it ASAP ?

mascarpon3:
but users should be OIDC authenticated. why would it make a difference ? 

we are talking about non-authenticated users ?

virgile-dev:
A user just surfaced this feedback 
> Since this morning, I have been unable to link pages using the '/link a doc' command. Essentially, when I click on '/lier un doc', in the field where I can specify a page name, it indicates that no pages are available ('Aucun document trouvé' / 'No document found')—even though I have many pages, and this feature was working perfectly until now.

Logging out and loging in again  fixed it. This means that the interlink feature is using the same function as the search box.

Let's make sure we also fix this with the PR #2090

PR #2090, opened 2026-03-19
🚨(backend) fix search without refresh token

mascarpon3:
i remove the decorator on the search function
and refresh the token in the function so
i can try catch it.

## Purpose

fix of https://github.com/suitenumerique/docs/issues/2087

## Proposal

- [ ]  remove the refresh in the decorator
- [ ] try catch the refresh in the function

## External contributions

- [ ] I have read and followed the [contributing guidelines](https://github.com/suitenumerique/docs/blob/main/CONTRIBUTING.md)
- [ ] I have read and agreed to the [Code of Conduct](https://github.com/suitenumerique/docs/blob/main/CODE_OF_CONDUCT.md)
- [ ] I have signed off my commits with `git commit --signoff` (DCO compliance)
- [ ] I have signed my commits with my SSH or GPG key (`git commit -S`)
- [ ] My commit messages follow the required format: `<gitmoji>(type) title description`

### DIFF
--- a/src/backend/core/api/utils.py
+++ b/src/backend/core/api/utils.py
@@ -198,7 +198,7 @@ def refresh_access_token(session):
             "client_secret": settings.OIDC_RP_CLIENT_SECRET,
             "refresh_token": refresh_token,
         },
-        timeout=5,
+        timeout=settings.OIDC_TIMEOUT,
     )
     response.raise_for_status()
     token_info = response.json()
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -1429,15 +1429,6 @@ class DocumentViewSet(
         if search_type == SearchType.TITLE:
             return self._title_search(request, params.validated_data, *args, **kwargs)
 
-        try:
-            request.session = refresh_access_token(request.session)
-        except AuthenticationFailed:
-            logging.warning(
-                "User unauthenticated or error while refreshing token, "
-                "falling back to title search."
-            )
-            return self._title_search(request, params.validated_data, *args, **kwargs)
-
         indexer = get_document_indexer()
         if indexer is None:
             # fallback on title search if the indexer is not configured
@@ -1447,7 +1438,7 @@ class DocumentViewSet(
             return self._search_with_indexer(
                 indexer, request, params=params, search_type=search_type
             )
-        except requests.exceptions.RequestException as e:
+        except (requests.exceptions.RequestException, AuthenticationFailed) as e:
             logger.error(
                 "Error while searching documents with indexer \n%s \nfall back on title search",
                 e,
@@ -1473,6 +1464,7 @@ class DocumentViewSet(
         """
         Returns a list of documents matching the query (q) according to the configured indexer.
         """
+        request.session = refresh_access_token(request.session)
         queryset = models.Document.objects.all()
 
         results = indexer.search(
--- a/src/backend/core/tests/conftest.py
+++ b/src/backend/core/tests/conftest.py
@@ -147,8 +147,8 @@ def user_token():
     return build_authorization_bearer("some_token")
 
 
-@pytest.fixture(name="oidc_settings")
-def fixture_oidc_settings(settings):
+@pytest.fixture
+def oidc_settings(settings):
     """Fixture to configure OIDC settings for the tests."""
     settings.OIDC_OP_TOKEN_ENDPOINT = "https://auth.example.com/token"
     settings.OIDC_OP_AUTHORIZATION_ENDPOINT = "https://auth.example.com/authorize"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ and this project adheres to
 - ♿️(frontend) fix waffle aria-label spacing for new-window links #2030
 - 🐛(backend) stop using add_sibling method to create sandbox document #2084
 - 🐛(backend) duplicate a document as last-sibling #2084
+- 🐛(backend) fix search without refresh token #2090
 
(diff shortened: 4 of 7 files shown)


src/backend/core/api/utils.py
```diff
@@ -179,3 +182,31 @@ def get_ident(self, request):
             if x_forwarded_for
             else request.META.get("REMOTE_ADDR")
         )
+
+
+def refresh_access_token(session):
```
lunika, line 187: This method is not tested, you should have dedicated unit tests.

src/backend/core/api/viewsets.py
```diff
@@ -1430,6 +1429,15 @@ def search(self, request, *args, **kwargs):
...
             return self._title_search(request, params.validated_data, *args, **kwargs)
 
+        try:
+            request.session = refresh_access_token(request.session)
+        except AuthenticationFailed:
+            logging.warning(
+                "User unauthenticated or error while refreshing token, "
+                "falling back to title search."
+            )
+            return self._title_search(request, params.validated_data, *args, **kwargs)
```
lunika: Not sure this is the behavior we want.

You should probably put this code in the `_search_with_indexer` method and catch this exception in the try/except block calling the `_search_with_indexer` method.

src/backend/core/tests/conftest.py
```diff
@@ -143,3 +145,28 @@ def user_token():
     A fixture to create a user token for testing.
     """
     return build_authorization_bearer("some_token")
+
+
+@pytest.fixture(name="oidc_settings")
+def fixture_oidc_settings(settings):
```
lunika:
```suggestion
@pytest.fixture
def oidc_settings(settings):
```

lunika:
Can you also please update the readme ?
Thanks

src/backend/core/api/viewsets.py
mascarpon3: honestly i think writing this in this function makes the code more readable. the main search function is in charged of all the dispatch to the sub-functions. 

what i realise i should have done though is writing it bellow the `if indexer is None` sheck

lunika:
```
@drf.decorators.action(detail=False, methods=["get"], url_path="search")
    def search(self, request, *args, **kwargs):
        """
        Returns an ordered list of documents best matching the search query parameter 'q'.

        It depends on a search configurable Search Indexer. If no Search Indexer is configured
        or if it is not reachable, the function falls back to a basic title search.
        """
        params = serializers.SearchDocumentSerializer(data=request.query_params)
        params.is_valid(raise_exception=True)
        search_type = self._get_search_type()

        if search_type == SearchType.TITLE:
            return self._title_search(request, params.validated_data, *args, **kwargs)

        indexer = get_document_indexer()
        if indexer is None:
            # fallback on title search if the indexer is not configured
            return self._title_search(request, params.validated_data, *args, **kwargs)

        try:
            return self._search_with_indexer(
                indexer, request, params=params, search_type=search_type
            )
        except (requests.exceptions.RequestException, AuthenticationFailed) as e:
            logger.error(
                "Error while searching documents with indexer \n%s \nfall back on title search",
                e,
            )
            # fallback on title search if the indexer is not reached
            return self._title_search(request, params.validated_data, *args, **kwargs)

    def _get_search_type(self) -> SearchType:
        """
        Returns the search type to use for the search endpoint based on feature flags.
        If a user has both flags activated the most advanced search is used
        (HYBRID > FULL_TEXT > TITLE).
        A user with no flag will default to the basic title search.
        """
        if waffle.flag_is_active(self.request, FeatureFlag.FLAG_FIND_HYBRID_SEARCH):
            return SearchType.HYBRID
        if waffle.flag_is_active(self.request, FeatureFlag.FLAG_FIND_FULL_TEXT_SEARCH):
            return SearchType.FULL_TEXT
        return SearchType.TITLE

    @staticmethod
    def _search_with_indexer(indexer, request, params, search_type):
        """
        Returns a list of documents matching the query (q) according to the configured indexer.
        """
        request.session = refresh_access_token(request.session)
        queryset = models.Document.objects.all()

        results = indexer.search(
            q=params.validated_data["q"],
            search_type=search_type,
            token=request.session.get("oidc_access_token"),
            path=(
                params.validated_data["path"]
                if "path" in params.validated_data
                else None
            ),
            visited=get_visited_document_ids_of(queryset, request.user),
        )

        return drf_response.Response(
            {
                "count": len(results),
                "next": None,
                "previous": None,
                "results": results,
            }
        )
```
Well let me know if this code is less or more readable ? 

Moreover, as the `refresh_access_token` is related to the `refresh_access_token`, I think it should be called in this method.

src/backend/core/api/utils.py
```diff
@@ -179,3 +182,31 @@ def get_ident(self, request):
...
+
+    response = requests.post(
+        settings.OIDC_OP_TOKEN_ENDPOINT,
+        data={
+            "grant_type": "refresh_token",
+            "client_id": settings.OIDC_RP_CLIENT_ID,
+            "client_secret": settings.OIDC_RP_CLIENT_SECRET,
+            "refresh_token": refresh_token,
+        },
+        timeout=5,
```
lunika:
```suggestion
        timeout=settings.OIDC_TIMEOUT,
```

src/backend/core/api/utils.py
lunika: Also, in this function you are not testing if the access token is expired or not. I don't see the interest in refreshing a still valid access token.

Also you are not refreshing the `oidc_id_token_expiration` parameter in the session

lunika:
Closed in favor of #2097

PR #2093, opened 2026-03-20
♿️(frontend) add aria-hidden to decorative icons in dropdown menu

Ovgodd:
## Description

Add `aria-hidden` to decorative SVG icons in the dropdown menu so they are ignored by assistive technologies. ( NVDA / VO )

## Proposal

- [x] Add `aria-hidden` and `focusable="false"` to custom SVG icons rendered in `DropdownMenu` options
- [x] Extract icon rendering logic into a `customIconElement` constant for clarity

### DIFF
--- a/src/frontend/apps/impress/src/components/dropdown-menu/DropdownMenu.tsx
+++ b/src/frontend/apps/impress/src/components/dropdown-menu/DropdownMenu.tsx
@@ -275,7 +275,6 @@ export const DropdownMenu = ({
                     <Box
                       $theme="neutral"
                       $variation={isDisabled ? 'tertiary' : 'primary'}
-                      aria-hidden="true"
                     >
                       {option.icon}
                     </Box>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to
 - 💄(frontend) improve comments highlights #1961
 - ♿️(frontend) improve BoxButton a11y and native button semantics #2103
 - ♿️(frontend) improve language picker accessibility #2069
+- ♿️(frontend) add aria-hidden to decorative icons in dropdown menu #2093
 
 ### Fixed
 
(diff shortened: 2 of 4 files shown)


src/frontend/apps/impress/src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -267,12 +282,13 @@ export const DropdownMenu = ({
                     />
                   )}
 
-                  {option.icon && typeof option.icon !== 'string' && (
+                  {customIconElement && (
                     <Box
                       $theme="neutral"
                       $variation={isDisabled ? 'tertiary' : 'primary'}
+                      aria-hidden="true"
```
AntoLC: You add `aria-hidden="true"` here, is it necessary to add it on the icon as well ?

Ovgodd: > You add `aria-hidden="true"` here, is it necessary to add it on the icon as well ?

Yes you are right, the aria-hidden on the Box alone would work, but putting it directly on each icon is cleaner (and respect RGAA rules ),
it's explicit about what we're hiding and why. 
If someone refactors the layout later, the icons stay hidden. 
But I I've removed it from the Box to avoid redundancy.

AntoLC: Yes but in every rerender we are recloning a element, to add a aria that is not totally necessary, so I am mitigate about this code snippet. Is it not cleaner so to add the `aria-hidden` directly on the icon when it is used ?

src/frontend/apps/impress/src/components/dropdown-menu/DropdownMenu.tsx
```diff
@@ -182,6 +185,18 @@ export const DropdownMenu = ({
...
+              ? isValidElement(option.icon)
+                ? cloneElement(
+                    option.icon as ReactElement<{
+                      'aria-hidden'?: boolean;
+                    }>,
+                    { 'aria-hidden': true },
+                  )
+                : option.icon
+              : null;
+
```
AntoLC: Not sure about this snippet, comment here: https://github.com/suitenumerique/docs/pull/2093#discussion_r2986488284

I think we should either, let the Box forcing the aria-hidden (parent level), or adding it to the icon directly.

Ovgodd: Hey ! 

Yes, adding it to the icon directly is a good compromise, I agree! I've updated all SVG icons in DocsGridActions and DocToolBox with aria-hidden="true", and removed the cloneElement logic from DropdownMenu.

PR #2097, opened 2026-03-20
🐛(backend) allow using search endpoint without refresh token enabled

### LINKED ISSUE #2087: Search endpoint does not work without OIDC refresh token

lunika:
## Purpose

The search endpoint was using the refresh_roken method decorator. This decorator force having a valid refresh token stored in the session for the entire viewset. The search endpoint still allow having the legacy search behavior and for this we don't need to configure at all the OIDC refrsh mechanism.

## Proposal

- [x]  🐛(backend) allow using search endpoint without refresh token enabled

Fix #2087

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,10 @@ and this project adheres to
 
 - ♿️(frontend) fix aria-labels for table of contents #2065
 
+### Fixed
+
+- 🐛(backend) allow using search endpoint without refresh token enabled #2097
+
 ## [v4.8.2] - 2026-03-19
 
 ### Added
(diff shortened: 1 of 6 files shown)


mascarpon3:
how is it solving anything ? when you turn oidc back on to activate Find you will run into the same issue.

lunika:
This PR is not made to fix the refresh token lifetime but  to allow the use of the legacy search when you didn't activate the refresh token storage.

So yes, it resolves the main issue we have: using the normal/legacy search without a refresh token stored in the session.

It does the same thing you made in your PR, differently, still relying on the django-lasuite middleware when needed. And it was what I mentioned in my last review, and you didn't want to consider it.

mascarpon3:
as you say, this allows users without refresh token to use the title search but you can not use Find.

this does not the same as [that](https://github.com/suitenumerique/docs/pull/2090). my fix allows to use Find.

PR #2103, opened 2026-03-23
♿️(frontend) improve BoxButton a11y and native button semantics

Ovgodd:
## Purpose

Improve `BoxButton` semantics and accessibility: avoid accidental form submission, expose a disabled state to assistive technologies, and align TypeScript refs with the actual DOM (`<button>`).

## Proposal

- [x] Set `type="button"` on `BoxButton` by default
- [x] Expose `aria-disabled` when the `disabled` prop is set (keep keyboard focus for menu patterns)
- [x] Type `forwardRef` and related refs as `HTMLButtonElement` (`DocIcon`, doc tree, `DropdownMenu`, `useDropdownKeyboardNav`)

### DIFF
--- a/src/frontend/apps/impress/src/features/docs/doc-management/components/DocIcon.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-management/components/DocIcon.tsx
@@ -44,7 +44,7 @@ export const DocIcon = ({
   const { t } = useTranslation();
   const { addLastFocus, restoreFocus } = useFocusStore();
 
-  const iconRef = useRef<HTMLDivElement>(null);
+  const iconRef = useRef<HTMLButtonElement>(null);
 
   const [openEmojiPicker, setOpenEmojiPicker] = useState<boolean>(false);
   const [pickerPosition, setPickerPosition] = useState<{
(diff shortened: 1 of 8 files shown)


src/frontend/apps/impress/src/features/docs/doc-management/components/DocIcon.tsx
```diff
@@ -44,7 +44,10 @@ export const DocIcon = ({
   const { t } = useTranslation();
   const { addLastFocus, restoreFocus } = useFocusStore();
 
-  const iconRef = useRef<HTMLDivElement>(null);
+  // Omit ref from buttonProps: BoxType refs target HTMLDivElement but BoxButton is a button.
+  const { ref: _buttonPropsRef, ...restButtonProps } = buttonProps ?? {};
```
AntoLC: The root problem is because `BoxButtonType` has a bad ref type, could we fix BoxButtonType instead ?
https://github.com/suitenumerique/docs/blob/4db585ef3c9416b6b0fe32f7620957f96b599390/src/frontend/apps/impress/src/components/BoxButton.tsx#L6-L8

Something like that:
``` ts
export type BoxButtonType = Omit<BoxType, 'ref'> & {
  disabled?: boolean;
  ref?: Ref<HTMLButtonElement>;
};
```

Ovgodd: indeed it is much cleaner

PR #2117, opened 2026-03-24
♿️(frontend) replace ARIA grid pattern with list in docs grid

Ovgodd:
## Purpose

The document grid used `role="grid"` with `role="row"`, `role="gridcell"`, `role="rowgroup"`, and `role="columnheader"`. This ARIA grid pattern requires full keyboard navigation with arrow keys which it's not implemented and demand to much complexity, . With role="grid", some screen readers switch to a mode where arrow keys are handled by the page; since the page did not implement grid navigation, that interaction did not work as expected.

## Proposal

- [x] Replace `role="grid"` / `role="row"` / `role="gridcell"` / `role="columnheader"` / `role="rowgroup"` with `role="list"` / `role="listitem"`, matching the actual semantics (a list of documents, not a data grid)
- [x] Mark visual column headers (`Name`, `Updated at`) as `aria-hidden="true"` since they are no longer part of the list semantics
- [x] Add translatable `screenReaderInstructions` and `announcements` on `DndContext` (`accessibility` prop) so SR copy can be localized via i18n

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,9 +13,12 @@ and this project adheres to
 ### Changed
 
 - 💄(frontend) improve comments highlights #1961
-- ♿️(frontend) improve BoxButton a11y and native button semantics
-  #2103
+- ♿️(frontend) improve BoxButton a11y and native button semantics #2103
 - ♿️(frontend) improve language picker accessibility #2069
+
+### Fixed
+
+- 🐛(y-provider) destroy Y.Doc instances after each convert request #2129
 
 ## [v4.8.3] - 2026-03-23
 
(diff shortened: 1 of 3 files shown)


src/frontend/apps/impress/src/features/docs/docs-grid/components/DocsGridItem.tsx
```diff
@@ -33,20 +32,13 @@ export const DocsGridItem = ({ doc, dragMode = false }: DocsGridItemProps) => {
   const { flexLeft, flexRight } = useResponsiveDocGrid();
   const { spacingsTokens } = useCunninghamTheme();
 
-  const handleKeyDown = (e: KeyboardEvent) => {
-    if (e.key === 'Enter' || e.key === ' ') {
-      e.preventDefault();
-      (e.target as HTMLAnchorElement).click();
-    }
-  };
```
AntoLC, line 41: I don't have the feeling that we can navigate to a document with the keyboard anymore.

Ovgodd:
> I think we should add a accessibility test on the doc grid, because this comment ([#2117 (comment)](https://github.com/suitenumerique/docs/pull/2117#discussion_r2980099014)) highlight that we can have strong regressions passing our tests.

this is a very good idea

Issue #1254, opened 2025-08-01
🐛(backend) Delete document request conflict


AntoLC:
## Bug Report

A owner can still access a deleted document, by accessing this deleted document, the request `tree` will be triggered, and work (200 response) but the response will not be has expected because the actual deleted doc will not be part of the children of this tree anymore, creating a loop frontend side.

## Possible solutions

- Like the other, the owner get a 404 ?
https://github.com/suitenumerique/docs/blob/32e3c70e33663f85c20735716a6fadcf94e2193b/src/backend/core/api/permissions.py#L141

- Or in the tree request, if it is the owner, we don't add a filter on the deleted documents:
https://github.com/suitenumerique/docs/blob/32e3c70e33663f85c20735716a6fadcf94e2193b/src/backend/core/api/viewsets.py#L881

lunika:
IMO this modification should be made once the trash feature is made.

AntoLC:
I do not succeed to reproduce anymore this issue for the moment.

virgile-dev:
Can't reproduce, closing

PR #2121, opened 2026-03-24
🐛(backend) do not filter deleted docs for the owner in the tree view

Ash-Crow:
## Purpose
Fixes the bug described in #1254 (Delete document request conflict)

In the tree request, deleted documents for
which the user is the owner are not filtered anymore.

## Proposal
- [x] If the user is logged in, their deleted documents are now included in the query for children.
- [x] Update unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to
 
 - 💄(frontend) improve comments highlights #1961
 
+### Fixed
+
+- 🐛(backend) do not filter deleted documents for the owner in the tree #2121
+
 ## [v4.8.3] - 2026-03-23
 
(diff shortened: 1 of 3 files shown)


AntoLC (commented):
I don't succeed to reproduce the bug, did you get it ?

virgile-dev:
Hey @Ash-Crow 
Like @AntoLC said, apparently the bug was solved when we released the trash feature.
Sorry I gave this you. If you can try to reproduce before you make a PR that'd be much appreciated.
Finally, I discussed the behaviour you implemented with @lunika. It's not the desired one.
We do not wish to display deleted child for owners.
Let's discuss it thursday to see if we close it.

PR #2122, opened 2026-03-24
♿️(frontend) fix sidebar resize handle for screen readers

### LINKED ISSUE #2110: Resizable separator: inactive with keyboard
## Bug Report

**Problematic behavior**
The resizable separator between the sidebar and the main content is reachable via Tab. The Home and End keys work correctly, but the left and right arrow keys do not allow resizing the panels.

**Expected behavior/code**
The left and right arrow keys must allow progressive resizing of the sidebar, just as Home and End allow reaching the extreme values.

**Steps to Reproduce**
1. Open a document in Docs
2. Navigate with the keyboard using Tab to the resizable separator
3. Press the left and right arrow keys
4. The panels do not resize — unlike Home and End which work

**Possible Solution**
```
<PanelResizeHandle
  onKeyDown={(e) => {
    // Home et End déjà implémentés
    // Ajouter la gestion des flèches
    if (e.key === 'ArrowLeft') resizePanel(-1);
    if (e.key === 'ArrowRight') resizePanel(1);
  }}
/>
```

**Critères RGAA**

**Critère 7.1 :** [Chaque script est-il, si nécessaire, compatible avec les technologies d'assistance ?](https://accessibilite.numerique.gouv.fr/methode/criteres-et-tests/#7.1)

**Impact**
A user navigating exclusively with the keyboard can reach the separator and toggle between minimum and maximum sizes via Home and End, but cannot perform progressive resizing with the arrow keys.

Ovgodd:
## Purpose

Fix left sidebar keyboard resize with a screen reader on, and improve SR output plus step size.

## Proposal

- [x]  Replace the default separator role with slider so that, when the handle is focused, screen readers pass Left/Right to the app instead of using them for reading (known issue with separator in NVDA).
- [x] from `aria-valuenow` into `aria-valuetext` (width in px).
- [x] Use `keyboardResizeBy={1}` for smaller steps.

### DIFF
--- a/src/frontend/apps/impress/src/features/left-panel/components/ResizableLeftPanel.tsx
+++ b/src/frontend/apps/impress/src/features/left-panel/components/ResizableLeftPanel.tsx
@@ -1,4 +1,5 @@
 import { useEffect, useRef, useState } from 'react';
+import { useTranslation } from 'react-i18next';
 import {
   ImperativePanelHandle,
   Panel,
@@ -15,6 +16,27 @@ const pxToPercent = (px: number) => {
   return (px / window.innerWidth) * 100;
 };
 
+const RESIZE_HANDLE_ID = 'left-panel-resize-handle';
+
+const getValueLabel = (
+  current: number,
+  min: number,
+  max: number,
+  t: (key: string) => string,
+): string => {
+  if (max <= min) {
+    return t('Sidebar width: medium');
+  }
+  const ratio = (current - min) / (max - min);
+  if (ratio < 1 / 3) {
+    return t('Sidebar width: narrow');
+  }
+  if (ratio < 2 / 3) {
+    return t('Sidebar width: medium');
+  }
+  return t('Sidebar width: wide');
+};
+
 type ResizableLeftPanelProps = {
   leftPanel: React.ReactNode;
   children: React.ReactNode;
@@ -28,6 +50,7 @@ export const ResizableLeftPanel = ({
   minPanelSizePx = 300,
   maxPanelSizePx = 450,
 }: ResizableLeftPanelProps) => {
+  const { t } = useTranslation();
   const { isDesktop } = useResponsiveStore();
   const { isPanelOpen } = useLeftPanelStore();
   const ref = useRef<ImperativePanelHandle>(null);
@@ -96,6 +119,24 @@ export const ResizableLeftPanel = ({
     };
   }, [isDesktop]);
 
+  /**
+   * Workaround: NVDA does not enter focus mode for role="separator"
+   * (https://github.com/nvaccess/nvda/issues/11403), so arrow keys are
+   * intercepted by browse-mode navigation and never reach the handle.
+   * Changing the role to "slider" makes NVDA reliably switch to focus
+   * mode, restoring progressive keyboard resize with arrow keys.
+   *
+   * Note: PanelResizeHandle does not expose a ref (no RefAttributes in its
+   * type definition), so we use id + getElementById as the only viable option.
+   * Only role needs to be overridden here; aria-* props are passed directly.
+   */
+  useEffect(() => {
+    if (!isPanelOpen) {
+      return;
+    }
+    document.getElementById(RESIZE_HANDLE_ID)?.setAttribute('role', 'slider');
+  }, [isPanelOpen]);
+
   const handleResize = (sizePercent: number) => {
     const widthPx = (sizePercent / 100) * window.innerWidth;
     savedWidthPxRef.current = widthPx;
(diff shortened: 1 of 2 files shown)


AntoLC:
@CodeRabbit review

src/frontend/apps/impress/src/features/left-panel/components/ResizableLeftPanel.tsx
```diff
@@ -132,6 +180,7 @@ export const ResizableLeftPanel = ({
       </Panel>
       {isPanelOpen && (
         <PanelResizeHandle
+          id={RESIZE_HANDLE_ID}
```
AntoLC, line 183: Is the `ref` not working ?

src/frontend/apps/impress/src/features/left-panel/components/ResizableLeftPanel.tsx
```diff
@@ -96,14 +100,58 @@ export const ResizableLeftPanel = ({
...
+
+    const updateValueText = () => {
+      const value = handle.getAttribute('aria-valuenow');
+      if (value) {
+        const widthPx = Math.round(
+          (parseFloat(value) / 100) * window.innerWidth,
+        );
+        handle.setAttribute(
+          'aria-valuetext',
+          t('Sidebar width: {{widthPx}} pixels', { widthPx }),
```
AntoLC, line 130: Not big fan of the `useEffect`. 
I can see that the role `slider` cannot be set directly to the component, fair enough to set it in the useEffect, but the other props can be set directly to the component.

We can do somethink like:
```tsx
<PanelResizeHandle
          id={RESIZE_HANDLE_ID}
          ...
          aria-orientation="vertical"
          aria-label={t('Resize sidebar')}
          aria-valuetext={t('Sidebar width: {{widthPx}} pixels', {
            widthPx: Math.round((panelSizePercent / 100) * window.innerWidth),
          })}
        />
```

Seems more clean than using the `seEffect` to mutate the component, wdyt ?

src/frontend/apps/impress/src/features/left-panel/components/ResizableLeftPanel.tsx
```diff
@@ -96,14 +100,58 @@ export const ResizableLeftPanel = ({
...
+
   const handleResize = (sizePercent: number) => {
     const widthPx = (sizePercent / 100) * window.innerWidth;
     savedWidthPxRef.current = widthPx;
     setPanelSizePercent(sizePercent);
   };
 
   return (
-    <PanelGroup direction="horizontal">
+    <PanelGroup direction="horizontal" keyboardResizeBy={1}>
```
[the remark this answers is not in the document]
Ovgodd: it is supported in our current version

`    autoSaveId?: string | null;
    className?: string;
    direction: Direction;
    id?: string | null;
    keyboardResizeBy?: number | null;
    onLayout?: PanelGroupOnLayout | null;
    storage?: PanelGroupStorage;
    style?: CSSProperties;
    tagName?: keyof HTMLElementTagNameMap;
    dir?: "auto" | "ltr" | "rtl" | undefined;
}>;`

src/frontend/apps/impress/src/features/left-panel/components/ResizableLeftPanel.tsx
Ovgodd: indeed PanelResizeHandle does not declare `RefAttributes<HTMLDivElement>` in its types (v3.0.6).

src/frontend/apps/impress/src/features/left-panel/components/ResizableLeftPanel.tsx
Ovgodd: Yes, good idea! I moved aria-orientation, aria-label and aria-valuetext directly to the component. I also changed the SR announcement instead of a raw percentage, it now says narrow, medium or wide based on the position within the allowed range, which feels more natural for screen reader users.

src/frontend/apps/impress/src/features/left-panel/components/ResizableLeftPanel.tsx
```diff
@@ -132,6 +169,26 @@ export const ResizableLeftPanel = ({
...
+              maxPanelSizePercent,
+            );
+            if (label === 'narrow') {
+              return t('Sidebar width: narrow');
+            }
+            if (label === 'medium') {
+              return t('Sidebar width: medium');
+            }
+            return t('Sidebar width: wide');
+          })()}
```
AntoLC: We can maybe simplify a little bit, then directly in the function you can call `t('Sidebar width: narrow')`

```suggestion
          aria-valuetext={getValueLabel(
            panelSizePercent,
            minPanelSizePercent,
            maxPanelSizePercent,
            t
          )}
```

AntoLC (approved):
Like pointed the rabbit, you should probably managed this case `max === min`.

src/frontend/apps/impress/src/features/left-panel/components/ResizableLeftPanel.tsx
Ovgodd: good point !

PR #2124, opened 2026-03-24
🐛(backend) create_for_owner: add accesses before saving doc content

### LINKED ISSUE #2123: Create accesses before saving content for meet transcription
## Bug Report

**Problematic behavior**
The current way it is done
1. Create doc
2. Save content 
3. Add accesses
This creates problems as the user doesn't have access to the first version of the document.

**Expected behavior/code**
The way we want to do this
1. Create doc
2. Add accesses 
3. Save content 

**Steps to Reproduce**
1. Do this...
2. Then this...
3. And then the bug happens!

**Environment**
- Docs version: 4.8.3

**Possible Solution**

Ash-Crow:
## Purpose
We add the User Accesses before saving content so the user is sure to have access to the the first version when creating a doc through create_for_owner (fixes #2123)

## Proposal
- [x] Changed the order of operations in `ServerCreateDocumentSerializer`
- [x] Added a unit test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,10 @@ ctrl/command/middle-mouse click #2170
 - 🐛(y-provider) destroy Y.Doc instances after each convert request #2129
 - 🐛(backend) remove deleted sub documents in favorite_list endpoint #2083
 
+### Fixed
+
+- 🐛(backend) create_for_owner: add accesses before saving doc content #2124
+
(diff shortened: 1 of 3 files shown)


lunika:
The tests are failing on the tests you added

Ash-Crow:
> The tests are failing on the tests you added

They should be fixed now.

PR #2133, opened 2026-03-25
Improve indexing command

mascarpon3:
## Purpose

to index all documents we need to be able to restart from last successful batch in case of crash. to do so we need to be able to specify time bounds on the bulk index to allow recovering from checkpoint after crash and asynchronously delegate it to celery worker. 
 
## Proposal

- [x] ✨ cache the last successful document.updated_at
- [x] ✨ add `lower_time_bound` and `upper_time_bound` args
- [x] ✨ apply async `batch_document_indexer_task`
- [x] ✨ make the command available in django admin

## External contributions

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -260,7 +260,7 @@ demo: ## flush db then create a demo for load testing purpose
 .PHONY: demo
 
 index: ## index all documents to remote search
-	@$(MANAGE) index
+	@$(MANAGE) index $(args)
 .PHONY: index
 
 # Nota bene: Black should come after isort just in case they don't agree...
--- a/src/backend/core/admin.py
+++ b/src/backend/core/admin.py
@@ -1,14 +1,53 @@
 """Admin classes and registrations for core app."""
 
 from django.contrib import admin, messages
+from django.contrib.admin.views.decorators import staff_member_required
 from django.contrib.auth import admin as auth_admin
-from django.shortcuts import redirect
+from django.core.management import call_command
+from django.http import HttpRequest
+from django.shortcuts import redirect, render
 from django.utils.translation import gettext_lazy as _
 
 from treebeard.admin import TreeAdmin
 
 from core import models
+from core.forms import RunIndexingForm
 from core.tasks.user_reconciliation import user_reconciliation_csv_import_job
+
+# Customize the default admin site's get_app_list method
+_original_get_app_list = admin.site.get_app_list
+
+
+def custom_get_app_list(_self, request, app_label=None):
+    """Add custom commands to the app list."""
+    app_list = _original_get_app_list(request, app_label)
+
+    # Add Commands app with Run Indexing command
+    commands_app = {
+        "name": _("Commands"),
+        "app_label": "commands",
+        "app_url": "#",
+        "has_module_perms": True,
+        "models": [
+            {
+                "name": _("Run indexing"),
+                "object_name": "RunIndexing",
+                "admin_url": "/admin/run-indexing/",
+                "view_only": False,
+                "add_url": None,
+                "change_url": None,
+            }
+        ],
+    }
+
+    app_list.append(commands_app)
+    return app_list
+
+
+# Monkey-patch the admin site
+admin.site.get_app_list = lambda request, app_label=None: custom_get_app_list(
+    admin.site, request, app_label
+)
 
 
 @admin.register(models.User)
@@ -227,3 +266,39 @@ class InvitationAdmin(admin.ModelAdmin):
     def save_model(self, request, obj, form, change):
         obj.issuer = request.user
         obj.save()
+
+
+@staff_member_required
+def run_indexing_view(request: HttpRequest):
+    """Custom admin view for running indexing commands."""
+    if request.method == "POST":
+        form = RunIndexingForm(request.POST)
+        if form.is_valid():
+            lower_time_bound = form.cleaned_data.get("lower_time_bound")
+            upper_time_bound = form.cleaned_data.get("upper_time_bound")
+            call_command(
+                "index",
+                batch_size=form.cleaned_data["batch_size"],
+                lower_time_bound=lower_time_bound.isoformat()
+                if lower_time_bound
+                else None,
+                upper_time_bound=upper_time_bound.isoformat()
+                if upper_time_bound
+                else None,
+                async_mode=True,
+            )
+            messages.success(request, _("Indexing triggered!"))
+            return redirect("run_indexing")
+        messages.error(request, _("Please correct the errors below."))
+    else:
+        form = RunIndexingForm()
+
+    return render(
+        request=request,
+        template_name="runindexing.html",
+        context={
+            **admin.site.each_context(request),
+            "title": "Run Indexing Command",
+            "form": form,
+        },
+    )
--- a/src/backend/core/management/commands/index.py
+++ b/src/backend/core/management/commands/index.py
@@ -40,7 +40,7 @@ class Command(BaseCommand):
         batch_size = options["batch_size"]
 
         try:
-            count = indexer.index(batch_size=batch_size)
+            count = indexer.index(batch_size=batch_size, crash_safe_mode=True)
         except Exception as err:
             raise CommandError("Unable to regenerate index") from err
 
--- a/src/backend/core/management/commands/index.py
+++ b/src/backend/core/management/commands/index.py
@@ -4,12 +4,16 @@ Handle search setup that needs to be done at bootstrap time.
 
 import logging
 import time
+from datetime import datetime
 
+from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
 
+from core import models
 from core.services.search_indexers import get_document_indexer
+from core.tasks.search import batch_document_indexer_task
 
-logger = logging.getLogger("docs.search.bootstrap_search")
+logger = logging.getLogger(__name__)
 
 
 class Command(BaseCommand):
@@ -24,8 +28,31 @@ class Command(BaseCommand):
             action="store",
             dest="batch_size",
             type=int,
-            default=50,
+            default=settings.SEARCH_INDEXER_BATCH_SIZE,
             help="Indexation query batch size",
+        )
+        parser.add_argument(
+            "--lower-time-bound",
+            action="store",
+            dest="lower_time_bound",
+            type=datetime.fromisoformat,
+            default=None,
+            help="DateTime in ISO format. Only documents updated after this date will be indexed",
+        )
+        parser.add_argument(
+            "--upper-time-bound",
+            action="store",
+            dest="upper_time_bound",
+            type=datetime.fromisoformat,
+            default=None,
+            help="DateTime in ISO format. Only documents updated before this date will be indexed",
+        )
+        parser.add_argument(
+            "--async",
+            action="store_true",
+            dest="async_mode",
+            default=False,
+            help="Whether to execute indexing asynchronously in a Celery task (default: False)",
         )
 
     def handle(self, *args, **options):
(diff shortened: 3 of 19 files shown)


src/backend/core/management/commands/index.py
```diff
@@ -27,6 +28,32 @@ def add_arguments(self, parser):
...
+        parser.add_argument(
+            "--crash-safe-mode",
+            action="store_true",
+            dest="crash_safe_mode",
+            default=False,
+            help="If True, order documents by updated_at and store "
+            "the last indexed document.updated_at in cache. "
+            "This allows resuming indexing from the last successful batch in case of a crash "
+            "but is more computationally expensive due to sorting.",
+        )
```
qbey, line 56: This shouldn't be an option : we always have to be in "crash safe mode".

src/backend/core/management/commands/index.py
```diff
@@ -35,18 +62,15 @@ def handle(self, *args, **options):
...
 
-        logger.info("Starting to regenerate Find index...")
-        start = time.perf_counter()
-        batch_size = options["batch_size"]
-
-        try:
-            count = indexer.index(batch_size=batch_size)
-        except Exception as err:
-            raise CommandError("Unable to regenerate index") from err
+        batch_document_indexer_task.apply_async(
```
qbey: I don't understand the role of this command now you can't kill it. There is no purpose of having this command to send something to celery. Or at least "--async" should be an option, because I expect a management command to be killable.

src/backend/core/templates/runindexing.html
```diff
@@ -0,0 +1,43 @@
...
+
+    <div>
+        <label for="crash_safe_mode" style="margin-right: 10px">Crash save mode</label>
+        <input id="crash_safe_mode" type="checkbox" name="crash_safe_mode" value="true" checked>
+    </div>
+
+    <input type="submit" value="{% translate 'Run Indexing' %}" style="margin-top: 20px;">
+</form>
+
+{% endblock %}
```
qbey, line 22: You may have a look at Django forms which helps building forms.

src/backend/core/tests/conftest.py
```diff
@@ -143,3 +143,10 @@ def user_token():
     A fixture to create a user token for testing.
     """
     return build_authorization_bearer("some_token")
+
+
+def create_document_with_updated_at(updated_at, **kwargs):
+    """Helper to create a document with a specific updated_at, bypassing auto_now=True."""
+    document = factories.DocumentFactory(**kwargs)
+    models.Document.objects.filter(pk=document.pk).update(updated_at=updated_at)
+    return models.Document.objects.get(pk=document.pk)
```
qbey:
```suggestion
```

Don't do that, prefer to use `factories.DocumentFactory(updated_at=XXX)` directly in tests.

src/backend/core/tasks/search.py
```diff
@@ -1,95 +0,0 @@
-"""Trigger document indexation using celery task."""
```
qbey, line 1: Making such a diff in a PR where you make other changes is not appropriate, this makes the review really hard.

src/backend/core/models.py
```diff
@@ -2028,3 +2029,17 @@ def get_abilities(self, user):
...
+
+    class Meta:
+        """Meta options."""
+
+        managed = False
+        verbose_name = _("Run Indexing")
+
+    def __str__(self):
+        """String representation."""
+        return self.Meta.verbose_name
```
qbey: You should read the documentation about "models `__str__`"

src/backend/core/models.py
```diff
@@ -2028,3 +2029,17 @@ def get_abilities(self, user):
             "partial_update": is_admin_or_owner,
             "retrieve": is_admin_or_owner,
         }
+
+
+class RunIndexing(models.Model):
+    """Proxy model for indexing management in admin."""
```
qbey: "Proxy" models are a thing in Django, this is not a proxy model: https://docs.djangoproject.com/en/5.2/topics/db/models/#proxy-models

src/backend/core/models.py
qbey:
```suggestion
```

This model is useless :/

src/backend/core/utils.py
```diff
@@ -168,3 +168,47 @@ def users_sharing_documents_with(user):
         elapsed,
     )
     return result
+
+
+def build_indexable_documents_queryset(lower_time_bound=None, upper_time_bound=None):
```
qbey: This should not be in "utils".

This should be a QuerySet of the document's model: https://docs.djangoproject.com/en/5.2/topics/db/managers/#managers

src/backend/core/utils.py
```diff
@@ -168,3 +168,47 @@ def users_sharing_documents_with(user):
...
+
+    conditions = Q()
+    if lower_time_bound and upper_time_bound:
+        conditions = (
+            Q(updated_at__gte=lower_time_bound, updated_at__lte=upper_time_bound)
+            | Q(deleted_at__gte=lower_time_bound, deleted_at__lte=upper_time_bound)
+            | Q(
+                ancestors_deleted_at__gte=lower_time_bound,
+                ancestors_deleted_at__lte=upper_time_bound,
+            )
```
qbey: I don't understand why we deal with deletion here? Deletion might be considered eventually but elsewhere  and most likely not related to the filtering bounds.

Makefile
```diff
@@ -259,8 +259,16 @@ demo: ## flush db then create a demo for load testing purpose
...
+index: batch_size ?=
+index: lower_time_bound ?=
+index: upper_time_bound ?=
+index: crash_safe_mode ?=
+index: ## index all documents to indexer (params: batch_size, lower_time_bound, upper_time_bound, crash_safe_mode)
+	@$(MANAGE) index \
+		$(if $(batch_size),--batch-size $(batch_size)) \
+		$(if $(lower_time_bound),--lower-time-bound "$(lower_time_bound)") \
+		$(if $(upper_time_bound),--upper-time-bound "$(upper_time_bound)") \
+		$(if $(crash_safe_mode),--crash-safe-mode)
```
qbey: I know the code was already here, but I would suggest to simply add a `manage` make command, instead of rewriting every commands with every arguments.

src/backend/core/admin.py
```diff
@@ -227,3 +232,41 @@ class InvitationAdmin(admin.ModelAdmin):
     def save_model(self, request, obj, form, change):
         obj.issuer = request.user
         obj.save()
+
+
+@admin.register(models.RunIndexing)
+class RunIndexingAdmin(admin.ModelAdmin):
```
qbey: Not the way to do this.

Either add a new action on the Document model admin, or simply add a new page.

You may even need to do nothing if you add the way to run tasks manually in the admin (but I'm not sure Docs has celery beats, which might be a requirement).

=> I would simply add an action.

src/backend/core/services/search_indexers.py
```diff
@@ -134,35 +139,36 @@ def index(self, queryset=None, batch_size=None):
...
-                count += len(serialized_batch)
+            if not serialized_batch:
+                continue
+
+            self.push(serialized_batch)
+            count += len(serialized_batch)
+
+            if crash_safe_mode:
+                django_cache.set(
+                    BULK_INDEXER_CHECKPOINT, serialized_batch[-1]["updated_at"]
```
qbey: I'm not sure how to use this, because from the admin I have to put a `lower_time_bound`... There is a missing piece. Not sure the cache is helpful if there is another way to get this information. Also if we stay with the cache, an info log would still be appreciated.

src/backend/core/tests/conftest.py
mascarpon3: of course i thought about that. but on the `core.model.BaseModel` the `updated_at` has a `auto_now=True` which prevents overighting it with the factory

src/backend/core/tasks/search.py
mascarpon3: i comited separately so the review does not get too complicated. 

i droped it to simplify.

src/backend/core/utils.py
mascarpon3: i fallowed the logic implemented in `batch_document_indexer_task`.

why is it an issue ? we also want to index deleted documents

src/backend/core/admin.py
mascarpon3: i really had no idea how to add a command in the admin. 

i read the doc about model action but i understood they are things like:

Capture d’écran 2026-03-31 à 14 13 33

actions are not related to model but to instances which makes no sens in this context.

mascarpon3: also we could want to run any command from the admin, not related to a model in particular, and so i must be able to come up with a solution to integrate a new tab in the admin without refering to an existing model.

src/backend/core/models.py
mascarpon3: i know what this method is. i just fixed a lint error about not defining it

mascarpon3: i came up with a solution. the only issue i do not like is we lose the button to access it from the home page.

src/backend/core/services/search_indexers.py
mascarpon3: what other way to get the information ? what missing piece ?

also you do not HAVE TO put a `lower_time_bound`.

src/backend/core/tests/conftest.py
mascarpon3: the only solutions i can find is either this one or replacing

```
auto_now=True
````

with

```
default=timezone.now
````

in the `updated_at` of the `core.model.BaseModel`

i also thought about using `freeze_time` but the factory fires and exception in this context. 

````
botocore.exceptions.ClientError: An error occurred (403) when calling the HeadObject operation: Forbidden
````

File: src/backend/core/tests/conftest.py [hunk 13]
```diff
@@ -22,6 +24,20 @@ def clear_cache():
...
 
+@pytest.fixture
+def caplog_with_propagate(caplog):
+    """
+    propagate=False on settings.LOGGING loggers.
+    This prevents caplog from capturing logs.
+
+    This fixture enables propagation on all configured loggers.
+    """
+    for logger_name in django_settings.LOGGING.get("loggers", {}):
```
qbey, line 35: Why don't use the `settings` fixture here?

src/backend/core/templates/admin/index.html
```diff
@@ -0,0 +1,26 @@
...
+                    </th>
+                    <td></td>
+                    <td></td>
+                </tr>
+            </tbody>
+        </table>
+    </div>
+
+</div>
+{% endblock %}
```
qbey, line 26: It works, but only on the Django admin homepage left menu, when on another page, the menu is not there anymore. While this is not a big issue, I think it shows this is not the best way to do it? 
Did you consider overriding `admin.site.get_app_list` which could do the same, in a more integrated way?

src/backend/core/templates/runindexing.html
```diff
@@ -0,0 +1,25 @@
...
+    {% csrf_token %}
+
+    <hr style="margin-bottom: 10px;">
+
+    <p>
+        This command triggers the indexing of all documents within the specified time bound.
+    </p>
+    <p>
+        See <b>docs/commands/index.md</b> for more details.
+    </p>
```
qbey, line 16: Users usually don't read the code, also it's not explicit you are talking about the code repository. I think you can remove this, or give the full explanation here.

src/backend/core/tests/test_models_documents.py
```diff
@@ -1691,3 +1693,79 @@ def test_models_documents_compute_ancestors_links_paths_mapping_structure(
...
+    """Test time_filter with no filters returns all documents."""
+    factories.DocumentFactory.create_batch(3)
+
+    queryset = models.Document.objects.filter_updated_at()
+
+    assert queryset.count() == 3
+
+
+def test_models_documents_manager_time_filter_oldest_updated_at():
+    """Test filtering by oldest_updated_at includes documents updated after timestamp."""
```
qbey, line 1708:
```suggestion
    """Test filtering by oldest_updated_at includes documents updated after or at lower_time_bound."""
```

src/backend/core/tests/test_models_documents.py
```diff
@@ -1691,3 +1693,79 @@ def test_models_documents_compute_ancestors_links_paths_mapping_structure(
...
+    document_recent = DocumentFactory(updated_at=lower_time_bound + timedelta(days=15))
+
+    queryset = models.Document.objects.filter_updated_at(
+        lower_time_bound=lower_time_bound
+    )
+
+    assert queryset.count() == 2
+    assert document_too_old not in queryset
+    assert document_at_boundary in queryset
+    assert document_recent in queryset
```
qbey, line 1722: I would have written

```suggestion
    assert sorted(queryset.values_list("pk", flat=True)) == sorted([document_at_boundary.pk, document_recent.pk])
```

But your code works as well

src/backend/core/models.py
```diff
@@ -858,6 +859,31 @@ def annotate_user_roles(self, user):
             user_roles=models.Value([], output_field=output_field),
         )
 
+    def filter_updated_at(self, lower_time_bound=None, upper_time_bound=None):
```
qbey, line 862: You had a good point saying that if a parent document is deleted you need to catch the deleted children, the removal of "deleted ancestor" means that it's managed elsewhere?

src/backend/core/models.py
```diff
@@ -859,6 +859,38 @@ def annotate_user_roles(self, user):
...
+        elif lower_time_bound:
+            conditions = models.Q(updated_at__gte=lower_time_bound) | models.Q(
+                ancestors_deleted_at__gte=lower_time_bound
+            )
+        elif upper_time_bound:
+            conditions = models.Q(updated_at__lte=upper_time_bound) | models.Q(
+                ancestors_deleted_at__lte=upper_time_bound
+            )
+
+        return self.filter(conditions)
```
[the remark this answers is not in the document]
mascarpon3: i don't think i should modify `Document.restore` method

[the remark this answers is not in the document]
mascarpon3: if i understand well the issue:

if a file is restored its `deleted_at` field is removed and its `updated_at` and `ancestors_deleted_at` fields are not set to now. based only on `updated_at`, `ancestor_deleted_at` or `deleted_at` time bounds, restored documents may be missed by indexing and improperly remain in `is_active: false` mode.

is that the issue ?

either we add a `restore_at` field or restart the command from the start each time it crashes. we do not control what document will be restored.

PR #2134, opened 2026-03-25
🔧(y-provider) increase Node.js memory limit

AntoLC:
## Purpose

By default, Node.js has a memory limit of
around 512MB, which can lead to out-of-memory
errors when processing large documents.
This commit increases the memory limit to
2GB for the y-provider server, allowing
it to handle larger documents without crashing.


AntoLC:
Closed in favor of: https://github.com/suitenumerique/docs/pull/2132

PR #2151, opened 2026-03-30
🔧(backend) settings CONVERSION_UPLOAD_ENABLED to control usage of docspec

lunika:
## Purpose

We want to control the conversion of document at upload time. We want to
disable this feature using a settings. The new settings
CONVERSION_UPLOAD_ENABLED should be used to enable or not the conversion
at upload feature. If disabled and a file is uploaded, the reponse will
return a 400

## Proposal

- [x] 🔧(backend) settings CONVERSION_UPLOAD_ENABLED to control usage of docspec
- [x] 🔧(backend) expose CONVERSION_UPLOAD_ENABLED in config endpoint


lunika:
@coderabbit review

src/backend/core/api/viewsets.py
```diff
@@ -680,6 +680,11 @@ def perform_create(self, serializer):
         # Process it if present
         uploaded_file = serializer.validated_data.pop("file", None)
 
+        if uploaded_file and not settings.CONVERSION_UPLOAD_ENABLED:
+            raise drf.exceptions.ValidationError(
+                {"file": ["uploading a file is not allowed"]}
```
jmaupetit:
```suggestion
                {"file": ["file upload is not allowed"]}
```

src/backend/core/tests/documents/test_api_documents_create_with_file.py
```diff
@@ -81,14 +83,51 @@ def test_api_documents_create_with_docx_file_success(mock_convert):
...
+    """
+    user = factories.UserFactory()
+    client = APIClient()
+    client.force_login(user)
+
+    settings.CONVERSION_UPLOAD_ENABLED = False
+
+    # Mock the conversion
+    converted_yjs = "base64encodedyjscontent"
+    mock_convert.return_value = converted_yjs
```
jmaupetit: Not sure this is required as the mock is not expected to be called.

src/backend/core/tests/documents/test_api_documents_create_with_file.py
```diff
@@ -81,14 +83,51 @@ def test_api_documents_create_with_docx_file_success(mock_convert):
...
+    file = BytesIO(file_content)
+    file.name = "My Important Document.docx"
+
+    response = client.post(
+        "/api/v1.0/documents/",
+        {
+            "file": file,
+        },
+        format="multipart",
+    )
```
jmaupetit, line 107: Will it break without the `name` property?

```suggestion
    response = client.post(
        "/api/v1.0/documents/",
        {
            "file": BytesIO(b"fake docx content"),
        },
        format="multipart",
    )
```

lunika: Yes, the serializer is checking the file extension, we still have a 400 but not for the same reason.

PR #2155, opened 2026-03-30
🥚(frontend) add easter egg on doc emoji creation

AntoLC:
## Purpose

The first April, we add a easter egg on doc emoji creation. 
When the user creates a doc emoji an egg emoji is displayed instead of the page emoji.

## Demo

https://github.com/user-attachments/assets/93d6e177-6312-45d1-bec3-274032b1b394


Ash-Crow:
Wasn't it supposed to be a fish 🐟? (as April fools' day is called "poisson d'avril" in French-speaking countries)

PR #2162, opened 2026-04-01
⚡️(frontend) add jitter to WS reconnection

AntoLC:
## Purpose

When a massive simultaneous disconnection occurs (e.g. infra restart), all clients would reconnect and invalidate their queries at exactly the same time, causing a possible DB spike.
Adding random jitter spreads these events over a time window so the load is absorbed gradually.

```ts
const RECONNECT_BASE_DELAY_MS = 1000;
const RECONNECT_JITTER_MAX_MS = 3000;
```
Spread during 2 seconds.

### DIFF
--- a/src/frontend/apps/impress/src/features/docs/doc-management/stores/useProviderStore.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-management/stores/useProviderStore.tsx
@@ -30,7 +30,17 @@ const defaultValues = {
 
 type ExtendedCloseEvent = CloseEvent & { wasClean: boolean };
 
+/**
+ * When a massive simultaneous disconnection occurs (e.g. infra restart), all
+ * clients would reconnect and invalidate their queries at exactly the same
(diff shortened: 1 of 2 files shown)


src/frontend/apps/impress/src/features/docs/doc-management/stores/useProviderStore.tsx
```diff
@@ -63,7 +73,12 @@ export const useProviderStore = create<UseCollaborationStore>((set, get) => ({
           }
 
           clearTimeout(reconnectTimeout);
-          reconnectTimeout = setTimeout(() => void provider.connect(), 1000);
+
+          // Jitter spreading for reconnection attempts
+          reconnectTimeout = setTimeout(
+            () => void provider.connect(),
+            RECONNECT_BASE_DELAY_MS + Math.random() * RECONNECT_JITTER_MAX_MS,
```
lunika, line 82: You should explain the usage of `Math.random()` in both lines were you use it.

PR #2172, opened 2026-04-03
🐛(backend) Fix unreachable exception handler for URLValidator

maboukerfa:
## Purpose

The exception block was never being executed because URLValidator raises django.core.exceptions.ValidationError, not drf.exceptions.ValidationError, so the except block was dead code.

## Proposal
Replace the `drf.exceptions.ValidationError` with Django 'ValidationError'

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -2135,7 +2135,7 @@ class DocumentViewSet(
         url_validator = URLValidator(schemes=["http", "https"])
         try:
             url_validator(url)
-        except drf.exceptions.ValidationError as e:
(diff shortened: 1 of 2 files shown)


lunika:
You are right, the exception is managed by the exception handler but not caught here.
The test should be failing with this modification; you have to update them.

maboukerfa:
> You are right, the exception is managed by the exception handler but not caught here. The test should be failing with this modification; you have to update them.

Thank you! 
I updated the related test.
Screenshot 2026-04-04 at 15 38 19

lunika:
Thank you for your contribution.
I will rebase, fix the linting and merge it.

PR #2212, opened 2026-04-14
♿️(frontend) make doc search result labels uniquely identifiable

### LINKED ISSUE #2066: Search modal: results indistinguishable by screen readers
## Bug Report

**Problematic behavior**
When multiple documents have the same name (here three instances of "Copy of Docs Shortcuts"), each result is rendered with the same accessible name "Open document Copy of Docs Shortcuts". The last modified date ("17 hours ago", "1 day ago") that would allow them to be distinguished is hidden from assistive technologies via `aria-hidden="true"`.

**Expected behavior/code**
Each file must have a unique and sufficiently informative accessible name to be distinguished from others. The last modified date must be included in the accessible name or accessible via `aria-describedby`.

**Steps to Reproduce**
1. Create multiple documents with the same name
2. Enable a screen reader
3. Open the search modal and enter a term matching these documents
4. Navigate between results using arrow keys
5. Notice that the screen reader announces the same label for each result.

**Possible Solution**
The date must be integrated into the `aria-label` attribute of each file to complete the accessible name, while remaining visually hidden if the design requires it.

Image
Image

## Critères RGAA

[Critère 6.1 : Chaque lien est-il explicite ?](https://accessibilite.numerique.gouv.fr/methode/criteres-et-tests/#6.1)

## Impact

A blind user navigating search results with a screen reader hears "Open document Copy of Docs Shortcuts" multiple times in a row without being able to identify which one to open. They cannot distinguish between similar documents and risk opening the wrong document.

Ovgodd:
## Purpose

Fix an accessibility issue in document search results where multiple items with the same title were announced with identical labels by screen readers.

## Proposal

- [x] Include each document's relative last-update date in `SimpleDocItem` accessible label.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,14 +14,15 @@ and this project adheres to
 
 ### Changed
 
- - ♿️(frontend) structure correctly 5xx error alerts #2128
+- ♿️(frontend) structure correctly 5xx error alerts #2128
+- ♿️(frontend) make doc search result labels uniquely identifiable #2212
 
 ## [v4.8.6] - 2026-04-08
 
 ### Added
 
-- 🚸(frontend) allow opening "@page" links with 
-ctrl/command/middle-mouse click #2170
+- 🚸(frontend) allow opening "@page" links with
+  ctrl/command/middle-mouse click #2170
 - ✅ E2E - Any instance friendly #2142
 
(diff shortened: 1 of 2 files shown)


AntoLC (approved):
I let you see with the rabbit.

PR #2214, opened 2026-04-15
🐛(frontend) sanitize pasted toolbar links

### LINKED ISSUE #863: Copying a url includes the formatting characters
## Bug Report

**Problematic behavior**
When copying a url in the doc, it copies the formatting characters <>, so when you paste it in the url field, the link is not working and you're redirected to the same Docs page. 

**Expected behavior/code**
Being able to copy/paste the url as text, without the formatting characters.

**Steps to Reproduce**
1. Paste an url in a Doc and copy it
2. Then click a word, and the link icon, and add the url
3. See that the formatting characters are included
4. Click the link icon and see you're not redirected to your link 

Image

It's a similar issue than when copying a code formatted text https://github.com/suitenumerique/docs/issues/853

Ovgodd:
## Purpose

Fix broken links when pasting a URL into the BlockNote link toolbar input.
BlockNote copies links in Markdown autolink format (`<url>`), and the angle
brackets are kept as-is when pasting, resulting of invalid hrefs.

https://github.com/user-attachments/assets/2b2ff33e-1c12-4f75-bed5-66ac1dea77d4

## Proposal

* [x] Add a `onPasteCapture` handler that strips `<>` from pasted URLs in
  input/textarea fields within the editor container

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,17 +11,18 @@ and this project adheres to
 - 🚸(frontend) redirect on current url tab after 401 #2197
 - 🐛(frontend) abort check media status unmount #2194
 - ✨(backend) order pinned documents by last updated at #2028
+- 🐛(frontend) sanitize pasted toolbar links #2214
 
 ### Changed
 
- - ♿️(frontend) structure correctly 5xx error alerts #2128
+- ♿️(frontend) structure correctly 5xx error alerts #2128
 
 ## [v4.8.6] - 2026-04-08
 
 ### Added
 
-- 🚸(frontend) allow opening "@page" links with 
-ctrl/command/middle-mouse click #2170
+- 🚸(frontend) allow opening "@page" links with
+  ctrl/command/middle-mouse click #2170
 - ✅ E2E - Any instance friendly #2142
 
 ### Changed
--- a/src/frontend/apps/impress/src/features/docs/doc-editor/components/BlockNoteEditor.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-editor/components/BlockNoteEditor.tsx
@@ -88,6 +88,38 @@ interface BlockNoteEditorProps {
   provider: HocuspocusProvider;
 }
 
+/**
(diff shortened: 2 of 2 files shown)


AntoLC (commented):
I think we should open a issue on Blocknote because it is a problem on their side at the base.

I am not huge fan of the patch, because we fix the symptom (paste), but not the root cause (copy), if you paste the link somewhere else (e.g: in the browser url input) you still have the problem. 
Plus we are now overriding the paste event, even if it is targeting only the input, it could bring unexpected side effects that we will have to support.

The problem comes probably from this file, I suggest we either open a issue, or propose a fix on their side:
https://github.com/TypeCellOS/BlockNote/blob/main/packages/core/src/api/clipboard/toClipboard/copyExtension.ts#L176-L199

----

We could eventually merge this PR in waiting for the fix on the BN side, but we should let the issue opened until Blocknote made the fix, and then remove the patch.

----

Wdyt ?

Ovgodd:
> I think we should open a issue on Blocknote because it is a problem on their side at the base.
> 
> I am not huge fan of the patch, because we fix the symptom (paste), but not the root cause (copy), if you paste the link somewhere else (e.g: in the browser url input) you still have the problem. Plus we are now overriding the paste event, even if it is targeting only the input, it could bring unexpected side effects that we will have to support.
> 
> The problem comes probably from this file, I suggest we either open a issue, or propose a fix on their side: https://github.com/TypeCellOS/BlockNote/blob/main/packages/core/src/api/clipboard/toClipboard/copyExtension.ts#L176-L199
> 
> We could eventually merge this PR in waiting for the fix on the BN side, but we should let the issue opened until Blocknote made the fix, and then remove the patch.
> 
> Wdyt ?

Yes, I agree, we need to handle the root case. I’m ok with merging this for now, and I’ll create a PR on the BlockNote side.

EDI : pr on BN [side](https://github.com/TypeCellOS/BlockNote/pull/2661)

Ovgodd:
pr on BN https://github.com/TypeCellOS/BlockNote/pull/2661

PR #2231, opened 2026-04-20
🐛(frontend) fix app shalow reload

AntoLC:
## Purpose

The app was doing a shalow reload when user was coming from another tab and the user data was staled. We stop to block the app during the loading state, depend the response the app will manage correctly its states.


Ovgodd (approved):

AntoLC:
@CodeRabbit review

PR #2233, opened 2026-04-20
💬(frontend) add missing link in onboarding description

### LINKED ISSUE #2016: Onboarding modal is missing a link
## Bug Report

**Problematic behavior**
In the [figma](https://www.figma.com/design/qdCWR4tTUr7vQSecEjCyqO/Docs?node-id=12752-16817&t=4RnBKUzuYevWB050-0) the template library step should has a link to a templates doc.

Image

**Expected behavior/code**
We need to add a config for the admin to set the doc id of the template library. 
If no link is set, it shouldn't appear in the modal.  So we need to add some conditionnality here.

In our case the docs.numerique.gouv.fr the template library is [here](https://docs.numerique.gouv.fr/docs/6de1e458-088f-4c71-a58d-b8606e7b1b64/) 

**Steps to Reproduce**
1. Do this...
2. Then this...
3. And then the bug happens!

**Environment**
- Docs version:
- Instance url:

**Possible Solution**

AntoLC:
## Purpose

We added a missing link in the onboarding step description to direct users to ready-made templates f or common use cases. This enhancement aims to improve the user experience by providing easy access to resources that can help users get started quickly and customize their workflow efficiently.

### DIFF
--- a/src/frontend/apps/impress/src/features/help/hooks/useOnboardingSteps.tsx
+++ b/src/frontend/apps/impress/src/features/help/hooks/useOnboardingSteps.tsx
@@ -1,7 +1,8 @@
 import { type OnboardingStep } from '@gouvfr-lasuite/ui-kit';
 import Image from 'next/image';
-import { useTranslation } from 'react-i18next';
+import { Trans, useTranslation } from 'react-i18next';
 
+import { useConfig } from '@/core';
 import { useCunninghamTheme } from '@/cunningham';
 
 import DragIndicatorIcon from '../assets/drag_indicator.svg';
@@ -16,6 +17,9 @@ export interface OnboardingStepsData {
 
 export const useOnboardingSteps = () => {
   const { t } = useTranslation();
+  const { data: config } = useConfig();
+  const readyTemplateUrl =
+    config?.theme_customization?.onboarding?.ready_template_url;
   const { contextualTokens, colorsTokens } = useCunninghamTheme();
   const activeColor =
     contextualTokens.content.semantic.brand.tertiary ??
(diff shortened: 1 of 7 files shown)


src/frontend/apps/impress/src/features/help/hooks/useOnboardingSteps.tsx
```diff
@@ -122,8 +126,20 @@ export const useOnboardingSteps = () => {
...
+            components={{
+              Link: (
+                <a
+                  target="_blank"
+                  rel="noopener noreferrer"
+                  href={readyTemplateUrl}
+                />
+              ),
+            }}
+          />
```
Ovgodd, line 143: You can add an sr-only text, screen reader users should know that this link opens a new window.

```suggestion
          <>
          <Trans
            t={t}
            i18nKey="Start from <Link>ready-made templates</Link> for common use cases, then customize them to match your workflow in minutes."
            components={{
              Link: (
                <a
                  target="_blank"
                  rel="noopener noreferrer"
                  href={readyTemplateUrl}
                  aria-describedby="onboarding-template-link-new-window"
                />
             <span id="onboarding-template-link-new-window" className="sr-only">
              {t('new window')}
            </span>
              ),
            }}
          />
         </>
```

Ovgodd (changes requested):
Nice ! just a little change for better accessibility

src/frontend/apps/impress/src/features/help/hooks/useOnboardingSteps.tsx
AntoLC: What about that ?
https://github.com/suitenumerique/docs/blob/96ed8916de35755224229c0dc7c82419698b0503/src/frontend/apps/impress/src/features/help/hooks/useOnboardingSteps.tsx#L135-L140

Ovgodd: Yes, works for me. My initial idea with `aria-describedby` was to keep the visible link text as the accessible name and add the “opens in a new tab” part as extra context.

But in this case the `aria-label` is fine too, as long as it includes the visible link text, which it does here. Let’s go with your suggestion to keep the JSX simpler.

Ovgodd (approved):

Issue #2046, opened 2026-03-17
Sidebar is not snappy


robin-lecomte:
As shown in this GIF, switching tabs is quite slow and harms the user experience. The transition between subpages should be faster.

As a bonus, the button to collapse the sidebar blinks.

![Image](https://github.com/user-attachments/assets/e034b9ba-4838-4694-b571-cf2318b7c72e)

AntoLC:
We can add a skeleton, but the front is waiting for the data to come from the backend and the connection to be made to the yjs server, to bypass that brings side effects.

<img width="811" height="375" alt="Image" src="https://github.com/user-attachments/assets/4052e031-6223-474d-89f2-bb6018becd0a" />

We can display quicker the part 1, but the part 2 need a bit of time.

AntoLC:
"_The button to collapse the sidebar blinks_"solve here:  
- [ ] https://github.com/suitenumerique/docs/pull/2089

robin-lecomte:
The issue isn’t the loading on the right, but rather the responsiveness of the tree.

Moreover, in LaSuite Drive, the sidebar is snappy. But if I understand correctly, it’s supposed to be the same component from the UI kit?

AntoLC:
LaSuite Drive give the feeling to be "snappy" because there is usually less children to display (just the favoris), children are simpler (no emoji picker), a click does not lead to a connection to a collaboration server to display the content. 
Nevertheless the rerendering problem is in Drive as well, the tree dom is rebuilding many unnecessary times, as shown in the video:

https://github.com/user-attachments/assets/3da50129-d49d-4a02-816e-2041ed32f6ea

The root problem is how the tree is developed in the [ui-kit](https://github.com/suitenumerique/ui-kit/blob/main/src/components/tree-view/TreeView.tsx), Drive and Docs used the ui-kit for the tree. 
If we want to improve this part, it could lead to a important tree refacto.

virgile-dev:
I find that lately, Docs is not snappy in general, it's kind of slow at loading pages and everything. We should prioritize optimizing this @AntoLC

AntoLC:
Let's see what this PR brings (https://github.com/suitenumerique/docs/pull/2171), it should improve the loading page.

PR #2254, opened 2026-04-29
⚡️(frontend) add skeleton on content loading

AntoLC:
## Purpose

Content is longer to load than other parts of the doc page because of the connection with websocket to the collaboration server. To improve the user experience, we add a skeleton on the content part of the doc page.

It will improve the snappiness perception noted in this issue: https://github.com/suitenumerique/docs/issues/2046

## Demo

https://github.com/user-attachments/assets/93eeaf14-f980-4e36-b1ac-fb6f2e70e2eb

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ and this project adheres to
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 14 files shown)


Ovgodd (approved):
I like it

PR #2264, opened 2026-05-04
⚡️(frontend) close websocket connection when user change tab

lunika:
## Purpose

We want to track user activity on a doc and close the WebSocket connection when the user becomes inactive. 
When a user change to another tab, after a delay of "inactivity" we disconnect the user provider from the collaboration server.
When the user come back we reconnect to the server again. 
It will reduce the connection to the collaboration server and reduce outburst during reconnection during a ingress ngnix restart.

## Proposal

* [x] ✨(backend) new settings COLLABORATION_WS_INACTIVITY_TIMEOUT
* [x] ⚡️(frontend) close websocket connection when user change tab


AntoLC:
@CodeRabbit review

AntoLC (approved):

PR #2271, opened 2026-05-05
🐛(backend) manage race condition between GET and PATCH content

lunika:
## Purpose

When a PATCH and a GET on the content endpoint are made at the same time
for different users a race condition can happen and the metadata returned
by the S3 head_object can be outdated when the object is fetched leading
to an error raised because the Content-Length header does not match the
size of the response body. To avoid this, we no longer used head_object
followed bu get_object, we have to manage
everything in one call with the get_object. The get_object also accepts
as parameters an etag or last-modified header and will return a 304 if
the content has not changed, so we can use this to not return the entire
body if this one has not changed.

## Proposal

* [x] 🐛(backend) manage race condition between GET and PATCH content

### DIFF
--- a/src/backend/core/api/utils.py
+++ b/src/backend/core/api/utils.py
@@ -1,5 +1,6 @@
 """Util to generate S3 authorization headers for object storage access control"""
 
+import datetime as dt
 import time
 from abc import ABC, abstractmethod
 
@@ -199,3 +200,31 @@ class AIUserRateThrottle(AIBaseRateThrottle):
 def get_content_metadata_cache_key(document_id):
     """Return the cache key used to store content metadata."""
     return f"docs:content-metadata:{document_id!s}"
+
+
+def parse_http_conditional_headers(request):
+    """Extract and normalize `If-None-Match` and `If-Modified-Since`.
+
+    The `W/` weak prefix is stripped from the ETag because reverse proxies
+    (e.g. nginx with gzip) rewrite strong ETags into weak ones, which would
+    otherwise break a strict equality check in production.
+    """
+    if_none_match = request.META.get("HTTP_IF_NONE_MATCH")
+    if if_none_match and if_none_match.startswith("W/"):
+        if_none_match = if_none_match.removeprefix("W/")
+
+    if_modified_since_dt = None
+    if not (if_modified_since := request.META.get("HTTP_IF_MODIFIED_SINCE")):
+        return if_none_match, if_modified_since_dt
+
+    try:
+        if_modified_since_dt = dt.datetime.strptime(
+            if_modified_since, "%a, %d %b %Y %H:%M:%S %Z"
+        )
+    except ValueError:
+        if_modified_since_dt = None
+    else:
+        if not if_modified_since_dt.tzinfo:
+            if_modified_since_dt = if_modified_since_dt.replace(tzinfo=dt.timezone.utc)
+
+    return if_none_match, if_modified_since_dt
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -1941,11 +1941,12 @@ class DocumentViewSet(
         Retrieve the raw content file from s3 and stream it.
 
         We implement a HTTP cache based on the ETag and LastModified headers.
-        We retrieve the ETag and LastModified from the S3 head operation, save them in cache to
-        reuse them in future requests.
+        The ETag and LastModified are retrieved in the S3 get_object operation to be consistent with
+        the content Body retrieved at the same time. These metadata are saved in cache for
+        future requests.
         We check in the request if the ETag is present in the If-None-Match header and if it's the
-        same as the one from the S3 head operation, we return a 304 response.
-        If the ETag is not present or not the same, we do the same check based on the LastModifed
+        same as the one from the S3 get_object, we return a 304 response.
+        If the ETag is not present or not the same, we do the same check based on the LastModified
         value if present in the If-Modified-Since header.
         """
         document = self.get_object()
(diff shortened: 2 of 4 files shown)


src/backend/core/api/utils.py
```diff
@@ -199,3 +200,31 @@ def get_ident(self, request):
...
+            if_modified_since_dt = dt.datetime.strptime(
+                if_modified_since, "%a, %d %b %Y %H:%M:%S %Z"
+            )
+        except ValueError:
+            if_modified_since_dt = None
+        else:
+            if not if_modified_since_dt.tzinfo:
+                if_modified_since_dt = if_modified_since_dt.replace(
+                    tzinfo=dt.timezone.utc
+                )
```
jmaupetit:
```suggestion
    if not if_modified_since := request.META.get("HTTP_IF_MODIFIED_SINCE"):
        return if_none_match, if_modified_since_dt
        
    try:
        if_modified_since_dt = dt.datetime.strptime(
            if_modified_since, "%a, %d %b %Y %H:%M:%S %Z"
        )
    except ValueError:
        pass
    else:
        if not if_modified_since_dt.tzinfo:
            if_modified_since_dt = if_modified_since_dt.replace(
                tzinfo=dt.timezone.utc
            )
  ```

File: src/backend/core/api/viewsets.py [hunk 2]
```diff
@@ -1955,73 +1956,69 @@ def content_retrieve(self, request, *args, **kwargs):
...
+            ):
+                return drf_response.Response(status=status.HTTP_304_NOT_MODIFIED)
 
-        # Strip the W/ weak prefix. Proxies (e.g. nginx with gzip) convert strong
-        # ETags to weak ones, so a strict equality check would fail on production
-        # even when unchanged.
-        if if_none_match and if_none_match.startswith("W/"):
-            if_none_match = if_none_match.removeprefix("W/")
+        # Prepare get_object S3 operation. The get_object manages ETag and last_modified
+        # headers and will raise a 304 client error if one of them matches the value existing in
```
jmaupetit:
```suggestion
        # headers will raise a 304 client error if one of them matches the value existing in
```

src/backend/core/api/viewsets.py
```diff
@@ -1955,73 +1956,69 @@ def content_retrieve(self, request, *args, **kwargs):
...
+        # S3.
+        get_kwargs = {"Bucket": default_storage.bucket_name, "Key": document.file_key}
```
jmaupetit: Maybe it would be more explicit?

```suggestion
        get_object_kwargs = {"Bucket": default_storage.bucket_name, "Key": document.file_key}
```

src/backend/core/api/viewsets.py
```diff
@@ -1955,73 +1956,69 @@ def content_retrieve(self, request, *args, **kwargs):
...
+        cache_key = utils.get_content_metadata_cache_key(document.id)
+        if content_metadata := cache.get(cache_key):
+            if if_none_match and if_none_match == content_metadata.get("etag"):
+                return drf_response.Response(status=status.HTTP_304_NOT_MODIFIED)
+
+            if (
+                if_modified_since_dt
+                and dt.datetime.fromisoformat(content_metadata.get("last_modified"))
+                <= if_modified_since_dt
+            ):
```
jmaupetit:
```suggestion
            if (if_none_match and if_none_match == content_metadata.get("etag")) or (
                if_modified_since_dt
                and dt.datetime.fromisoformat(content_metadata.get("last_modified"))
                <= if_modified_since_dt
            ):
```

src/backend/core/api/viewsets.py
```diff
@@ -1955,73 +1956,69 @@ def content_retrieve(self, request, *args, **kwargs):
...
+        try:
+            s3_response = default_storage.connection.meta.client.get_object(
+                **get_kwargs
+            )
+        except ClientError as exc:
+            code = exc.response["Error"]["Code"]
+            if code in ("304", "PreconditionFailed", "NotModified"):
+                return drf_response.Response(status=status.HTTP_304_NOT_MODIFIED)
+            if code in ("NoSuchKey", "404"):
+                return StreamingHttpResponse(b"", content_type="text/plain", status=200)
```
jmaupetit: What about using the amazing match/case syntax? 😍 

```suggestion
            match code:
                case "304" | "PreconditionFailed" | "NotModified":
                    return drf_response.Response(status=status.HTTP_304_NOT_MODIFIED)
                case "NoSuchKey" | "404":
                    return StreamingHttpResponse(b"", content_type="text/plain", status=200)
                case _:
                    raise
```

src/backend/core/api/viewsets.py
```diff
@@ -1955,73 +1956,69 @@ def content_retrieve(self, request, *args, **kwargs):
...
+            raise
```
jmaupetit: if you use the match/case syntax this statement can be removed

src/backend/core/api/viewsets.py
```diff
@@ -1955,73 +1956,69 @@ def content_retrieve(self, request, *args, **kwargs):
...
+
+        last_modified = s3_response["LastModified"]
+        etag = s3_response["ETag"]
+        size = s3_response["ContentLength"]
+
+        # Refresh the metadata cache so future conditional requests can
+        # check them earlier
```
jmaupetit: Maybe this would be good enough?

```suggestion
        # Refresh the metadata cache
```

src/backend/core/api/viewsets.py
```diff
@@ -1955,73 +1956,69 @@ def content_retrieve(self, request, *args, **kwargs):
...
-            else:
-                if not since.tzinfo:
-                    since = since.replace(tzinfo=dt.timezone.utc)
-                if last_modified <= since:
-                    return drf_response.Response(status=status.HTTP_304_NOT_MODIFIED)
-
-        def _stream(file_key):
-            with default_storage.open(file_key, "rb") as f:
-                while chunk := f.read(8192):
+                while chunk := body.read(8192):
```
jmaupetit: Do you think the chunk size should be configurable?

src/backend/core/api/viewsets.py
```diff
@@ -1955,73 +1956,69 @@ def content_retrieve(self, request, *args, **kwargs):
...
                     yield chunk
+            finally:
+                body.close()
```
jmaupetit: Are you sure the try/finally pattern is required here to close the http response stream (as you return a generator)?

src/backend/core/tests/test_api_utils_parse_http_conditional_headers.py
```diff
@@ -0,0 +1,75 @@
...
+import datetime as dt
+
+from rest_framework.test import APIRequestFactory
+
+from core.api.utils import parse_http_conditional_headers
+
+
+def _request(**headers):
+    """Build a request with the given HTTP headers."""
+    return APIRequestFactory().get("/", headers=headers)
```
jmaupetit: Why not declaring a pytest fixture instead?

src/backend/core/tests/test_api_utils_parse_http_conditional_headers.py
```diff
@@ -0,0 +1,75 @@
...
+            **{
+                "if-none-match": 'W/"deadbeef"',
+                "if-modified-since": "Wed, 21 Oct 2015 07:28:00 GMT",
+            }
+        )
+    )
+    assert if_none_match == '"deadbeef"'
+    assert if_modified_since_dt == dt.datetime(
+        2015, 10, 21, 7, 28, 0, tzinfo=dt.timezone.utc
+    )
```
jmaupetit, line 50: All tests here could be configured as one single test using [indirect parametrization](https://docs.pytest.org/en/latest/example/parametrize.html#apply-indirect-on-particular-arguments) of the proposed fixture.

jmaupetit (commented):
I only have minor comments, else 👏

src/backend/core/api/viewsets.py
jmaupetit: Also to avoid using a while loop, you can use `iter_chunks`:


```suggestion
                for chunk in body.iter_chunks(chunk_size):
```

src/backend/core/api/viewsets.py
lunika: I thought it was not allowed by the Geneva Convention to use a switch in Python.

src/backend/core/api/viewsets.py
lunika: the iter_chunks is doing the same thing. I have no strong opinion. I would prefer to have a solution to have an async generator.

src/backend/core/api/viewsets.py
lunika: Nop I don't really know. I keep it from the previous implementation.

jmaupetit (approved):

PR #2272, opened 2026-05-05
🧵(backend) remove lock from db table

AntoLC:
## Purpose

To avoid race conditions when creating a new document, we used to create a lock on the database table the time of the creation, to prevent concurrent creations and duplicate path keys.
This approach was not ideal as it could lead to deadlocks and performance issues.
We now catch the integrity error raised by the database when a duplicate path key is created, and we retry the creation with a new path key.
This approach is more efficient as it is retrying only when there is a conflict, and it does not require locking the entire table.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ and this project adheres to
 
 - ⚡️(frontend) add skeleton on content loading #2254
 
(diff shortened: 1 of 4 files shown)


lunika:
Managed in #2274

PR #2273, opened 2026-05-06
🐛(frontend) fix loading comments transaction

### LINKED ISSUE #2265: 🐛Multiple PATCH Save
## Bug Report

When a doc has comments, the guard to check if there is something to save seems to not work correctly.
We should fix it asap: 
https://github.com/suitenumerique/docs/blob/4a0f51212fb7a8e423dfa362b516a033304781d7/src/frontend/apps/impress/src/features/docs/doc-editor/hook/useSaveDoc.tsx#L68

AntoLC:
## Purpose

When we load the comments we have to notify the subscribers of the `DocsThreadStore`. This generates a Yjs transaction that is currently treated as a user-initiated content change that will trigger a patch request when the doc will try to save.
We now update the transaction origin when we notify the subscribers so that we can reliably identify and ignore those transactions in the `useSaveDoc` hook.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to
 
 ### Fixed
 
+- 🐛(frontend) fix patch and comments #2273
 - 🐛(frontend) interlinking are exported correctly in print mode #2269
 - 💬(frontend) add missing link in onboarding description #2233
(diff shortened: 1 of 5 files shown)


AntoLC:
@CodeRabbit review

Ovgodd (approved):

PR #2274, opened 2026-05-06
🐛(backend) replace document creation table locks with retry strategy

lunika:
## Purpose

We have situation where the number of locks in the database can increase
dangerously creating deadlock situation. To remove this situation we
decided to change the strategy to manage document creation concurrency.
We decided to use a retry strategy, trying to create the document
multiple times while a usable path is found. To avoid having an
inifinite loop, we use a max_attempts counter configurable using the
setting TREEBEARD_PATH_COMPUTE_RETRY_MAX_ATTEMPTS

## Proposal

* [x] 🐛(backend) replace document creation table locks with retry strategy
* [x] ♻️(backend) split core/utils.py module


src/backend/core/utils/treebeard.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Treebeard computes the materialized path by reading the current siblings;
+    under concurrency two callers may compute the same value. Depending on
+    timing this surfaces either as:
+
+    - `django.core.exceptions.ValidationError` raised by `full_clean()` /
+      `validate_unique()` before the INSERT (BaseModel.save calls full_clean),
+    - or `IntegrityError` from the database unique index when the validate
+      step misses the conflict.
+    """
```
jmaupetit, line 28: Can you also add example messages so that it's more explicit?

src/backend/core/utils/treebeard.py
```diff
@@ -0,0 +1,56 @@
...
+            with transaction.atomic():
+                return create_fn()
+        except (IntegrityError, DjangoValidationError) as exc:
+            if not _is_tree_path_collision(exc) or attempt == max_attempts - 1:
+                raise
+            logger.info(
+                "tree path collision on attempt %d/%d, retrying",
+                attempt + 1,
+                max_attempts,
+            )
```
jmaupetit, line 59: Shouldn't we temporize a bit?

```suggestion
            )
            time.sleep(attempt * 0.1)
```

lunika: Yes you are right, I will do it.

jmaupetit (approved):

PR #2275, opened 2026-05-06
🏗️(core) migrate from pip to uv

lunika:
## Purpose

We want to migrate our projects from pip to uv to take the benefits of the lock file and have reproducible installations. A first uv.lock file is committed, and the Dockerfile and compose are modified to work with uv

## Proposal

* [x] 🏗️(core) migrate from pip to uv
* [x] ♻️(backend) migrate from setuptool to uv_build as build backend
* [x] 🔧(actions) migrate from pip to uv

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,24 +1,38 @@
 # Django impress
 
 # ---- base image to inherit from ----
-FROM python:3.13.3-alpine AS base
-
-# Upgrade pip to its latest release to speed up dependencies installation
-RUN python -m pip install --upgrade pip
+FROM python:3.13.13-alpine AS base
 
 # Upgrade system packages to install security updates
 RUN apk update && apk upgrade --no-cache
 
+# Upgrade pip to its latest release to speed up dependencies installation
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -1,24 +1,38 @@
...
-# Upgrade pip to its latest release to speed up dependencies installation
-RUN python -m pip install --upgrade pip
+FROM python:3.13.13-alpine AS base
 
 # Upgrade system packages to install security updates
 RUN apk update && apk upgrade --no-cache
 
+# Upgrade pip to its latest release to speed up dependencies installation
+# We must do that to avoid having an outdated pip version with security issues
+RUN python -m pip install --upgrade pip
```
jmaupetit: Are you sure it's still required as `uv` provides a [Pip interface](https://docs.astral.sh/uv/pip/)?

jmaupetit (approved):

Dockerfile
lunika: pip is not used anymore, this first comment line can be removed to keep only the second one. Or we could probably remove pip.

PR #2276, opened 2026-05-07
🩺(project) reload app if front and back unsync

### LINKED ISSUE #2209: 🐛Frontend and backend unsync
## Bug Report

In some cases, mainly after a release, the frontend can be unsync with the backend. It can lead to errors because of types changing.

## Solution

To fix that, the backend could expose the current version from the config endpoint, if the frontend has a different version than the one in the endpoint, we can force a hard refresh to force the sync.

AntoLC:
## Purpose

We observe some cases where the frontend and backend versions can get out of sync, which can cause issues.
To mitigate this, we want to implement a mechanism that detects when the frontend and backend versions are mismatched and triggers a reload of the application to ensure they are in sync.

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -2841,6 +2841,7 @@ class ConfigView(drf.views.APIView):
                 dict_settings[setting] = getattr(settings, setting)
 
         dict_settings["theme_customization"] = self._load_theme_customization()
+        dict_settings["RELEASE_VERSION"] = settings.RELEASE
 
         return drf.response.Response(dict_settings)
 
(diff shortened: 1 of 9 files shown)


AntoLC:
@coderabbit review

src/backend/core/api/viewsets.py
```diff
@@ -2855,6 +2857,7 @@ def get(self, request):
                 dict_settings[setting] = getattr(settings, setting)
 
         dict_settings["theme_customization"] = self._load_theme_customization()
+        dict_settings["RELEASE_VERSION"] = get_release()
```
```suggestion
        dict_settings["RELEASE_VERSION"] = settings.RELEASE
```

src/backend/core/api/viewsets.py
```diff
@@ -73,6 +73,8 @@
     users_sharing_documents_with,
 )
 
+from impress.settings import get_release
```
```suggestion
```

src/backend/core/tests/test_api_config.py
```diff
@@ -46,7 +46,8 @@ def test_api_config(is_authenticated):
         user = factories.UserFactory()
         client.force_login(user)
 
-    response = client.get("/api/v1.0/config/")
+    with patch("core.api.viewsets.get_release", return_value="1.0.0"):
+        response = client.get("/api/v1.0/config/")
```
lunika: You can change `RELEASE` in the `overrive_settings`

lunika (approved):

PR #2280, opened 2026-05-07
♻️(ci) use arm64 runner to build images for arm64 architecture

lunika:
## Purpose

We have trouble building Docker images for the ARM64 architecture when we use
QEMU emulation. We want to try building them using an arm64 native
runner

## Proposal

* [x] ♻️(ci) use arm64 runner to build images for arm64 architecture
* [x] 🔧(ci) enable trivy step

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -41,23 +41,17 @@ permissions:
   contents: read
 
 jobs:
-  build-and-push:
+  meta:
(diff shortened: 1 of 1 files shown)


jmaupetit (approved):
Neat!

StephanMeijer:
Yeah, this is the way.

StephanMeijer (approved):

PR #2291, opened 2026-05-12
✅(e2e) fix some flakiness

AntoLC:
## Purpose

Strong flakiness started to appear.
This commit aims to fix some of them by improving the stability of the tests and removing some redundant checks.

### DIFF
--- a/env.d/development/common.e2e
+++ b/env.d/development/common.e2e
@@ -1,7 +1,7 @@
 # For the CI job test-e2e
-BURST_THROTTLE_RATES="200/minute"
+BURST_THROTTLE_RATES="1000/minute"
 COLLABORATION_API_URL=http://y-provider:4444/collaboration/api/
-SUSTAINED_THROTTLE_RATES="200/hour"
+SUSTAINED_THROTTLE_RATES="1000/minute"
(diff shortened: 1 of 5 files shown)


env.d/development/common.e2e
```diff
@@ -1,7 +1,7 @@
 # For the CI job test-e2e
-BURST_THROTTLE_RATES="200/minute"
+BURST_THROTTLE_RATES="1000/minute"
 COLLABORATION_API_URL=http://y-provider:4444/collaboration/api/
-SUSTAINED_THROTTLE_RATES="200/hour"
+SUSTAINED_THROTTLE_RATES="1000/minute"
```
Ovgodd, line 4: This is not to big ? from 200/hour to 1000/minute ? is it intended ?

Ovgodd (approved):
cool ! just one question but it is ok for me !

env.d/development/common.e2e
AntoLC: Yes by purpose, I can see a test got stuck because of a throttle, so I increase them a lot to be sure we will not be annoyed by them anymore. This config file is just for the e2e tests, and we don't aim to test the throttle with the e2e tests.

PR #2296, opened 2026-05-17
✨(y-provider) preserve custom blocks on HTML/markdown conversion

sylvinus:
Wire the docs BlockNote schema (callout, pdf, uploadLoader, interlinking link, page break) into the conversion editor so /api/convert no longer drops or mangles these blocks.

This is important for the Docs as CMS usecase, it is surprising for users to see these Docs blocks silently ignored in the output.

### DIFF
--- a/src/frontend/servers/y-provider/src/blockSpecs/InterlinkingLinkInline.ts
+++ b/src/frontend/servers/y-provider/src/blockSpecs/InterlinkingLinkInline.ts
@@ -1,0 +1,61 @@
+import { createInlineContentSpec } from '@blocknote/core';
+
+const interlinkingPropSchema = {
+  docId: { default: '' as string },
+  disabled: {
+    default: false as boolean,
+    values: [true, false] as const,
+  },
+  trigger: {
+    default: '/' as const,
+    values: ['/', '@'] as const,
+  },
+  title: { default: '' as string },
+} as const;
+
+const interlinkingConfig = {
+  type: 'interlinkingLinkInline' as const,
+  propSchema: interlinkingPropSchema,
+  content: 'none' as const,
+};
+
+// Matches the frontend route (LinkSelected.tsx) so exported HTML/markdown
+// links resolve identically client-side.
+const interlinkingHref = (docId: string) =>
+  `/docs/${encodeURIComponent(docId)}/`;
+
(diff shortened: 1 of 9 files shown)


sylvinus:
@lunika @AntoLC we deployed this on ANCT's instance and it works correctly, we now have custom blocks in exports :) Still open to your feedbacks though of course!

src/frontend/servers/y-provider/src/blockSpecs/InterlinkingLinkInline.ts
```diff
@@ -0,0 +1,61 @@
...
+const interlinkingConfig = {
+  type: 'interlinkingLinkInline' as const,
+  propSchema: interlinkingPropSchema,
+  content: 'none' as const,
+};
+
+// Matches the frontend route (LinkSelected.tsx) so exported HTML/markdown
+// links resolve identically client-side.
+const interlinkingHref = (docId: string) =>
+  `/docs/${encodeURIComponent(docId)}/`;
```
AntoLC: Isn't it a problem to not have the domain ? 
The link will probably target something that does not exist, no ?

AntoLC (commented):
It is quite error prone, as any change frontend side would have to be made here as well, and nothing is asserting that it is sync correctly.
The best is too create common packages, but personally I don't have the time to do it for the moment.

On my side we can still merge it in waiting to have time to improve, I will create a issue about improvement.

----

Last observation, we are thinking to move the converter from this server (`y-provider`) to the `docs-spec` server, to keep this server dedicated to the collaboration, `docspec` is already doing conversion (docx to markdown) so it will be dedicated for everything about conversion.

----

I approved it: ✅ 
I let @lunika do the final check to be sure it is fine on his side as well.

src/frontend/servers/y-provider/src/blockSpecs/InterlinkingLinkInline.ts
sylvinus: On our side we parse such "/docs/xxx" links and apply our own mapping. You should we should add the instance base URL as well?

sylvinus:
I just updated this branch to avoid comments making some blocks disappear from the output. It's working correctly now on anct's instance.

AntoLC (approved):
Nice.
I will merge it this week.

PR #2305, opened 2026-05-20
🐛(backend) use computed_link_reach in handle_onboarding_document

### LINKED ISSUE #2015: Sub-doc are not added as favourite at onboarding
## Bug Report

**Problematic behavior**
https://github.com/suitenumerique/docs/pull/1971  creates an exception for restricted document.
Sub-docs for which `computed_link_reach` is public are not added.

**Expected behavior/code**
The exception should not use  `link_reach` but `computed_link_reach` instead. So that sub-docs whose parent `link_reach` is set to to public can be added as favourite at onboarding. 

**Steps to Reproduce**
1. Do this...
2. Then this...
3. And then the bug happens!

**Environment**
- Docs version: 4.8

**Possible Solution**

lunika:
## Purpose

In the model method User::_handle_onboarding_documents_access we do not allow using documents with restricted link_reach to be added as onboarding documents. To check the real link_reach of the document, we must use instead the computed_link_reach to be sure that a sub document can also be used and compute its correct link_reach.

## Proposal

* [x] 🐛(backend) use computed_link_reach in handle_onboarding_document

Fix #2015

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to
 
 - 🐛(docs) run migration 0027 without superuser role
 - 🐛(backend) prevent admins/owners from overwriting other users comments
+- 🐛(backend) use computed_link_reach in handle_onboarding_document #2305
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


lunika:
@coderabbit review

jmaupetit (approved):

PR #2325, opened 2026-05-26
♿️(frontend) improve keyboard tab order in document list

### LINKED ISSUE #2308: Improve keyboard navigation in the “All documents” list
## Feature Request

In the “All documents” list, the tab order is not coherent enough for keyboard users.

Currently, users need to press `Tab` several times before reaching the next document.

This makes navigation slower and harder to understand, especially when the list contains many documents.

12.8 – Dans chaque page web, l’[ordre de tabulation](https://accessibilite.numerique.gouv.fr/methode/glossaire/#ordre-de-tabulation) est-il [cohérent](https://accessibilite.numerique.gouv.fr/methode/glossaire/#comprehensible-ordre-de-lecture)

**Audit "Accueil(connecté) - Docs"**

Improve the keyboard navigation pattern in the document list.

The expected behavior is:

- one `Tab` stop to reach the document item;
- one additional `Tab` stop to reach the actions related to that document;
- then the next `Tab` should move to the next document item.

When the document item is reached, the main information should be announced:

- document name;
- last update date;
- number of participants who have access to the document.

This would make the list easier to navigate and understand for keyboard and screen reader users.

This should be checked in the “All documents” list.

The expected navigation pattern should be:

```text
Document item → Document actions → Next document item
```

Testing should be done with keyboard-only navigation and screen reader navigation.

Yes.

Ovgodd:
## Purpose

Improve keyboard and screen reader navigation in the "All documents" list. Tab order was inconsistent: users had to press Tab several times before reaching the next document. Each row should expose one stop for the document item, then one for its actions.

## Proposal

* [x] Reduce tab stops per row to **document link > actions menu > next document**
* [x] Announce **document name**, **last update date**, and **participant count** on the document link via `aria-label`
* [x] Hide the decorative date link from assistive technologies (`aria-hidden`)
* [x] Update e2e selectors to match the new accessible name format

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ and this project adheres to
 - 🐛(docs) run migration 0027 without superuser role
 - 🐛(backend) prevent admins/owners from overwriting other users comments
 
+### Changed
+
+- ♿️(frontend) improve keyboard tab order in document list #2325
 
 ## [v5.1.0] - 2026-05-11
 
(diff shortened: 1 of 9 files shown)


AntoLC (commented):
As talk together, let's see if we can harmonize keyboard navigation between feature.

PR #2328, opened 2026-05-26
🐛(y-provider) return empty output when converting empty Yjs document

sylvinus:
The convert handler returned 500 "No valid blocks were generated" whenever the reader produced an empty blocks array, which is the normal state of a freshly-created Yjs document. Treat empty input as a valid case and return 200 with an empty body for every supported output format.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to
 
 - 🐛(docs) run migration 0027 without superuser role
 - 🐛(backend) prevent admins/owners from overwriting other users comments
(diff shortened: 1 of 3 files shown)


zeylos:
tested successfuly in production on our instance
image

AntoLC:
Why calling this webservice with an empty body ?

sylvinus:
@AntoLC because our CMS fetches all pages without knowing in advance if some are empty or not :) (the parent pages are often empty)

AntoLC (approved):

PR #2358, opened 2026-05-26
⚡️(helm) create a dedicated svc and deployment for yprovider converter

lunika:
## Purpose

We want to isolate the converter service built in the yprovider server in order to not mix conversion and collaboration together. We decided to create a dedicated service and deployment, by default they use the values defined in yProvider values but all can be overridden in yProvider.converter.
Also, by default these new service and deployment are not enabled, they must be explicitly enabled and configured. If not it works like now with only one service and deployment for the yProvider server.

## Proposal

* [x] ⚡️(helm) create a dedicated svc and deployment for yprovider converter

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to
 - ✨(frontend) comment side panel #2279
 - ✨(buildpack) add PaaS deployment support, tested with Scalingo #2293
 - 🔧(backend) allow configuring settings OIDC_OP_USER_ENDPOINT_FORMAT
+- ⚡️(helm) create a dedicated svc and deployment for yprovider converter #2368
 
 ### Changed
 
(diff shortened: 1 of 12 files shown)


AntoLC:
@coderabbit review

AntoLC (commented):
Should we replicate this configuration in the docker environment as well, to stay more or less iso with the production, 1 container about collab and 1 about conversion ?

AntoLC (approved):
If you search `YPROVIDER_HOST`, I see some places that maybe need changes as well:
https://github.com/suitenumerique/docs/blob/a20ba1a78efcb6a9ad3ac8bbf70ec846f0cbce2f/env.d/production.dist/common#L6
https://github.com/suitenumerique/docs/blob/8655750dca4a8488c2128ec393afb9617a317736/docs/examples/helm/impress.values.yaml#L71

----

A bit out of scope, but I noticed that `COLLABORATION_API_URL` is not in the helm charts (dev/feature).

----

I tried it in local as well, works great !

lunika:
> A bit out of scope, but I noticed that COLLABORATION_API_URL is not in the helm charts (dev/feature).

It is not because the collaborative check edition is not enabled.
I will do it in a separated commit

PR #2375, opened 2026-06-02
🐛(frontend) fix emoji pdf not matching

### LINKED ISSUE #2263: 🐛 Export svg not match
## Bug Report

➡️ 
I think this svg by example is not the same when we export, it gives someone dancing, try to figure out why and fix this behavior.

AntoLC:
## Purpose

The emoji algorithm to find the emoji was not totally correct, it could choose a bad family of emojis.
The first part of the emoji "code" is what should be used first to find the "family" of the emoji, we can then try to find the exact match in that specific family.


Ovgodd (approved):
oh nice, it works well !
tested with, "sad""smile""spider"

PR #2379, opened 2026-06-02
♿️(frontend) use semantic <p> elements in document info card

### LINKED ISSUE #2350: Document information card: unstructured information in paragraph elements
**Observed behavior**

In the document information card, text elements such as the user's role (owner, editor, reader) and the last update date are not contained in `<p>` elements. This information is therefore rendered without appropriate semantic structure.

**Expected behavior**

Each piece of textual information must be encapsulated in a `<p>` element in order to be properly structured and rendered by assistive technologies.

**Steps to reproduce**

1. Open the information card of a document.
2. Inspect the source code of the elements displaying the user's role and the last update date.
3. Observe that these contents are not in `<p>` elements.

**RGAA criteria**

[Critère 9.1 : Dans chaque page web, l'information est-elle structurée par l'utilisation appropriée de titres ?](https://accessibilite.numerique.gouv.fr/methode/criteres-et-tests/#9.1)

**Prioritization**

P2

Ovgodd:
## Purpose

Use a semantic `<dl>` (definition list) instead of `<p>` for the user role and last update date in the document information card (`DocHeaderInfo`), as suggested in review. These are key/value metadata pairs, not paragraphs.

## Proposal

* [x] Wrap the info card in a `<dl>` via `Box as="dl"`
* [x] Add screen-reader-only `<dt>` labels for each metadata entry
* [x] Render role and date values as `<dd>` via `Text as="dd"`
* [x] Extend `Text` component `as` prop to accept `dt` and `dd`
* [x] Reset native `<dl>`/`<dd>` margins to preserve existing layout

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ and this project adheres to
 
 - 🐛(frontend) redirect homepage to login when homepage feat is disabled #2521
 
+### Changed
+
+- ♿️(frontend) use semantic `<dl>` structure in document info card #2379
+
 ## [v5.4.1] - 2026-07-09
 
 ### Changed
--- a/src/frontend/apps/impress/src/components/Text.tsx
+++ b/src/frontend/apps/impress/src/components/Text.tsx
@@ -9,7 +9,18 @@ const { sizes } = tokens.themes.default.globals.font;
 type TextSizes = keyof typeof sizes;
 
 export interface TextProps extends BoxProps {
-  as?: 'p' | 'span' | 'div' | 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
+  as?:
+    | 'p'
+    | 'span'
+    | 'div'
+    | 'h1'
+    | 'h2'
+    | 'h3'
+    | 'h4'
+    | 'h5'
+    | 'h6'
+    | 'dt'
+    | 'dd';
   $ellipsis?: boolean;
   $weight?: CSSProperties['fontWeight'];
   $textAlign?: CSSProperties['textAlign'];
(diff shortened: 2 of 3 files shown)


AntoLC (commented):
I don't think it is a correct semantic here, "[p](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/p)" means paragraph, it is means to separate block of text, it is why it has a even a native margin attribute to it. If we want to add "p" it should be more on the "Box" component so.

That said, if we want to be top notch semantic, we should use something more appropriate like [dl](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dl) and its friends.

Ovgodd:
> I don't think it is a correct semantic here, "[p](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/p)" means paragraph, it is means to separate block of text, it is why it has a even a native margin attribute to it. If we want to add "p" it should be more on the "Box" component so.
> 
> That said, if we want to be top notch semantic, we should use something more appropriate like [dl](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dl) and its friends.

@cyberbaloo what solution do you prefer / is better ?

CHANGELOG.md
```diff
@@ -10,11 +10,20 @@ and this project adheres to
...
 ## [v5.4.1] - 2026-07-09
 
 ### Changed
 
 - ♻️(backend) reset collaboration connection in cascade for all children #2507
+- ♻️(backend) allow global search in sub documents
+- ✨(backend) add a breadcrumb in the search response
+- ♻️(frontend) move doc action buttons to fix toolbar #2360
+- ♿️(frontend) add aria-hidden to decorative avatar SVGs in share modal #2324
+- 🏗️(frontend) move comments to its own folder feature #2374
```
AntoLC: I don't think these changes should be introduced.

AntoLC (commented):
Like says the rabbit, `dateToDisplay` already contains text informations, see:
https://github.com/suitenumerique/docs/pull/2379#discussion_r3586169505

https://github.com/suitenumerique/docs/blob/76dd9c1dd3eef173006644d00f26fc5bfc8bad70/src/frontend/apps/impress/src/features/docs/doc-header/components/DocHeaderInfo.tsx#L29-L41

So you should maybe separate the text (Last update: / Days remaining:) inside a `dt` tag directly.

AntoLC (approved):

PR #2383, opened 2026-06-02
♿️(frontend) improve presenter mode screen reader and keyboard support

Ovgodd:
## Purpose

Make the presenter mode accessible to keyboard and screen reader users:
announce slide changes, trap focus inside the overlay, and restore focus
to the trigger button on close.

## Proposal

* [x] Announce "Slide X of Y: <title>" on navigation via a `polite` live region (`@react-aria/live-announcer`).
* [x] Remove the redundant slide-counter `<span>` label: the live announcer now handles it, so keeping the `aria-label` made screen readers read the same info twice. The visual counter is hidden with `aria-hidden="true"`.
* [x] Trap focus inside the overlay with react-aria `FocusScope contain`.
* [x] Strip the ProseMirror editor role (`textbox`/`contenteditable`, `tabindex="-1"`) so slides aren't announced as editable, keeping links tabbable.
* [x] Update e2e tests to target content via `role="group"`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,9 @@ and this project adheres to
 - 👷(CI) remove test-e2e-other-browser job #2404
 - ♿️(frontend) use heading element for pinned documents section title #2380
 - ♿️(frontend) use anchor links for table of contents entries #2390
+- ♿️(frontend) improve presenter mode screen reader and keyboard support #2383
 
-### Fixed 
+### Fixed
 
 - 🐛(frontend) overlap of block menu dropdown #2406
 
--- a/src/frontend/apps/impress/src/features/docs/doc-header/components/DocToolBox.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-header/components/DocToolBox.tsx
@@ -135,7 +135,9 @@ export const DocToolBox = ({ doc }: DocToolBoxProps) => {
       label: t('Present'),
       icon: <Present width={24} height={24} aria-hidden="true" />,
       callback: () => {
-        setIsPresenterOpen(true);
+        requestAnimationFrame(() => {
+          setIsPresenterOpen(true);
+        });
       },
       isHidden: Boolean(doc.deleted_at) || isMobile,
       testId: `docs-actions-present-${doc.id}`,
@@ -284,7 +286,6 @@ export const DocToolBox = ({ doc }: DocToolBoxProps) => {
           doc={doc}
           onClose={() => {
             setIsPresenterOpen(false);
-            restoreFocus();
           }}
         />
       )}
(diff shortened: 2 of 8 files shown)


src/frontend/apps/impress/src/features/docs/doc-header/components/DocToolBox.tsx
AntoLC, line 1: Natively react-aria manage this part I get the feeling, no need to add the last focus.

src/frontend/apps/impress/src/features/docs/doc-presenter/hooks/useSlides.ts
```diff
@@ -40,3 +40,42 @@ export const splitBlocksIntoSlides = <T extends Block>(blocks: T[]): T[][] => {
...
+export const getSlideTitle = (
+  blocks: { type: string; content?: unknown }[],
+): string => {
+  const heading = blocks.find((block) => block.type === 'heading');
+  const source =
+    heading ??
+    blocks.find((block) => extractInlineText(block.content).trim().length > 0);
+
+  return source ? extractInlineText(source.content).trim() : '';
+};
```
AntoLC, line 81: Blocknote does not have a funtion to do it ? 

Could you add a test e2e to assert that it is working as expected on this part ?

Ovgodd: I found [this discussion on BlockNote](https://github.com/TypeCellOS/BlockNote/discussions/2482) about a blocksToPlainTextLossy API, it was requested but never implemented. 

we could use getText() from tiptapEditor but it returns the full document text and we need the title of each individual slide, I did not find another function for this.

As for the test, did you mean a unit test for getSlideTitle / extractInlineText? I've added one in useSlides.spec.ts covering heading priority, paragraph fallback, nested inline content (links), empty slides, and whitespace trimming (9 cases, all green). 

Or did you mean just asserting the title inside the existing e2e test "announces the slide position through a live region"? I've updated that too, it now checks the full message including the title (e.g. "Slide 1 of 3: Slide one" )

src/frontend/apps/impress/src/features/docs/doc-header/components/DocToolBox.tsx
Ovgodd: Good catch! I switched to react-aria's `restoreFocus `prop on `FocusScope`. It worked on first open but failed on the others,

FocusScope was capturing the "Present" menu item (destroyed when the dropdown closes).

Fixed by closing the dropdown first and deferring the presenter open to a requestAnimationFrame, giving the dropdown time to restore focus to its trigger before FocusScope captures it.

CHANGELOG.md
```diff
@@ -25,6 +24,7 @@ and this project adheres to
 - ♻️(frontend) move doc action buttons to fix toolbar #2360
 - ♿️(frontend) add aria-hidden to decorative avatar SVGs in share modal #2324
 - 🏗️(frontend) move comments to its own folder feature #2374
+- ♿️(frontend) improve presenter mode screen reader and keyboard support #2383
```
AntoLC: When rebase, don't forget to add it under `Unreleased`

AntoLC (approved):

PR #2384, opened 2026-06-03
♿️(frontend) align search modal field label with placeholder

### LINKED ISSUE #2347: Search modal — Search field: inconsistent label with placeholder
**Observed behavior**

The search field has two competing labels:

- an `aria-label="Saisie de recherche rapide"` directly on the `<input>` element, which takes priority over the `<label>` in calculating the accessible name;
- a `<label for="...">Rechercher des documents</label>` correctly coupled to the field via `id` and `for`.

The only visible text associated with the field is the `placeholder` "Tapez le nom d'un document". Voice restitution is therefore different from the text displayed on screen, and the three formulations are inconsistent with each other.

**Expected behavior**

The field must have a single label, carried by the `<label>` element, whose formulation is consistent with the `placeholder`. The `aria-label` attribute on the `<input>` must be removed. The `placeholder` can complement the label but cannot replace it.

**Steps to reproduce**

1. Open the search modal.
2. Navigate with a screen reader to the search field.
3. Observe that voice restitution differs from visible text.

**Possible solution**

**Before:**

```html
<label for="search-input">Rechercher des documents</label>
<input id="search-input" aria-label="Saisie de recherche rapide" placeholder="Tapez le nom d'un document">
```

**After:**

```html
<label for="search-input">Tapez le nom d'un document</label>
<input id="search-input" placeholder="Tapez le nom d'un document">
```

**RGAA criteria**

[Critère 11.1 : Chaque champ de formulaire a-t-il une étiquette ?](https://accessibilite.numerique.gouv.fr/methode/criteres-et-tests/#11.1)
[Critère 11.2 : Chaque étiquette associée à un champ de formulaire est-elle pertinente ?](https://accessibilite.numerique.gouv.fr/methode/criteres-et-tests/#11.2)

**Impact**

A visually impaired user or a user with a cognitive disability starts typing a word in the search field. The visible text disappears and voice restitution does not match what is displayed on screen. They may no longer know what they are filling in.

**Prioritization**

P1

Ovgodd:
## Purpose

The search modal input exposed three inconsistent names (`aria-label`, `<label>`, and `placeholder`). Screen reader users heard a different label than the visible placeholder

## Proposal

- [x] Remove redundant `aria-label` on `QuickSearchInput` so the accessible name comes from cmdk's `<label>` (`aria-labelledby`) instead of overriding it
- [x] Align `DocSearchModal` `label` and `placeholder` on `Type the name of a document`
- [x] Update e2e selectors in `doc-search.spec.ts` to match the new combobox accessible name

### DIFF
--- a/src/frontend/apps/impress/src/components/quick-search/QuickSearch.tsx
+++ b/src/frontend/apps/impress/src/components/quick-search/QuickSearch.tsx
@@ -65,7 +65,6 @@ export const QuickSearch = ({
         >
           {showInput && (
             <QuickSearchInput
-              label={label}
               withSeparator={hasChildrens(children)}
               inputValue={inputValue}
               onFilter={onFilter}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to
 - ♿️(frontend) enable blocknote heading ids for toc anchors #2449
 - ♿️(frontend) focus export modal on format select #2421
 - ♿️(frontend) configurable legal submenu in HelpMenu, remove Crisp #2416
+- ♿️(frontend) align search modal field label with placeholder #2384
 
 ## [v5.3.0] - 2026-06-19
 
--- a/src/frontend/apps/e2e/__tests__/app-impress/doc-grid-move.spec.ts
+++ b/src/frontend/apps/e2e/__tests__/app-impress/doc-grid-move.spec.ts
@@ -213,7 +213,7 @@ test.describe('Doc grid move', () => {
         .getByRole('heading', { name: 'Choose a new parent doc' }),
     ).toBeVisible();
 
-    const input = page.getByRole('combobox', { name: 'Search' });
+    const input = page.getByRole('combobox', { name: 'Search for a doc...' });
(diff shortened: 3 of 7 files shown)


src/frontend/apps/impress/src/components/quick-search/QuickSearch.tsx
```diff
@@ -74,7 +76,11 @@ export const QuickSearch = ({
             </QuickSearchInput>
           )}
           {beforeList}
-          <Command.List id={listId} aria-label={label} role="listbox">
+          <Command.List
+            id={listId}
+            aria-label={listLabel ?? label}
+            role="listbox"
+          >
             <Box>{children}</Box>
```
AntoLC: "cmdk" manage accessibility by default, overwriting props does not seems to work on this component, "Search results" is not in the dom: 
 
image

Ovgodd: You're right, I added it to keep a separate label for the listbox, but cmdk already handles that. I removed it

AntoLC (approved):

PR #2396, opened 2026-06-04
💄(frontend) display emoji button on hover

AntoLC:
## Purpose

The emoji button is now displayed when hovering the doc header. 
We reduced the size as well.

## Demo

https://github.com/user-attachments/assets/f91fd511-5467-4c6b-8d06-9395eedc8d41


Ovgodd (approved):
💯

PR #2404, opened 2026-06-05
👷(CI) remove test-e2e-other-browser job

AntoLC:
## Purpose

We don't maintain the `test-e2e-other-browser` job, and it often fails due to various reasons.
To avoid unnecessary noise in our CI pipeline, we will remove this job.

We are adding `workflow_dispatch` trigger to the `e2e-tests` workflow, so that we can still run e2e tests on other browsers when needed from the Github actions UI.

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -1,6 +1,23 @@
 name: E2E Tests
 
 on:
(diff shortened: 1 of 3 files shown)


AntoLC:
Let's stop this insanity...

lunika (approved):

PR #2408, opened 2026-06-08
🔧(backend) delete object created in object storage during tests

lunika:
## Purpose

The tests are using the local object storage configured in the docker compose stack. Every time a test using the DocumentFactory is executed, an object is created in the object storage and will stay here but never used. The data/media directory is growing and time to time we have to delete it in order to free some gigabytes. In this commit we add a global conftest responsible to delete the object created during the test execution. The space will be really free when the object storage container will be shutdown.

## Proposal

* [x] 🔧(backend) delete object created in object storage during tests

### DIFF
--- a/src/backend/conftest.py
+++ b/src/backend/conftest.py
@@ -1,0 +1,159 @@
+"""
+Root pytest configuration for the backend test suite.
+
+Automatically cleans up the objects created in the MinIO/S3 bucket during a test
+session, while preserving any object that already existed before the session
+started (e.g. data created by the local development environment, which shares
+the same ``data/media`` MinIO volume).
+
+Why this is needed
+------------------
+The test bucket (``impress-media-storage``) has versioning enabled, so deleting
+an object only adds a *delete marker* and keeps every previous version on disk.
+Tests also create objects through several code paths (``default_storage.save``,
+raw ``client.put_object`` / ``client.upload_fileobj``), some of which never
+clean up after themselves. As a result the shared ``data/media`` volume grows
+without bound, one test run after another.
+
+Strategy
+--------
+Rather than trying to track each individual write, we snapshot every existing
+object version *before* the session starts and, once the session finishes,
+permanently delete every version (and delete marker) that appeared during the
+session. This reclaims the disk space taken by test objects regardless of how
+they were created, and leaves pre-existing objects untouched.
+
+xdist safety
+------------
+The snapshot/cleanup runs only in the main process. Under ``pytest-xdist`` the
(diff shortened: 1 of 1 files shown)


src/backend/conftest.py
```diff
@@ -0,0 +1,160 @@
...
+                client.delete_objects(
+                    Bucket=bucket_name, Delete={"Objects": batch, "Quiet": True}
+                )
+                deleted += len(batch)
+                batch = []
+        if batch:
+            client.delete_objects(
+                Bucket=bucket_name, Delete={"Objects": batch, "Quiet": True}
+            )
+            deleted += len(batch)
```
kernicPanel: Maybe we could avoid the `client.delete_objects` duplication  here, and add some error logging ?

kernicPanel (approved):
Just an idea, other than that, LGTM.

src/backend/conftest.py
```diff
@@ -0,0 +1,160 @@
...
+        if client is None:
+            return
+
+        created = _list_all_versions(client, bucket_name) - _PRE_EXISTING_OBJECTS
+        if not created:
+            return
+
+        # delete_objects accepts at most 1000 keys per request.
+        deleted = 0
+        batch = []
```
kernicPanel: `batch` is not used anymore.
```suggestion
```

src/backend/conftest.py
```diff
@@ -0,0 +1,160 @@
...
+        for batch in itertools.batched(created, n=1000, strict=False):
```
kernicPanel, line 139: `strict=False` is the default.
```suggestion
        for batch in itertools.batched(created, n=1000):
```

src/backend/conftest.py
```diff
@@ -0,0 +1,160 @@
...
+            client.delete_objects(
+                Bucket=bucket_name,
+                Delete={
+                    "Objects": [
+                        {"Key": key, "VersionId": version_id}
+                        for key, version_id in batch
+                    ],
+                    "Quiet": True,
+                },
+            )
```
kernicPanel, line 149: IMO this would be more readable:

```suggestion
            objects_to_delete = [
                {"Key": key, "VersionId": version_id} for key, version_id in batch
            ]
            client.delete_objects(
                Bucket=bucket_name,
                Delete={"Objects": objects_to_delete, "Quiet": True},
            )
```

src/backend/conftest.py
lunika: Yes but ruff is not happy when `strict` is not used.

PR #2420, opened 2026-06-12
🐛(frontend) fix removed item in the tree

AntoLC:
## Purpose

`react-arborist's` scrollTo calls react-window's scrollToItem, which mutates the internal scrollOffset state. When navigating to a deep item in a large tree, this causes all items above the target to be removed from the DOM (virtualized away), making the tree appear empty above the selected node. 
We no-op it to prevent that — the panel's own overflow-y handles scrolling.

----

We tried to improve the performance of tree as well.

The tree rerenders crazily, creating performance issues, especially on big trees. 
To fix it properly we will need a big refacto, on the uikit side.
Before doing that, we try to optimize the current code by memoizing the tree view, and by using the selected node from the url instead of the one from the context.
We improve the treeitem link, we can now easily opening a new tab by doing CTRL+Click.

### DIFF
--- a/src/frontend/apps/impress/src/features/docs/doc-tree/components/DocSubPageItem.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-tree/components/DocSubPageItem.tsx
@@ -7,13 +7,12 @@ import {
   TreeViewNodeTypeEnum,
   useTreeContext,
 } from '@gouvfr-lasuite/ui-kit';
-import { useRouter } from 'next/navigation';
+import { useRouter } from 'next/router';
 import { useEffect, useRef, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { css } from 'styled-components';
 
 import { Box, Icon, StyledLink, Text } from '@/components';
-import { useCunninghamTheme } from '@/cunningham';
 import {
   Doc,
   DocIcon,
@@ -52,6 +51,7 @@ export const DocSubPageItem = (props: TreeViewNodeProps<Doc>) => {
 
 const DocSubPageLoadMore = (props: TreeViewNodeProps<Doc>) => {
   const treeContext = useTreeContext<Doc>();
+  const { t } = useTranslation();
   const loaderRef = useRef<HTMLDivElement>(null);
   const inFlightRef = useRef<boolean>(false);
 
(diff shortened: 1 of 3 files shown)


AntoLC:
@codeRabbit review

src/frontend/apps/impress/src/features/docs/doc-tree/components/DocSubPageItem.tsx
```diff
@@ -91,8 +91,10 @@ const DocSubPageLoadMore = (props: TreeViewNodeProps<Doc>) => {
       $align="center"
       $justify="center"
       $padding={{ vertical: 'xs' }}
+      role="status"
+      aria-label={t('Loading more documents')}
```
Ovgodd:
```suggestion
      role="status"
      >
      <span className="sr-only">{t('Loading more documents')}</span
```

little suggestion : `role="status"` announces text content changes inside the region, not the aria-label. 

A .sr-only child with the text would work properly, but it's minor here since the spinner is short-lived.

aria-label names an element and is read when the screen reader encounters it, it does not auto-announce on appear/update. For loading states, put the message as text inside a `role="status"` region (e.g. .sr-only), not only as aria-label

src/frontend/apps/impress/src/features/docs/doc-tree/components/DocSubPageItem.tsx
```diff
@@ -188,18 +176,38 @@ const DocSubPageItemContent = (props: TreeViewNodeProps<Doc>) => {
   };
 
   return (
-    <Box
+    <StyledLink
```
Ovgodd, line 179: we now have butons (`DocIcon `and `DocTreeItemActions`) inside an `<a>` tag, that break the `<a>` nested rules , is it intended ?

Ovgodd (changes requested):
nice ! just a suggestion about accessibility and one question regarding `buttons` in `<a>` (`StyledLink`)

src/frontend/apps/impress/src/features/docs/doc-tree/components/DocSubPageItem.tsx
AntoLC: How will you do ?
As it is in production, we have only the text that is inside a `a` tag, and a onClick on the `treeitem` (treeItem was in double btw), by doing so, we had the Ctrl+click working only partially because of that.
image

Ovgodd: hmm maybe with a stretched linj pattern : https://getbootstrap.com/docs/4.4/utilities/stretched-link/
the goal is to Make any HTML element or component clickable by “stretching” a nested link via CSS.

`<a>` wraps the text but use `::after {position: absolute, inset: 0}`
and Buttons with` position: relative; z-index:1 `( I didn't try )

But we can do this in another PR I think !

Ovgodd (approved):
👍

Issue #2326, opened 2026-05-26
Mention users in doc and comments


robin-lecomte:
<img width="2386" height="1072" alt="Image" src="https://github.com/user-attachments/assets/f1b7741c-390d-43f5-9896-c4675ccbaad7" />

Ability to mention users (1) in comments and (2) inline within the text.  
For inline mentions, I suggest reusing the same system as interlinking, combining both options when typing "@".

[FIGMA](https://www.figma.com/design/qdCWR4tTUr7vQSecEjCyqO/Docs?node-id=14669-35947&t=dUeSAcd8KtaMwSyd-1)

virgile-dev:
@robin-lecomte discussed with Antho, creating a thread from the side panel is quite complex. 
A couple question we need to solve with this  feature:
- I suggest we expose only members and link_trace users (later we will deal with mentioning users that don't have access which will trigger a sharing flow)
- The `@` operator is used for interlinking as well, we probably want user to mention a page in the comments, so we need to redesign the dropdown to expose 2 sections : users and subdocs
- I suggest, that at first we use an email notification when a user is mentionned

robin-lecomte:
<img width="2032" height="1440" alt="Image" src="https://github.com/user-attachments/assets/b9117d76-a4d5-41ce-94c5-582c2f47adf3" />

Updated the mockups accordingly. @virgile-dev, I’ll let you mark as design approved if that works for you.

[FIGMA](https://www.figma.com/design/qdCWR4tTUr7vQSecEjCyqO/Docs?node-id=14669-35947&t=KCp7J8y6FX5jYbwp-1)

PR #2447, opened 2026-06-20
✨(backend) add user mentions with email notifications

maboukerfa:
## Purpose

Add a mention feature: when a user is mentioned in a document body or a comment thread, they receive an email notification linking straight to the anchor of the mention.

Related issue : https://github.com/suitenumerique/docs/issues/2326

## Proposal

- Make the document access list visible to all collaborators
- Add a `POST /api/v1.0/documents/{id}/mention/` endpoint that records the mention and sends an email notification to the mentioned user.
- Send the mention notification email asynchronously via a Celery task to keep the request fast
- Throttle the mention endpoint to protect against spam / abuse

## Notes

* Email notifications are suppressed when the same user was already notified in
  the same context (document body or thread) within the cooldown window
  (`MENTION_NOTIFICATION_COOLDOWN_MINUTES`, default 15).

### Email content

<table>
  <tr>
    <td align="center">
      <strong>Subject :</strong>You were mentioned in the document mentions test<br>
      Screenshot 2026-06-20 at 21 06 56
    </td>
    <td align="center">
      <strong>Subject :</strong>You were mentioned in a comment on the document mentions test<br>
      Screenshot 2026-06-20 at 21 08 35
    </td>

  </tr>
</table>


maboukerfa:
@lunika i made a tradeoff in https://github.com/suitenumerique/docs/pull/2447/changes/9b74a2fe8953b3c50c211038ce62d4cd81239194 by sending the email notification async in a celery job, but it introduces a race condition, if a user spams the `/mention` endpoint we can send multiple emails  to the target. What do you think ?

src/backend/core/api/serializers.py
```diff
@@ -1005,6 +1005,67 @@ def get_abilities(self, thread):
...
+            "id",
+            "document_id",
+            "mentioned_by_user_id",
+            "created_at",
+            "notified_at",
+        ]
+
+    def validate_mentioned_user_id(self, user):
+        """Ensure the mentioned user has access to the document."""
+        document = models.Document.objects.get(pk=self.context["document"].pk)
```
lunika:
```suggestion
        document = models.Document.objects.annotate_user_roles(user).get(pk=self.context["document"].pk)
```

lunika:
> @lunika i made a tradeoff in [9b74a2f](https://github.com/suitenumerique/docs/commit/9b74a2fe8953b3c50c211038ce62d4cd81239194) by sending the email notification async in a celery job, but it introduces a race condition, if a user spams the `/mention` endpoint we can send multiple emails to the target. What do you think ?

Maybe you can use the django cache framework to fix the race condition. We are using Redis as a backend so the cache is shared between all Django services, including Celery.

Using [cache.add](https://docs.djangoproject.com/fr/6.0/topics/cache/#django.core.cache.cache.add) is probably the solution by testing its return value and you "compute" the cache key using all the elements you need. The first celery task calling `cache.add` will have the return value `True` and then the other `False`

```
guard_key = (
            "mention-notify:"
            f"{self.document_id}:{self.mentioned_user_id}:{self.thread_id or ''}"
        )
        if not cache.add(
            guard_key,
            str(self.pk),
            timeout=settings.MENTION_NOTIFICATION_COOLDOWN_MINUTES * 60,
        ):
            return False or raise as you want

    Then you can continue the notify method.
```

src/backend/core/migrations/0033_mention.py
```diff
@@ -0,0 +1,107 @@
...
+import uuid
+
+import django.db.models.deletion
+from django.conf import settings
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("core", "0032_remove_linktrace_is_masked"),
```
lunika, line 12: Migration 0033 already exists now; you have to increment this one.

src/backend/core/models.py
```diff
@@ -2029,6 +2031,154 @@ def __str__(self):
...
+        return False otherwise.
+        """
+        user = self.mentioned_user
+        if user is None or not user.email or self.is_notification_in_cooldown():
+            return False
+
+        # Guard against concurrent notification tasks for the same context.
+        # cache.add is atomic on the shared Redis backend: only the first task
+        # acquires the key and proceeds, the others get False
+        if not cache.add(
```
lunika, line 2139: The key cache is added, but if something fails later and no mention is sent, the entry cache remains in place. You should delete it in that case.

src/backend/core/models.py
```diff
@@ -2029,6 +2031,154 @@ def __str__(self):
...
+            context = {
+                "title": subject,
+                "message": message,
+                "link": f"{domain}/docs/{self.document_id}/#{self.anchor_id}",
+            }
+
+        self.document.send_email(subject, [user.email], context, language)
+
+        self.notified_at = timezone.now()
+        self.save(update_fields=["notified_at", "updated_at"])
```
lunika, line 2178: If `self.document.send_email(subject, [user.email], context, language)` fails, then the `notified_at` is updated.

src/backend/core/api/viewsets.py
```diff
@@ -2719,11 +2759,11 @@ def list(self, request, *args, **kwargs):
             | models.Document.objects.filter(pk=self.document.pk)
         ).filter(ancestors_deleted_at__isnull=True)
 
+        # All users with access see the full list of accesses (with limited
+        # user details for unprivileged roles) so that any collaborator
+        # allowed to comment can mention the others.
```
lunika, line 2765: `reader` cannot comment; they should not have access to the full list.

src/backend/core/api/serializers.py
```diff
@@ -1005,6 +1005,69 @@ def get_abilities(self, thread):
...
+            "created_at",
+            "notified_at",
+        ]
+
+    def validate_mentioned_user_id(self, user):
+        """Ensure the mentioned user has access to the document."""
+        document = models.Document.objects.annotate_user_roles(user).get(
+            pk=self.context["document"].pk
+        )
+        if document.get_role(user) is None:
```
lunika, line 1056: A `reader` can be mentioned but will not be able to access the `comment` or the content in the document. Mentioning a `reader` should not be allowed.

src/backend/core/models.py
```diff
@@ -1398,6 +1398,7 @@ def get_abilities(self, user):  # pylint: disable=too-many-locals
             "link_configuration": is_owner_or_admin,
             "invite_owner": is_owner and not is_deleted,
             "leave": can_leave,
+            "mention": can_comment and user.is_authenticated,
```
lunika, line 1401: Having an explicit access should be check otherwise user accessing a document without explicit will have nobody to mention, the `access_view` is only available for user with access role.

src/backend/core/models.py
```diff
@@ -2029,6 +2031,154 @@ def __str__(self):
...
+    sent if no notification was already sent to the same user in the same
+    context (the document or a specific thread) within the cooldown period.
+    """
+
+    document = models.ForeignKey(
+        Document,
+        on_delete=models.CASCADE,
+        related_name="mentions",
+    )
+    anchor_id = models.TextField()
```
lunika, line 2047: this value should be validated. It will be injected in the email without any validation.

maboukerfa: I wasn't sure of the format of the anchor at that time, but with the jump to block feature i think we can restrict it to be a uuid right?

src/backend/core/api/serializers.py
maboukerfa: Alright! we can think of a pop-up to grant the mentioned user (if he's a `reader`) additional privileges in next iterations

src/backend/core/api/viewsets.py
maboukerfa: what about a `Commenter` ?

src/backend/core/models.py
maboukerfa: Alright, since it had a TTL of 15 minutes i didn't bother, but ok i'll try to catch the errors and delete the entry

src/backend/core/api/serializers.py
maboukerfa: thinking about it sounds a bit off, so we have to explicitly promote a user to commenter before mentioning him ?

src/backend/core/api/viewsets.py
lunika: Yes a commenter can.

src/backend/core/models.py
lunika: Yes

src/backend/core/api/serializers.py
lunika: Bumping a user role is, I think, outside the scope of this PR, and a UX designer should be involved.

virgile-dev:
Hey @maboukerfa 
Cool PR thanks!
I took the liberty to rework the copy of the email. Here goes.

### **Email 1: Mentioned in a Document**
**Subject:**  {{User_name}} mentioned you in "{{document_title}}" 

**Title:**
 {{User_name}} mentioned you in the following document: 
[{{document_title}}]({{document_link#mention_anchor}}) 

**Button:** [{{Open}}]({{document_link#mention_anchor}}) 

*Docs: Your essential tool for organizing, sharing, and collaborating on documents as a team.*

*A solution by La Suite Numérique.*

### **Email 2: Mentioned in a Comment**
**Subject:**  {{User_name}} mentioned in a comment in {{document_title}} 

**Title:**
 {{User_name}} mentioned you in a comment in the following document
[{{document_title}}]({{document_link#comment_anchor}}) 

**Button:** [{{Open}}]({{document_link#comment_anchor}}) 

*Docs: Your essential tool for organizing, sharing, and collaborating on documents as a team.*

*A solution by La Suite Numérique.*

src/backend/core/api/serializers.py
maboukerfa: alright, so changing it to be only commenter can see other commenters/privileged users. and readers can't be mentioned

src/backend/core/models.py
maboukerfa: email fail silently, do you have any suggestions to catch the fail?

lunika (approved):

PR #2452, opened 2026-06-22
♿️(frontend) limit share modal opening announcement for screen readers

### LINKED ISSUE #2334: Share modal: spurious restitution of accessible names unrelated to the dialog window
**Observed behavior**

When opening the share modal (editor mode), after its accessible name is announced, the screen reader vocalizes information unrelated to the window's content: "Search results", "Suggestions", "Document visibility card", "Link copied OK". These strings come from `aria-label` attributes or `<label>` elements present in the code — likely search, suggestion, and notification components that remain present while the share modal is open.

**Expected behavior**

When a modal is open, only the dialog window's content should be announced. Elements not related to this modal must be hidden from assistive technologies.

Only content relevant to the current task (sharing the document) should be perceivable.

**Steps to reproduce**

1. Open a document in editor mode.
2. Open the share modal.
3. Navigate with a screen reader when the modal opens.
4. Observe that the accessible names "Search results", "Suggestions", "Document visibility card", and "Link copied OK" are vocalized.

**RGAA criteria**

[Critère 7.1 : Chaque script est-il, si nécessaire, compatible avec les technologies d'assistance ?](https://accessibilite.numerique.gouv.fr/methode/criteres-et-tests/#7.1)

**Impact**

A blind user navigating with a screen reader opens the share modal. Before even reaching the window's controls, they hear spurious information belonging to other page components. This inconsistent announcement misinforms the user about the modal's actual content and disrupts their navigation.

**Priority**

P1

Ovgodd:
## Purpose

Limit the screen reader announcement when opening the share modal without the search input.

## Proposal

* [x] Autofocus the close button when the search input is hidden

### DIFF
--- a/src/frontend/apps/impress/src/features/docs/doc-share/components/DocShareModal.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-share/components/DocShareModal.tsx
@@ -40,7 +40,8 @@ import {
 import { QuickSearchGroupMember } from './DocShareMember';
 import { DocShareModalFooter } from './DocShareModalFooter';
 
-const DEBOUNCE_MS = 300;
+const SEARCH_QUERY_DEBOUNCE_MS = 300;
+const ACCESSIBILITY_REEXPOSURE_DELAY_MS = 300;
 
 const ShareModalStyle = createGlobalStyle`
   .--docs--doc-share-modal [cmdk-item] {
@@ -122,7 +123,7 @@ export const DocShareModal = ({ doc, onClose, isRootDoc = true }: Props) => {
 
   const onFilter = useDebouncedCallback((str: string) => {
     setUserQuery(str);
-  }, DEBOUNCE_MS);
+  }, SEARCH_QUERY_DEBOUNCE_MS);
 
   const onRemoveUser = (row: User) => {
     setSelectedUsers((prevState) => {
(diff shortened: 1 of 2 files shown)


src/frontend/apps/impress/src/features/docs/doc-share/components/DocShareModal.tsx
```diff
@@ -206,6 +228,7 @@ export const DocShareModal = ({ doc, onClose, isRootDoc = true }: Props) => {
       >
         <ShareModalStyle />
         <Box
+          aria-hidden={isContentAccessible ? undefined : true}
```
AntoLC, line 234: I am not sure to understand why do we do that ? Seems strange to hide a part of the modal a few milliseconds, without apparent reason.

Ovgodd: Yes good point, I should have added a comment, just pushed one!

I first tried with just `autoFocus` on the close button but it was not working, NVDA was still reading all the modal body on open (Suggestions, link settings, Copy link…). After several tests I found that this temporary `aria-hidden` (300ms) is the only way to block this initial reading. After the timeout it's removed and everything works normally with Tab navigation.

Ovgodd:
waiting extra check from @cyberbaloo

Ovgodd:
https://github.com/suitenumerique/docs/issues/2334#issuecomment-5454639447

PR #2454, opened 2026-06-23
🐛(frontend) fix service worker causing reload on tab focus

AntoLC:
## Purpose

During rolling deploys, multiple replicas serve different builds of service-worker.js (each build has a random buildId baked in).
This causes a controllerchange loop: the new SW activates via skipWaiting, the page reloads, then the next request may hit an older replica,triggering another update and another reload — surfacing as a reload every time the tab becomes visible.

We added a sessionStorage guard so a controllerchange triggers at most one automatic reload per session; subsequent flaps between versions are silently ignored, while legitimate updates still cause a single reload automatically.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to
 ### Fixed
 
 - 🐛(frontend) fix removed item in the tree #2420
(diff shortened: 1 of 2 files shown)


Ovgodd (approved):
lgtm !

indeed hard to reproduce

Issue #2364, opened 2026-05-28
[EPIC] UI V3 Experiment


robin-lecomte:
<img width="2044" height="638" alt="Image" src="https://github.com/user-attachments/assets/959ba44a-af07-4ba1-b0fe-b0ae0418d275" />

Proposal of a new design for the UI to be mobile-first and aligned with LaSuite’s new visual standards.

@virgile-dev and @AntoLC hope to discuss this with you asap :)

[FIGMA](https://www.figma.com/design/qdCWR4tTUr7vQSecEjCyqO/Docs?node-id=14791-49375&t=reINZ6FFpvVqng9Q-1)

AntoLC:
- [x] #2423 
- [x] #2455 
- [x] https://github.com/suitenumerique/docs/pull/2463
- [x] #2471
- [x] https://github.com/suitenumerique/docs/pull/2534
- [x] #2539
- [x] https://github.com/suitenumerique/docs/issues/2417

PR #2455, opened 2026-06-23
🚚(frontend) move Waffle to bottom left

AntoLC:
## Purpose

We moved the waffle button to the bottom left inside the left panel.
This change is part of the new header design, see #2364

## Demo

image


src/frontend/apps/impress/src/features/home/components/HomeHeader.tsx
Ovgodd, line 87: The waffle was moved to the bottom-left in LeftPanelFooter, but it's still in the homepage header. Is that intended? 

```suggestion
//should we remove it ? 
{!isSmallMobile && (
<Box $direction="row" $gap="1rem" $align="center">
<LanguagePicker />
<Waffle />
</Box>
)}
</Box>
```

Ovgodd (changes requested):
nice ! just few things

and : Waffle is missing in the mobile panel is it intentional? 

Image

src/frontend/apps/impress/src/features/home/components/HomeHeader.tsx
AntoLC: Here it is as intended, we don't have a LeftPanel in the Home feature.

Ovgodd: indeed !

Ovgodd (approved):
👍

PR #2480, opened 2026-06-30
🛂(backend) remove unused default authentication backend

lunika:
## Purpose

The authentication backend
mozilla_django_oidc.contrib.drf.OIDCAuthentication
is present in the default authentication classes for the REST_FRAMEWORK
settings. This backend should not be used by our application and can
lead to the usage of our main api with an access_token instead of the
cookie session.
We need to override the drf SessionAuthentication backend to implement
the authenticate_header method. Without this, a 403 status code is
returned, but it is not valid. It must a be 401

## Proposal

* 🛂(backend) remove unused default authentication backend

### DIFF
--- a/src/backend/core/authentication/backends.py
+++ b/src/backend/core/authentication/backends.py
@@ -10,11 +10,15 @@ from lasuite.marketing.tasks import create_or_update_contact
 from lasuite.oidc_login.backends import (
     OIDCAuthenticationBackend as LaSuiteOIDCAuthenticationBackend,
 )
+from rest_framework.authentication import (
+    SessionAuthentication as BaseSessionAuthentication,
+)
 
 from core.models import DuplicateEmailError
 from core.utils.analytics import PosthogEventName, posthog_capture
 
 logger = logging.getLogger(__name__)
+
 
 # Settings renamed warnings
 if os.environ.get("USER_OIDC_FIELDS_TO_FULLNAME"):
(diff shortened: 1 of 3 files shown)


qbey (approved):
Could be nice to have it in django-lasuite, but you can merge already, we'll move it later.

src/backend/core/authentication/backends.py
```diff
@@ -81,3 +85,19 @@ def post_get_or_create_user(self, user, claims, is_new_user):
...
+    DRF chooses between 401 and 403 based on whether the first authentication class returns a
+    truthy `authenticate_header`. The stock `SessionAuthentication` does not implement it
+    (it inherits `BaseAuthentication.authenticate_header`, which returns `None`), so anonymous
+    requests get a misleading 403. Returning a value restores the correct 401, signalling that the
+    client may authenticate and retry.
+    """
+
+    def authenticate_header(self, request):
+        """Return a non-empty challenge so DRF responds with 401 rather than 403."""
+        return "Session"
```
rsebille, line 103: Just here by curiosity so feel free to ignore me :), but `Session` doesn't seems to be a valid challenge for [`WWW-Authenticate`](https://httpwg.org/specs/rfc9110.html#field.www-authenticate) : https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml#authschemes
From my understanding we could have a custom one but `WWW-Authenticate` header is here to guide the content of the `Authorization` header, which is not how session works ^^.
So from a pure HTTP point of view, the 403 make more senses (_kinda_) as the client doesn't have any means (nor clues) on how to authenticate : "I need a session OK, but how do I get one?" ;), but I agree that 403 is not right either as no credentials were sent, _probably_ 🤷.

lunika: I fully agree with you, but having a 403 is not acceptable for us.

Issue #2466, opened 2026-06-25
[EPIC] Improve present mode


virgile-dev:
This feature is getting a lot of positive feedback. 
I gathered the suggestions to improve it in the sub-issues bellow.

PR #2484, opened 2026-07-01
✨(frontend) Open and share presenter slide links

### LINKED ISSUE #2397: Link for presentation mode
## Feature Request

Docs recently introduced a presentation mode that lets users display a document as slides in a simple and lightweight way.

However, it is currently not possible to share a link that opens a document directly in presentation mode. This creates friction when users want to share a document specifically as a presentation, for example for a meeting, workshop, onboarding flow, or public resource.

I would like to be able to share a deep link that opens a Docs document directly in presentation mode.

A simple solution could be to support a URL parameter, for example:

`?mode=presentation`

or:

`?view=present`

When this parameter is present, the document would open directly in presentation mode instead of the standard editing or reading view.

The current alternative is to share the normal document link and ask users to manually activate presentation mode after opening the document.

This works, but it adds an extra step and makes the feature harder to use in contexts where the presenter wants to share a direct, ready-to-open presentation link.

This could be documented as part of the presentation mode documentation.

Example documentation:

> To share a document directly in presentation mode, add the presentation parameter to the document URL.
>
> Example:
> `https://docs.example/document-id?mode=presentation`
>
> Anyone with access to the document will be able to open it directly in presentation mode.

This would be useful for templates, public guides, training materials, meeting decks, and any document designed to be consumed as slides.

It could also be exposed through the sharing UI, for example with an option such as:

> Copy presentation link

Not at this stage, but I would be happy to help clarify the expected behavior or test the feature if needed.

PanchoutNathan:
## Purpose

- Support presenter deep-links with ?view=present&slide=N.
- Keep the URL in sync with the current presenter slide.
- Add a "Copy link to slide" action in the presenter controls.
- Ignore presenter deep-links on mobile viewports.

Refs #2466
Closes #2397

### DIFF
--- a/src/frontend/apps/impress/src/features/docs/doc-presenter/components/PresenterFloatingBar.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-presenter/components/PresenterFloatingBar.tsx
@@ -1,14 +1,17 @@
 import { Button } from '@gouvfr-lasuite/cunningham-react';
+import { DropdownMenu, DropdownMenuItem } from '@gouvfr-lasuite/ui-kit';
 import {
   ChevronLeft,
   ChevronRight,
+  Link,
   Maximize,
   Minimize,
+  Share,
   XMark,
 } from '@gouvfr-lasuite/ui-kit/icons';
-import { useEffect, useRef } from 'react';
+import { MouseEvent, useEffect, useMemo, useRef, useState } from 'react';
 import { useTranslation } from 'react-i18next';
-import { css } from 'styled-components';
+import { createGlobalStyle, css } from 'styled-components';
 
 import { Box, Text } from '@/components';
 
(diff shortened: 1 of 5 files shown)


src/frontend/apps/impress/src/features/docs/doc-presenter/components/PresenterFloatingBar.tsx
```diff
@@ -18,13 +21,15 @@ interface PresenterFloatingBarProps {
   isFullscreen: boolean;
   onPrev: () => void;
   onNext: () => void;
+  onCopyLink: () => void;
   onToggleFullscreen: () => void;
   onClose: () => void;
 }
 
 const barCss = css`
+  height: 32px;
```
AntoLC: I don't think you should force a height, or it should be min height, the elements inside should naturally lead the height.

src/frontend/apps/impress/src/features/docs/doc-presenter/components/PresenterOverlay.tsx
```diff
@@ -149,7 +173,10 @@ export const PresenterOverlay = ({
   }
 
   return createPortal(
-    <FocusScope contain autoFocus restoreFocus>
+    // The ui-kit DropdownMenu renders its popover in a React Aria portal.
+    // Keeping the FocusScope contained here prevents that portalled menu from
+    // receiving focus, which makes the dropdown close as soon as it opens.
+    <FocusScope autoFocus restoreFocus>
```
AntoLC: Not sure the comment create real value for developers, it seems more like a explanation about why the prop `contain` was removed.

AntoLC (approved):
2 nits, but LGTM, cool feat.

PR #2488, opened 2026-07-01
🔧(helm) mount custom cacert in yprovider deployment

lunika:
## Purpose

In the yprovider deployment we now need to mount the cacert containing the certificates generated by mkcert. Then, we have to declare the environment variable NODE_EXTRA_CA_CERTS with the path where this cacert is mounted. With this new config we don't have tls issue anymore.

## Proposal

* [x] 🔧(helm) mount custom cacert in yprovider deployment


AntoLC (approved):
Is it not necessary for the `feature` config (preview) ?
https://github.com/suitenumerique/docs/blob/main/src/helm/env.d/feature/values.impress.yaml.gotmpl

lunika:
No, it is not; we are using a valid certificate, not self signed.

PR #2490, opened 2026-07-03
🐛(frontend) stop force index redirect when delete doc

### LINKED ISSUE #2465: Redirect to home when deleting from "My Docs"
## Bug Report

**Problematic behavior**
After deleting a document from the "My Docs" page, the user is redirected to the homepage instead of staying on the "My Docs" page. This disrupts the workflow and forces the user to navigate back to "My Docs" manually.

**Expected behavior/code**
After deleting a document, the user should remain on the "My Docs" page. The page should refresh or update to reflect the deletion (e.g., the deleted document should disappear from the list), but the user should not be redirected away from "My Docs."

**Steps to Reproduce**
1. Navigate to the "My Docs" page.
2. Select a document to delete.
3. Confirm the deletion.
4. Observe that the user is redirected to the homepage instead of staying on "My Docs."

**Environment**
- Docs version: 5.3
- Instance URL: docs.numerique

AntoLC:
## Purpose

Depends the state, after deleted a doc, we were automatically redirecting to the index page, which is not always the desired behavior. This change stops the automatic redirect after deleting a document, allowing for more flexible navigation based on user context.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to
 - 🐛(frontend) fix removed item in the tree #2420
 - 🐛(frontend) fix service worker causing reload on tab focus #2454
 - 🐛(backend) update restore ability for inherited deletion #2148
+- 🐛(frontend) stop force index redirect when delete doc #2490
 
 ### Removed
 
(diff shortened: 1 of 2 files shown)


Ovgodd (approved):
👍

PR #2501, opened 2026-07-07
♻️(backend) scope document search by document id instead of path

lunika:
## Purpose

The search in a document tree was triggered by the usage of the document
path. The path is something guessable by incrementing it you can
discover public documents. We decided to change this to use the document
id which is not guessable and prevent discovering public documents.
Thanks to @maboukerfa for discovering it.

## Proposal

* [x] ♻️(backend) scope document search by document id instead of path
* [x] ♻️(frontend) change search using document path to its id

### DIFF
--- a/src/backend/core/api/viewsets.py
+++ b/src/backend/core/api/viewsets.py
@@ -1542,15 +1542,23 @@ class DocumentViewSet(
         """
         queryset = models.Document.objects.all()
 
+        # The indexer filters descendants by path prefix, so resolve the document
+        # id to its path before querying it.
+        path = None
+        document_id = params.validated_data.get("document")
+        if document_id:
+            try:
+                path = models.Document.objects.get(pk=document_id).values_list(
+                    "path", flat=True
+                )
+            except models.Document.DoesNotExist as exc:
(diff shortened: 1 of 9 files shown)


AntoLC (approved):
Frontend side seems good ✅

src/backend/core/api/viewsets.py
```diff
@@ -1542,15 +1542,23 @@ def _search_using_indexer(indexer, request, params, search_type):
...
 
+        # The indexer filters descendants by path prefix, so resolve the document
+        # id to its path before querying it.
+        path = None
+        document_id = params.validated_data.get("document")
+        if document_id:
+            try:
+                path = models.Document.objects.values_list("path", flat=True).get(
+                    pk=document_id
+                )
```
jmaupetit: ⚠️ Nitpick: I prefer focusing on the filtering part before fields selection.

```suggestion
                path = models.Document.objects.get(pk=document_id).values_list(
                    "path", 
                    flat=True,
                )
```

But if you think SQL you are on the right side 😉

jmaupetit (approved):
✅ The backend part looks good to me.

PR #2508, opened 2026-07-08
📝(docs) update readme

virgile-dev:
Reflect new features and update screenshots


src/mail/yarn.lock
AntoLC, line 1: Not sure why this file was updated, but it should not.

AntoLC (commented):
Dead link:
Image

---

Could we add a small AI section with a gif ? The ux of the Blocknote AI is quite incredible, it worth to be shown I think. It could bring improvement on the Blocknote side as well on this feature.

AntoLC (approved):

PR #2515, opened 2026-07-09
📝(doc) add email.eu to the list of public instances

maartendra:
Adds [email.eu](https://email.eu/) to the public instances page.

Email.eu is a sovereign EU business workspace that runs La Suite Docs. This mirrors the [existing entry in La Suite Meet](https://github.com/suitenumerique/meet/blob/main/README.md#known-instances).

Thanks for the great work! 🙏

### DIFF
--- a/documentation/instances.md
+++ b/documentation/instances.md
@@ -43,9 +43,9 @@ If you run a public instance and would like it listed here, feel free to open a 
 ### email.eu
 
 **Organization:** Email.eu  
-**Type:** Sovereign business workspace  
-**Access:** Account required  
-<https://email.eu/>
+**Type:** Demo instance  
+**Access:** Open, no account needed  
+<https://email-eu.wiki.email.eu/docs/f23cd0ac-87d6-4b8e-8b55-5a15c814a4b6/>
 
 ### notes.liiib.re
 
--- a/documentation/instances.md
+++ b/documentation/instances.md
@@ -40,6 +40,13 @@ If you run a public instance and would like it listed here, feel free to open a 
 **Type:** Demo instance  
 <https://docs.demo.mosacloud.eu/>
 
+### email.eu
+
+**Organization:** Email.eu  
+**Type:** Sovereign business workspace  
+**Access:** Account required  
+<https://email.eu/>
+
 ### notes.liiib.re
 
 **Organization:** lasuite.coop  


documentation/instances.md
```diff
@@ -40,6 +40,13 @@ If you run a public instance and would like it listed here, feel free to open a
 **Type:** Demo instance  
 <https://docs.demo.mosacloud.eu/>
 
+### email.eu
+
+**Organization:** Email.eu  
+**Type:** Sovereign business workspace  
+**Access:** Account required  
+<https://email.eu/>
```
lunika: I see nothing related to docs. Can you share the product page ?

Thanks

maartendra: Hi @lunika, thanks for taking a look!

We ship Docs under the name **Wiki** — it's our collaborative knowledge-base pillar, running on `suitenumerique/docs`. That's why the homepage doesn't surface "docs" directly.

Here's the product page: https://email.eu/product/wiki

The instance itself lives at `wiki.email.eu`, but it's behind SSO (account required), so a link there just lands on a login screen.

virgile-dev: Hello @maartendra we've put the actual url where the product is hosted.
So the proper link is  wiki.email.eu. Can you update ?
Nice to see a new reuser !

maartendra: Thanks for the pointer! I looked into using `wiki.email.eu`, but that host doesn't actually resolve — our Docs deployment is multi-tenant: each customer gets their own isolated instance at `<tenant>.wiki.email.eu` (SSO-gated, one per organisation), so there's no single shared `wiki.email.eu` apex to point at.

That's why the entry points at https://email.eu/ as the public landing. If you'd rather link a page specifically about the Docs product, https://email.eu/product/wiki describes it. Happy to use whichever fits the list best!

virgile-dev: Unfortunately this an instance list not a vendors list. 
One way so solve this is to point to a demo instance like mosa Cloud and lasuite.coop
Wdyt?

maartendra: Sorted! I've set up an open, public demo page (no account, no login) and pointed the entry there:

https://email-eu.wiki.email.eu/docs/f23cd0ac-87d6-4b8e-8b55-5a15c814a4b6/

Anyone can open it and edit live in real time. Thanks for the nudge to do this properly, and for the warm welcome to the ecosystem!

PR #2519, opened 2026-07-10
✨(frontend) add dark / light / system colour-mode switcher

MashdorDev:
## Purpose

Docs already ships a `dark` Cunningham token set, but there is no way for a user to select it and no following of the OS colour scheme. This adds an in-app dark / light / system colour-mode switcher, which is one of the recurring asks around theming.

## Proposal

* [x] Add a `themeMode` (`light` / `dark` / `system`) to the theme store, persisted to `localStorage`
* [x] In `system` mode, resolve and follow the OS `prefers-color-scheme` live
* [x] Pre-hydration inline script in `_document` sets the theme class on `<html>` so the correct scheme paints on first frame (no flash), with `suppressHydrationWarning`
* [x] A user's explicit mode preference takes precedence over the plain light/dark `FRONTEND_THEME` backend default (non-default brand themes like `dsfr` still apply)
* [x] A `ThemeSwitch` control in the left-panel footer (app) and the home header (landing page)
* [x] Unit tests for the theme-mode store logic

The store maps the mode onto the existing Cunningham themes (`light` -> `default`, `dark` -> `dark`), so no new tokens are introduced.

## External contributions

### General requirements

### CI requirements

* [ ] I have signed my commits with my SSH or GPG key (`git commit -S`)

### AI requirements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to
 
 - ♿️(frontend) restore skip to content link after header redesign #2510
 - 🌐(i18n) rename cn_CN to zh_CN, add eo_PL and zh_TW locales #2486
+- ✨(frontend) add dark / light / system colour-mode switcher #2519
 
 ### Fixed
 
--- a/src/frontend/apps/impress/src/components/FloatingBar.tsx
+++ b/src/frontend/apps/impress/src/components/FloatingBar.tsx
@@ -19,7 +19,11 @@ const FLOATING_STYLES = css`
     position: absolute;
     inset: 0;
     z-index: -1;
-    background: linear-gradient(180deg, #fff 0%, rgba(255, 255, 255, 0) 100%);
+    background: linear-gradient(
+      180deg,
+      var(--c--contextuals--background--surface--primary) 0%,
+      transparent 100%
+    );
     backdrop-filter: blur(1px);
     -webkit-backdrop-filter: blur(1px);
     mask-image: linear-gradient(180deg, black 50%, transparent 100%);
--- a/src/frontend/apps/impress/src/components/ThemeSwitch.tsx
+++ b/src/frontend/apps/impress/src/components/ThemeSwitch.tsx
@@ -1,0 +1,65 @@
+import { useTranslation } from 'react-i18next';
+import { css } from 'styled-components';
+
+import { ThemeMode, useCunninghamTheme } from '@/cunningham';
+
+import { Box } from './Box';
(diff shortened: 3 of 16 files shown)


lucletoffe:
In case it helps unblock this: the branch no longer merges cleanly onto `main`, so I rebased the four commits onto `61c2183` with @MashdorDev's authorship preserved, and pushed the result here:

https://github.com/lucletoffe/docs/tree/darkmode-rebase-main

The only conflict was `CHANGELOG.md` — the `### Added` section had moved on since this PR was opened. Resolved by keeping both sets of entries. No code changes were needed on top of the original work. `yarn vitest run` is green on the rebased branch (39 files / 264 tests).

One data point that might be useful: we are running a build of these commits on `v5.4.1` (which is exactly where this branch is based — `git merge-base` returns the `v5.4.1` tag) on a self-hosted instance. Verified by headless render of an impress-served route in both modes: `html.cunningham-theme--default` with a `#FFFFFF` dominant, `html.cunningham-theme--dark` with a `#2F3033` dominant. The switcher, the `localStorage` persistence and the pre-hydration inline script all behave as described.

Two cosmetic things we noticed, neither introduced by this PR:
- the brand ramp differs between the two themes (indigo in light, blue in dark), because the `dark` theme in `cunningham-tokens.ts` carries its own brand colours;
- the DINUM illustrations (401 page and friends) keep a white background in dark mode, since the assets are not transparent.

Feel free to take the rebase or ignore it — just wanted the conflict not to be the thing keeping this stalled.

MashdorDev:
Thanks Luc. The branch is rebased onto `61c2183` and force-pushed, so it merges cleanly again. I resolved `CHANGELOG.md` the same way you did. I did the rebase independently first and diffed it against your branch: the trees came out identical.

Your check also caught something I had missed, so there is a fifth commit now. Nothing painted a background on `html`, `body` or the app wrapper, so the browser default white showed through while the text tokens switched to the dark ramp, leaving the home page heading near-white on white. `body` now takes `--c--contextuals--background--surface--primary`, which is `gray-000` in `default` and `gray-800` in `dark`. That gives the `#2F3033` you measured.

Your other two points match what I see and neither comes from this PR: the brand ramp differs because the `dark` theme carries its own brand colours in `cunningham-tokens.ts`, and the DINUM illustrations keep a white plate because the assets are not transparent. The second one is more visible now that the canvas is actually dark, but fixing it needs new assets.

@virgile-dev the branch is current and CI is green if you want to take a look.

PR #2522, opened 2026-07-13
✨(frontend) warn the user when trying to upload a file size that exceeds the limit

### LINKED ISSUE #2460: 🐛 Big file upload fails as unknown error with poor UX feedback
## Bug Report

**Problematic behavior**
When trying to insert a [big video](https://fichiers.numerique.gouv.fr/explorer/items/files/ceefa412-f8e2-4412-a0c3-75da76633369) file (>10MB) on docs.numerique.gouv.fr, an unknown error is show to the user.

Image

**Expected behavior/code**

* Something less cryptic should happen

**Steps to Reproduce**
1. Download the linked file,
2. Create a new doc on docs.numerique.gouv.fr
3. Drag and drop the file in the docs
4. 💥 

**Environment**
- Docs version: Today's version :)

**Possible Solution**

* Before the upload a check on the file size vs authorized file size should be performed, and the user notified for a nice UX (a link to the doc for embedding the video directly would be great).

Nginx rejects the payload with a 413 that is not handled in the frontend currently, I do think it would be nicier to check the file size before upload.

magopian:
Fixes https://github.com/suitenumerique/docs/issues/2460

## Purpose

Currently the message displayed to the user conveys no information on the error that occured when uploading a big file which size exceeds the limit.

## Proposal

* [x] expose the `DOCUMENT_IMAGE_MAX_SIZE` configuration through the `/api/v1.0/config` endpoint
* [x] check the size of the file before trying to upload it
* [x] optional: also display a nice error message to the user if the production nginx server answers with a `HTTP 413` message

## External contributions

### General requirements

*Skip the checkbox below 👇 if you're fixing an issue or adding documentation*
* [ ] Before submitting a PR for a new feature I made sure to contact the product manager

### CI requirements

### AI requirements

*Skip the checkboxes below 👇 If you didn't use AI for your contribution*


src/frontend/apps/impress/src/features/docs/doc-editor/components/BlockNoteEditor.tsx
```diff
@@ -263,6 +271,41 @@ export const BlockNoteEditor = ({ doc, provider }: BlockNoteEditorProps) => {
...
+      }
+    };
+
+    container.addEventListener('drop', handleDrop, true);
+    container.addEventListener('paste', handlePaste, true);
+    return () => {
+      container.removeEventListener('drop', handleDrop, true);
+      container.removeEventListener('paste', handlePaste, true);
+    };
+  }, [checkFileSize]);
```
AntoLC: Seems unnecessary to listen these events to display the error message about the upload size.
In my understanding it is to not see the "loading" part, but we should act when we get the error, not before, so either by updating or removing the block, but only when the error occurs.

src/frontend/apps/impress/src/features/docs/doc-editor/hook/useUploadFile.tsx
```diff
@@ -19,6 +23,21 @@ export const useUploadFile = (docId: string) => {
...
+            t('File size exceeds the maximum allowed size of {{size}}MB.', {
+              size: Math.round(maxSize / (1024 * 1024)),
+            }),
+          ],
+        });
+        setSizeError(error);
+        throw error;
+      }
+      setSizeError(null);
+
```
AntoLC: I would add this snippet directly in `createDocAttachment`, passing the maxSize as a prop, you will not have to deal with states like that.

AntoLC (commented):
The error message is displaying correctly in dev mode actually, and in production the error message works in some case:

Image

The root problem is with files more than 20mo, ngnix as a limitation on that:
https://github.com/suitenumerique/docs/blob/708fa837cc40cd2927e5b4f678d7c5e298ddce31/src/helm/env.d/feature/values.impress.yaml.gotmpl#L176

But yes, good job, adding a guard frontside solves this issue.

I would not overengineered too much with the events on pasting/dropping, it is the concern of the upload part.

src/frontend/apps/impress/src/features/docs/doc-editor/components/BlockNoteEditor.tsx
magopian: You're right, i tried to detect the drag and drop and copy and paste (handled by BlockNote without going through `useUploadFile`) to avoid the "loading..." text.
I'm going to simplify the code, have the same exact behavior for the following 3 scenarios, and then cleanup the "loading..." text afterwards (if there's an easy way to do it, otherwise maybe it's not that big of a deal when recovering from an error?).

Here are the 3 scenarios I tested to upload a file that is too big:
1. drag and drop
2. copy and paste
3. the "+" context menu

src/frontend/apps/impress/src/features/docs/doc-editor/hook/useUploadFile.tsx
magopian: Very good point on removing the `checkFileSize` utility. I simplified the state by factoring out a `buildAndReportSizeError()` utility.

Regarding removing the state entirely, I'm having a issue: it seems the size check needs to be done synchronously. When `uploadFile` is async and the check happens inside `mutateAsync`, BlockNote has already inserted a "loading" block by the time the promise rejects, and there's a race condition where it re-renders after the error message is displayed. The end result is there's a "loading..." block left there, and no error message.

I can re-investigate this and see if I can still find a way to remove the state if you think it's still better, let me know!

magopian: Ok, I found a simple fix, the code should now be simpler and more maintainable hopefully, what do you think?

magopian:
Closing in favor of #2577 which is less over-engineered 🙏

Issue #2315, opened 2026-05-22
Define emojis as decorative by default in documents


Ovgodd:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**

Emojis used inside documents can create unnecessary noise for assistive technologies.

In many cases, emojis are only used to visually support an already explicit text. When they are exposed to screen readers, they can make the content harder to understand instead of improving it.

RGAA Rule1.2 – Chaque [image de décoration](https://accessibilite.numerique.gouv.fr/methode/glossaire/#image-de-decoration) est-elle correctement ignorée par les technologies d’assistance ?

For example:

```text
Status: Validated ✅
```

In this case, the emoji is redundant because the text already provides the information.

However, when an emoji is the only way to convey information, the meaning is not accessible enough.

Example to avoid:

```text
Status: ✅
```

Here, the information “validated” only relies on the emoji.

**Describe the solution you'd like**

Emojis used in documents should be treated as decorative by default when they only support an explicit text.

The goal is to ensure that important information is always available as text and does not rely only on visual emoji interpretation.

**Do you want to work on it through a Pull Request?**

Yes.

Ovgodd:
@cyberbaloo @virgile-dev 

After looking into the codebase, this seems more like a content problem than a code problem.

Emojis typed by users inside documents are just Unicode characters rendered by the editor (BlockNote). There's no way to automatically tell if an emoji is decorative ("Validated ✅") or if it carries meaning ("Status: ✅"). It's the same character in both cases.

If we wrap all emojis with aria-hidden="true", we break accessibility when the emoji is the only way to get the information. And leaving them visible to screen readers (current behavior) is actually the safest default.

So the real solution is about content awareness: authors should always use explicit text alongside emojis, and not rely on emojis alone to convey information.

If there are specific places in the application UI (not user content) where decorative emojis are exposed to screen readers, we can fix those case by case.

WDYT ?

virgile-dev:
@Ovgodd I don't think emojis inside the content of the doc should be flaggued as decorative. 
I think the emoji what @cyberbaloo was targeting in the audit are the ones in the document title. The ones that appear both in the tree on the left panel and at the top of the doc. These ones should be falggued as decorative IMO.

cyberbaloo:
[...]

PR #2527, opened 2026-07-15
♿(frontend) hide decorative emojis in document titles from SR

Ovgodd:
## Purpose

This PR hides title emojis from assistive technologies in the application UI while keeping accessible text labels.

issue [2315](https://github.com/suitenumerique/docs/issues/2315)

## Proposal

* [x] Mark title emojis as `aria-hidden="true"` in the left panel tree (`SimpleDocItem`, `DocSubPageItem`)
* [x] Mark title emojis as decorative in the document header read-only title (`DocTitleText`)
* [x] Pass emoji separately in the left panel collapse button so `aria-label` uses text only
* [x] Use title without emoji in tree item `aria-label` values
* [x] Leave document body emojis unchanged (user content)
* [x] Manual SR check: tree items, header title, and collapse button announce text without emoji description

### DIFF
--- a/src/frontend/apps/impress/src/features/docs/doc-header/components/DocTitle.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-header/components/DocTitle.tsx
@@ -38,11 +38,16 @@ export const DocTitleText = () => {
   const { isMobile } = useResponsiveStore();
   const { currentDoc } = useDocStore();
   const { untitledDocument } = useTrans();
+  const { emoji, titleWithoutEmoji } = getEmojiAndTitle(
+    currentDoc?.title ?? '',
+  );
+  const displayTitle = titleWithoutEmoji || untitledDocument;
 
   return (
     <Box className={CLASS_DOC_TITLE} $direction="row" $align="center">
       <Text as="h2" $margin="none" $size={isMobile ? 'h4' : 'h2'}>
-        {currentDoc?.title || untitledDocument}
+        {emoji && <span aria-hidden="true">{emoji} </span>}
+        {displayTitle}
       </Text>
     </Box>
   );
--- a/src/frontend/apps/impress/src/features/docs/doc-tree/components/DocSubPageItem.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-tree/components/DocSubPageItem.tsx
@@ -144,7 +144,6 @@ const DocSubPageItemContent = (props: TreeViewNodeProps<Doc>) => {
     }
   };
 
-  const docTitle = doc.title || untitledDocument;
   const isCurrentPage = router.query?.id === doc.id;
   const isDeleted = !!doc.deleted_at;
   const actionsRef = useRef<HTMLDivElement>(null);
@@ -187,8 +186,8 @@ const DocSubPageItemContent = (props: TreeViewNodeProps<Doc>) => {
       tabIndex={-1}
       aria-label={
         isDeleted
-          ? t('{{title}} (deleted)', { title: docTitle })
-          : t('Open document {{title}}', { title: docTitle })
+          ? t('{{title}} (deleted)', { title: displayTitle })
+          : t('Open document {{title}}', { title: displayTitle })
       }
       aria-current={isCurrentPage ? 'page' : undefined}
       data-testid={`doc-sub-page-item-${doc.id}`}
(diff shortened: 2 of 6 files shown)


src/frontend/apps/impress/src/features/docs/doc-header/components/DocTitle.tsx
```diff
@@ -38,11 +38,16 @@ export const DocTitleText = () => {
   const { isMobile } = useResponsiveStore();
   const { currentDoc } = useDocStore();
   const { untitledDocument } = useTrans();
+  const { emoji, titleWithoutEmoji } = getEmojiAndTitle(
+    currentDoc?.title ?? '',
+  );
+  const displayTitle = titleWithoutEmoji || untitledDocument;
```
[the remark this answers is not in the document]
Ovgodd: This behavior is intentional, covered by tests, and the multi-emoji prefix scenario does not occur through the app’s normal flow. No change needed.

src/frontend/apps/impress/src/features/left-panel/components/LeftPanelCollapseButton.tsx
```diff
@@ -11,9 +11,11 @@ import { useLeftPanelStore } from '../stores';
 export const LeftPanelCollapseButton = ({
   ariaLabel,
   buttonTitle,
+  buttonEmoji,
 }: {
   ariaLabel: string;
   buttonTitle?: string;
+  buttonEmoji?: string | null;
```
AntoLC: Let's keep this component simple, `buttonTitle` can be a reactNode with the span emoji in it.

```suggestion
  buttonTitle?: ReactNode;
```

src/frontend/apps/impress/src/features/docs/doc-tree/components/DocSubPageItem.tsx
```diff
@@ -144,7 +144,6 @@ const DocSubPageItemContent = (props: TreeViewNodeProps<Doc>) => {
     }
   };
 
-  const docTitle = doc.title || untitledDocument;
```
AntoLC, line 147: Not sure why we remove this part, what if doc.title is empty ?

AntoLC (commented):
I am personally doubting this issue is legit. It is a emoji added by a user, so we have to cross the fingers that the user didn't add a emoji that actually contain value, because it will not be displayed anymore. Btw a user can still add a emoji at the end of the title if he wants.

image

Ovgodd:
> I am personally doubting this issue is legit. It is a emoji added by a user, so we have to cross the fingers that the user didn't add a emoji that actually contain value, because it will not be displayed anymore. Btw a user can still add a emoji at the end of the title if he wants.

Yeah I had the same concern at first actually, raised it in the issue discussion. But after getting feedback from @virgile-dev and @cyberbaloo ( I dont knoy if you saw them ? ), the take was that emojis in document titles should be treated as decorative since the title text should already be self-explanatory. Wdyt?

image
image

virgile-dev:
@Ovgodd ok

src/frontend/apps/impress/src/features/docs/doc-tree/components/DocSubPageItem.tsx
Ovgodd: Good catch the diff looking suspicious

We didn't drop the fallback, it's just moved up: `displayTitle = titleWithoutEmoji || untitledDocument` (same behavior as before, just without the duplicate `docTitle` variable).

Ovgodd:
@AntoLC

> I am personally doubting this issue is legit. It is a emoji added by a user, so we have to cross the fingers that the user didn't add a emoji that actually contain value, because it will not be displayed anymore. Btw a user can still add a emoji at the end of the title if he wants.


Just reopening this thread 🙂

The leading emoji is already treated as a document icon in the product (`DocIcon`, add/remove icon, `updateDocEmoji`), not really as free-form title text. And if someone puts an emoji elsewhere in the title, it still gets announced.

Happy to keep discussing if you still feel unsure about it!

AntoLC (approved):
I am still quite sure that this pull request and the issue are not legit, but i will not fight about it.

It is emoji added by users, we cannot control what users are doing with the meaning of the emoji that they add, it is different that a emoji in a button:
Image

If users classify their docs in this way (by example students name that are admit or not), now it is impossible to know the meaning of the title with assistive technologie:
- ❌ Dupond Francois
- ❌ Fremont Claude
- ✅ Sarah Tutu

image

PR #2536, opened 2026-07-23
📝(docs) fix resource server code snippet

jmaupetit:
## Proposal

One needs to be authenticated while requesting the external API.


lunika (commented):
Thanks !

PR #2539, opened 2026-07-27
♻️(frontend) change favorite to star

AntoLC:
## Purpose

Change the favorite to star, it will be part of the grid list as the others, we can select them directly from a filter in the url.

Ref: #2364 

## Demo

image


AntoLC:
@coderabbit review

Ovgodd (changes requested):
Nice,

One small thing : the menu now says Star / Unstar, but the live announcements in `useCreateFavoriteDoc `/ `useDeleteFavoriteDoc `still say “pin”:

- Document pinned successfully!
- Failed to pin the document.
- Document unpinned successfully!
- Failed to unpin the document.

Could we switch those to starred / unstarred so the screen reader matches the UI?

Ovgodd (approved):

PR #2542, opened 2026-07-27
🐛(y-provider) send X-Forwarded-Proto on internal backend calls

### LINKED ISSUE #2541: Collaboration never connects when y-provider reaches the backend directly: SECURE_SSL_REDIRECT 301s the internal call
## Summary

On a self-hosted deployment where the y-provider reaches the backend over the internal docker network (`COLLABORATION_BACKEND_BASE_URL=http://docs-backend:8000`), **no collaboration connection can ever be established**. Every `onConnect` fails with `Backend error: Unauthorized`.

The cause is not authentication. It is `SECURE_SSL_REDIRECT`.

## Mechanism

1. `Production.SECURE_SSL_REDIRECT = True` is hardcoded in `src/backend/impress/settings.py` and cannot be driven by an environment variable.
2. Django only knows TLS is terminated in front of it via `SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")`.
3. The y-provider's internal call does not go through the reverse proxy, so nothing sets that header. Django answers **301 to `https://docs-backend:8000`**.
4. The client follows the redirect and speaks TLS to a port serving plain HTTP.

`src/frontend/servers/y-provider/src/api/collaborationBackend.ts` builds requests with a **fixed** header set:

```ts
headers: {
  cookie: requestHeaders['cookie'],
  origin: requestHeaders['origin'],
  'X-Y-Provider-Key': Y_PROVIDER_API_KEY,
},
```

Nothing from the incoming request is relayed, so *every* call is affected — `fetchCurrentUser` and `fetchDocument` alike. This is not limited to unauthenticated probes.

## Reproduction (from inside the y-provider container, v5.4.1)

```
$ wget -S --spider http://docs-backend:8000/api/v1.0/config/
  HTTP/1.1 301 Moved Permanently
  Location: https://docs-backend:8000/api/v1.0/config/
  ...
  error:0A00010B:SSL routines:tls_validate_record_header:wrong version number
  wget: error getting response: Connection reset by peer

$ wget -S --spider --header="X-Forwarded-Proto: https" http://docs-backend:8000/api/v1.0/config/
  HTTP/1.1 200 OK
```

Backend log showed 46 × `301 Moved Permanently` on `/api/v1.0/documents/<id>/` within the last 200 lines, all from the y-provider's container IP.

y-provider log before:

```
onConnect: backend error write EPROTO ... wrong version number
[onConnect] Backend error: Unauthorized
```

after setting the header:

```
Connection established on room: ba8c9fba-… canEdit: true
```

## Why it can go unnoticed

Document *creation* through the conversion endpoint works fine — it is a different path. An instance can look healthy, list documents, and still be unable to open a single one for editing. Ours was in that state from day one and we only found it while verifying an unrelated deploy.

## Possible fixes

The smallest one is to set the header on the internal call in `collaborationBackend.ts`:

```ts
'X-Forwarded-Proto': 'https',
```

We are running that on our instance and collaboration works. Happy to send it as a PR.

That said, you may prefer a different shape — relaying the incoming `x-forwarded-proto` instead of hardcoding it, exempting the API paths from the redirect, or simply documenting that `COLLABORATION_BACKEND_BASE_URL` must point through a proxy that sets the header. Tell us which you'd rather have and we'll write it that way.

Also happy to be told we have misconfigured something — but `SECURE_SSL_REDIRECT` not being env-driven leaves little room to avoid this when the internal call is plain HTTP.

Happens on `v5.4.1` and the code is unchanged on `main` at the time of writing.

lucletoffe:
## Purpose

Fixes #2541.

On a self-hosted deployment where `COLLABORATION_BACKEND_BASE_URL` points at the backend directly (internal docker network, e.g. `http://docs-backend:8000`), **collaboration can never connect**. Every `onConnect` fails with `[onConnect] Backend error: Unauthorized`, whatever the credentials.

The cause is not authentication, it is `SECURE_SSL_REDIRECT`:

1. `Production.SECURE_SSL_REDIRECT = True` is hardcoded in `src/backend/impress/settings.py` and cannot be driven by an environment variable.
2. Django only learns that TLS was terminated in front of it through `SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")`.
3. The y-provider call to the backend is internal: it does not cross the reverse proxy, so nothing sets that header. Django answers `301` to `https://<host>:8000`, and the client follows it — speaking TLS to a port that serves plain HTTP (`wrong version number`).

Reproduced from inside the y-provider container (v5.4.1):

```
$ wget -S --spider http://docs-backend:8000/api/v1.0/config/
  HTTP/1.1 301 Moved Permanently
  Location: https://docs-backend:8000/api/v1.0/config/
  error:0A00010B:SSL routines:tls_validate_record_header:wrong version number

$ wget -S --spider --header="X-Forwarded-Proto: https" http://docs-backend:8000/api/v1.0/config/
  HTTP/1.1 200 OK
```

The backend log showed 46 × `301 Moved Permanently` on `/api/v1.0/documents/<id>/` within the last 200 lines, all from the y-provider container IP.

## Proposal

* [x] `collaborationBackend.ts` always sends `X-Forwarded-Proto: https`, next to the `cookie` and `origin` headers it already relays.

  `https` is hardcoded rather than relayed from the incoming request. An ingress configured with `use-forwarded-headers: true` — required as soon as a load-balancer sits upstream — passes the client's own header through unchanged. A client that sends `X-Forwarded-Proto: http` on its WebSocket upgrade would therefore cause Django to answer a 301, axios would follow the redirect to a plain-HTTP port, and the connection would break. That is a vector this fix has no reason to introduce. Since `SECURE_PROXY_SSL_HEADER` is only declared in the `Production` class where `SECURE_SSL_REDIRECT = True` already makes a plain-HTTP deployment unreachable, `https` is the safe default given the current Production settings.

* [x] Test suite updated: the two cases that tested relay behaviour are removed; one test checks that `X-Forwarded-Proto: https` is always sent to the backend even when the incoming request carries `X-Forwarded-Proto: http`.
* [x] Changelog entry.

Ran locally on this branch: `yarn COLLABORATION_SERVER run test` (6 files, 56 tests passing), `run lint`, `run build`.

### Why not fix this on the backend instead?

A maintainer might reasonably suggest extending `SECURE_REDIRECT_EXEMPT` (`src/backend/impress/settings.py`) with the internal API prefixes called by the y-provider, rather than touching the y-provider itself. That approach is legitimate and would fix the problem at its source in Python.

The trade-off is maintenance surface: every internal path that could be called before the reverse proxy sets the header would need to be listed and kept in sync as the API evolves. The y-provider fix applies unconditionally for any topology where `COLLABORATION_BACKEND_BASE_URL` points at the backend directly.

It is worth noting that the topology recommended by the official Helm chart (`documentation/examples/helm/impress.values.yaml`, where `COLLABORATION_BACKEND_BASE_URL` is set to the public HTTPS URL) does **not** trigger this bug at all — the call goes through the ingress, which sets the header correctly. Routing an in-cluster call out through the public ingress and back in adds unnecessary latency and a dependency on external DNS resolution from inside the cluster; the direct-backend URL is the natural choice for in-cluster deployments, and this fix makes it work without requiring any backend change.

### Alternative considered

Making `SECURE_SSL_REDIRECT` configurable through an environment variable would also work, but it weakens a security default for every deployment in order to fix an internal call. Sending `https` unconditionally keeps the production posture untouched.

### General requirements

### CI requirements

### AI requirements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to
 
 ### Fixed
 
+- 🐛(y-provider) send X-Forwarded-Proto on internal backend calls #2541
 - 🐛(frontend) redirect homepage to login when homepage feat is disabled #2521
 
 ### Changed
--- a/src/frontend/servers/y-provider/__tests__/collaborationBackend.test.ts
+++ b/src/frontend/servers/y-provider/__tests__/collaborationBackend.test.ts
@@ -63,4 +63,29 @@ describe('CollaborationBackend', () => {
 
     axiosGetSpy.mockRestore();
   });
+
+  test('always sends X-Forwarded-Proto: https regardless of the incoming header', async () => {
+    const axiosGetSpy = vi.spyOn(axios, 'get').mockResolvedValue({
+      status: 200,
+      data: { id: 'test-user-id', email: 'test@example.com' },
+    });
+
+    const { fetchCurrentUser } = await import('@/api/collaborationBackend');
+
+    await fetchCurrentUser({
+      cookie: 'test-cookie',
+      'x-forwarded-proto': 'http',
+    });
+
+    expect(axiosGetSpy).toHaveBeenCalledWith(
+      'http://app-dev:8000/api/v1.0/users/me/',
+      expect.objectContaining({
+        headers: expect.objectContaining({
+          'X-Forwarded-Proto': 'https',
+        }),
+      }),
+    );
+
+    axiosGetSpy.mockRestore();
+  });
 });
--- a/src/frontend/servers/y-provider/src/api/collaborationBackend.ts
+++ b/src/frontend/servers/y-provider/src/api/collaborationBackend.ts
@@ -52,6 +52,19 @@ interface Doc {
   };
 }
 
+/**
+ * In production the backend sets SECURE_SSL_REDIRECT and only trusts
+ * X-Forwarded-Proto to tell whether TLS was terminated upstream. This call is
+ * internal, so when COLLABORATION_BACKEND_BASE_URL points at the backend
+ * directly it never crosses the reverse proxy: nothing sets the header, Django
+ * answers a 301 to https://<host>:<port>, and the client then speaks TLS to a
+ * port that serves plain HTTP.
+ *
+ * Always send https: it is the safe default given the current Production
+ * settings. Relaying the client's own X-Forwarded-Proto would let a client
+ * that sends X-Forwarded-Proto: http on its WebSocket upgrade force a 301 and
+ * break its own connection — a vector this fix has no reason to introduce.
+ */
 async function fetch<T>(
   path: string,
   requestHeaders: IncomingHttpHeaders,
@@ -63,6 +76,7 @@ async function fetch<T>(
         cookie: requestHeaders['cookie'],
         origin: requestHeaders['origin'],
         'X-Y-Provider-Key': Y_PROVIDER_API_KEY,
+        'X-Forwarded-Proto': 'https',
       },
     },
   );


lucletoffe:
Good catch, fixed in e6f76923: the value is now restricted to `http` or `https` before being forwarded, anything else (including a chained `https, http`) falls back to `https`. Added a test for it — 58 tests passing in the y-provider workspace.

lucletoffe:
Correction to my earlier comment — the branch was squashed into a single commit after I wrote it, and the approach changed with it.

`e6f76923` is no longer part of this branch, and the value is no longer relayed at all: `X-Forwarded-Proto: https` is now hardcoded (`70c9249`).

Why the change. An ingress configured with `use-forwarded-headers: true` — required as soon as a load balancer sits upstream — passes the client's own `X-Forwarded-Proto` through unchanged. A client sending `X-Forwarded-Proto: http` on its WebSocket upgrade would then make Django answer a 301, axios would follow it to a plain-HTTP port, and the connection would break. Validating the relayed value narrows that down but still leaves the client in control of it, and this fix has no reason to introduce that vector at all.

Since `SECURE_PROXY_SSL_HEADER` is only declared in the `Production` class, where `SECURE_SSL_REDIRECT` is enabled too, `https` is the one value that is always correct on this internal call.

Apologies for the noise — the earlier comment describes code that is no longer in the PR.

lucletoffe:
`lint-git` was failing on the commit title only — 76 characters against gitlint's 72 limit. Force-pushed `b1affa7b`: title shortened to `🐛(y-provider) always send X-Forwarded-Proto: https to the backend` (65), and the branch rebased onto `eed828d8` so it is no longer behind `main`. The diff is unchanged.

The workflow runs are waiting on maintainer approval (`action_required`) — nothing else to do on my side.

PR #2544, opened 2026-07-29
🔒️(yprovider) readOnly connections awareness propagation

AntoLC:
## Purpose

We noticed that a read-only connection was still able to propagate awareness updates (cursor/selection presence) to other clients, even though `connectionConfig.readOnly` already blocked document sync updates.

Override `MessageReceiver.apply` to no-op for any read-only connection instead of delegating to the original implementation, so neither document updates nor awareness updates ever propagate from it. 
The connection itself is left open; only the message is dropped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,10 @@ and this project adheres to
 - 🐛(backend) ignore CSPs for API docs in development
 - 🐛(frontend) export images embedded with a relative url #2573
 
+### Security
+
+- 🔒️(yprovider) readOnly connections awareness propagation #2544
+
 ### Changed
 
 - ♿️(frontend) use semantic `<dl>` structure in document info card #2379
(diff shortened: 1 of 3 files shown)


dmonad:
Looks good! I will add an option to ignore awareness messages in yhub as well.

The one thing to look out for is whether the connection resets regularly. y-websocket-server, for example, will disconnect the user if it did not receive an update from that user.

AntoLC:
> Looks good! I will add an option to ignore awareness messages in yhub as well.
> 
> The one thing to look out for is whether the connection resets regularly. y-websocket-server, for example, will disconnect the user if it did not receive an update from that user.

@dmonad - Thank you for the review - Yes you were right, the ping pong to keep the connection alive was based on awareness , at least in Hocuspocus v3.4.4, I think they change this mechanism in their last version (>v4).

I did a fixup commit - when we get a awareness message from `readOnly` connection, we still intercept it, but send our own awareness message to keep this ping pong mechanism.

Wdyt ?

dmonad:
Please investigate if that fixes the issue. y-websocket doesn't disconnect, it is the server that disconnects after a timeout. Returning the awareness message to the client might not help.

AntoLC:
This PR should not be necessary anymore, closed in favor of https://github.com/suitenumerique/docs/pull/2556

Issue #2417, opened 2026-06-10
Unify dropdown menus and streamline actions across views


virgile-dev:
## Feature Request

**Is your feature request related to a problem or unsupported use case? Please describe.**
Currently, the dropdown menus triggered by clicking the **`...`** (ellipsis) button vary depending on the view:
- Document editor view
- Document tree 
- Document list view

This inconsistency creates confusion for users navigating between these interfaces.

Additionally,  Arnaud made aware of the some user feedback on this menu
- The **"Copy Markdown"** is not understood by most users who don't know what markdown (and ctrl A + ctrl C has the same effect)
- The **"Print"** option is currently burried under the **"Download"** menu, which is unintuitive. It should be moved to the main **`...`** dropdown menu for better discoverability.

---

**Describe the solution you'd like**
During a discussion with @robin-lecomte  and @arnaud-robin , we agreed on the following improvements:
1. **Unify dropdown menus**: Ensure the **`...`** dropdown menu is identical across all document views (editor, tree, and list) to maintain consistency.
2. **Remove Copy Markdown**:
   - Delete the **"Copy Markdown"** button, as its functionality is already covered by standard keyboard shortcuts.
3. **Relocate the "Print" option**:
   - Move **"Print"** from the **"Download"** submenu to the main **`...`** dropdown menu for easier access.

robin-lecomte:
<img width="2972" height="1242" alt="Image" src="https://github.com/user-attachments/assets/39e6b66a-d462-4184-b508-4702821a191e" />

I suggest organizing this new unified menu into four specific categories:
1. Share (copy link + share)
2. View and export (present + download + print)
3. Manage (star, duplicate, move)
4. Delete

[FIGMA](https://www.figma.com/design/qdCWR4tTUr7vQSecEjCyqO/Docs?node-id=15417-85448&t=JmLBNc6qQoELeZeP-1)

virgile-dev:
This is great thanks @robin-lecomte 
I'm unsure about 2 things.
Is this this the right place to put the word count info. Maybe somewhere in the doc view ?

<img width="1140" height="254" alt="Image" src="https://github.com/user-attachments/assets/d7e36ddf-5827-4173-9066-039908204848" />

I'm not sure we want to display the page estimate. 
We have no pages in Docs, it might not be consistent with export and the print result. I'd leave it out.

robin-lecomte:
[...]

PR #2550, opened 2026-07-31
🚸(frontend) print from document options menu

AntoLC:
## Purpose

We move the print button from the export select to the document toolbox, so that it is more accessible and easier to find.
We updated the UI of the doc toolbox to fit with the design system. 

Part of https://github.com/suitenumerique/docs/issues/2417.

## Demo

image


Ovgodd (approved):
good!

PR #2551, opened 2026-07-31
📄(frontend) allowed partially export when MIT

### LINKED ISSUE #1885: ♻️ Export partially MIT
## Feature

All the feature export is disabled when `NEXT_PUBLIC_PUBLISH_AS_MIT` is at `true`, but actually we have some features (PRINT / HTML) inside the export that are MIT, so they could be proposed even when the user use Docs in MIT mode.

## What to do

When `NEXT_PUBLIC_PUBLISH_AS_MIT` is at `true`, we should still propose the export but just with the MIT feature.

AntoLC:
## Purpose

When MIT, we were not able to export anything.
But actually, the export HTML and Print are allowed for MIT, so we should allow them.
We now allow partially export when MIT, but the AGPL export is still not allowed (pdf / ODT / docx).


src/frontend/apps/impress/src/features/docs/doc-export/components/ModalExport.tsx
Ovgodd:
```suggestion
{exportAGPL
  ? t(
      'Export your document to print or download in .docx, .odt, .pdf or .html(zip) format.',
    )
  : t(
      'Export your document to print or download in .html(zip) format.',
    )}
```
nit: Now that the export modal is available in MIT mode, this description still mentions .docx, .odt and .pdf, which are not proposed in that configuration.

Should we adapt the description text depending on the licence mode?

Ovgodd (changes requested):

Ovgodd (approved):

PR #2561, opened 2026-08-04
🛂 use jwt token for converter services

AntoLC:
## Purpose

This pull request migrates the authentication mechanism between the Django backend and the y-provider service from a shared secret (`Y_PROVIDER_API_KEY`) to short-lived, JWT-signed admin tokens. The Django service now signs tokens with its private key and publishes its public key via a JWKS endpoint, which the y-provider verifies. The change is reflected in both backend and frontend code, as well as in documentation and tests. Additionally, test utilities for JWT key generation and signing are refactored and centralized.

These changes improve security by eliminating long-lived shared secrets and ensure robust, testable authentication between Django and y-provider.

### DIFF
--- a/src/backend/core/tests/test_services_converter_services.py
+++ b/src/backend/core/tests/test_services_converter_services.py
@@ -3,6 +3,7 @@
 from base64 import b64decode
 from unittest.mock import MagicMock, patch
 
+import jwt
 import pytest
 import requests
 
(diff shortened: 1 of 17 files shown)


src/frontend/servers/y-provider/src/middlewares.ts
```diff
@@ -16,11 +12,27 @@ export const corsMiddleware = cors({
...
+// Cached across requests: fetches the Django backend's public keys lazily and
+// keeps them until their "kid" no longer matches a token, per jose's own policy.
+const jwks = createRemoteJWKSet(new URL(JWKS_URL));
+
+/**
+ * Verify that the given token is an admin JWT signed by the Django backend.
+ */
+const isValidAdminToken = async (token: string): Promise<boolean> => {
+  try {
+    const { payload } = await jwtVerify(token, jwks, { algorithms: ['RS256'] });
```
lunika: Maybe the string `RS256` can be set in a constant `JWT_ALGORITHM`. The day we change it we only have to change the constant value.

AntoLC:
@coderabbit review

src/backend/core/tests/test_services_converter_services.py
```diff
@@ -12,13 +13,29 @@
...
-def test_auth_header(settings):
-    """Test authentication header generation."""
-    settings.Y_PROVIDER_API_KEY = "test-key"
+
+@pytest.fixture(autouse=True)
+def jwt_settings(settings):
+    """Setup valid settings for the JWT service used to sign the auth header."""
+    settings.JWT_PRIVATE_KEY = PRIVATE_KEY
+    settings.JWT_TOKEN_LIFETIME = 3600
+    return settings
```
lunika: This return is not needed

lunika (approved):

PR #2575, opened 2026-08-10
💄(frontend) use the same highlight color for cells and moves

### LINKED ISSUE #2080: Color of highlight is not the same
### What’s broken?

The color of cell highlight and move highlight is not the same.

### What did you expect to happen?

It'd be great to use the same colors.

### Steps to reproduce

Image

### BlockNote version

_No response_

### Environment

Firefox 148 64 bit NixOS

### Additional context

_No response_

### Contribution

- [ ] I'd be interested in contributing a fix for this issue

RISK-alt:
## Purpose

Fixes #2080

Three different colors are used to highlight things in the editor, none of them from the palette:

| what | where it comes from | color |
| --- | --- | --- |
| selected table cells | `.ProseMirror .selectedCell:after` (`@blocknote/core` style.css) | `#c8c8ff66`, i.e. `#e9e9ff` over white |
| drop cursor of a moved block | `dropCursor.color` default in the `DropCursor` extension | `#ddeeff` |
| drop indicator of a moved table row/column | `.bn-table-drop-cursor` | `#adf` |

The first two are the ones compared in the issue screenshot.

Following @robin-lecomte's answer, all three now use `var(--c--contextuals--background--semantic--brand--tertiary)` (`brand-050`, `#eef1fa`).

## Proposal

* [x] set `dropCursor.color` on the main editor to the brand token, the option `@YousefED` mentions in TypeCellOS/BlockNote#2569 is exposed by `useCreateBlockNote`, so no `!important` on the inline style the extension writes
* [x] override `.selectedCell:after` and `.bn-table-drop-cursor` with the same token in `doc-editor/styles.tsx`

The cell overlay is drawn on top of the cell content (`position: absolute; inset: 0; z-index: 2`), which is why BlockNote uses a translucent color. An opaque token there hides the text, so the overlay gets `mix-blend-mode: multiply`: over the white cell it renders exactly `#eef1fa`, and the content stays readable. This is the same treatment already used for the comment marks in `doc-comments/styles.tsx`.

Two points worth a look from a design standpoint, both a one line change if you want them different:

* `brand-050` on white gives a 1.06:1 contrast ratio for the drop cursor, against 1.13:1 for the current `#ddeeff`, so the indicator of a moved block becomes slightly fainter than it is today. `--c--contextuals--background--semantic--brand--secondary` (`#dde2f5`) would keep it closer to the current visibility while staying in the palette.
* I included `.bn-table-drop-cursor` (moving a table row or column) since it is the same "move" indicator inside the very table the issue is about, and leaving it at `#adf` would keep a third color around. Happy to drop it from the PR if you would rather keep the scope to the two colors of the screenshot.

## Testing

No automated test: the change is a color override with no behaviour attached, and the editor has no visual regression setup. Checked manually that the token resolves through the inline style the drop cursor extension sets (`element.style.backgroundColor = "var(--c--contextuals--background--semantic--brand--tertiary)"` computes to `rgb(238, 241, 250)`), and that the blended overlay keeps the cell content readable.

`tsc --noEmit`, `eslint` and `prettier --check` are clean, `doc-editor` unit tests pass (21).

### General requirements

* [x] I have read and followed the contributing guidelines
* [x] I have read and agreed to the Code of Conduct

### CI requirements

### AI requirements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to
 ### Changed
 
 - ♿️(frontend) use semantic `<dl>` structure in document info card #2379
+- 💄(frontend) use the same highlight color for cells and moves #2575
 
 ### Fixed
 
--- a/src/frontend/apps/impress/src/features/docs/doc-editor/components/BlockNoteEditor.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-editor/components/BlockNoteEditor.tsx
@@ -196,6 +196,9 @@ export const BlockNoteEditor = ({ doc, provider }: BlockNoteEditorProps) => {
         },
         showCursorLabels: showCursorLabels as 'always' | 'activity',
       },
+      dropCursor: {
+        color: 'var(--c--contextuals--background--semantic--brand--tertiary)',
+      },
       dictionary: {
         ...localesBN[langLocalesBN as keyof typeof localesBN],
         ...(localesBNMultiColumn && {
--- a/src/frontend/apps/impress/src/features/docs/doc-editor/styles.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-editor/styles.tsx
@@ -150,6 +150,19 @@ export const DocsEditorStyle = createGlobalStyle`
     }
 
     /**
+    * Tables
+    */
+    .ProseMirror .selectedCell:after {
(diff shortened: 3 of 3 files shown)


AntoLC (approved):
Nice, thank you for your contribution 🎉

PR #2577, opened 2026-08-10
 🥅(frontend) warn before uploading a file over the size limit

### LINKED ISSUE #2460: 🐛 Big file upload fails as unknown error with poor UX feedback

RISK-alt:
## Purpose

Fixes #2460

Dropping a file over the limit into the editor shows the bare "unknown error" from the issue's screenshot. Two things are missing.

**Nothing checks the size before sending.** The backend does enforce `DOCUMENT_IMAGE_MAX_SIZE` in [`FileUploadSerializer.validate_file`](https://github.com/suitenumerique/docs/blob/c9e32e34/src/backend/core/api/serializers.py#L649-L656), but the setting is not part of `/api/v1.0/config/`, so the frontend has nothing to check against. The user waits for the whole upload before learning it was refused — and with nginx in front, the request never even reaches Django.

**And the 413 carries no readable cause.** `errorCauses` parses the body unconditionally:

```ts
const errorsBody = (await response.json()) as Record<string, string | string[]> | null;
```

nginx answers its 413 with an HTML error page, so `response.json()` rejects, `errorCauses` throws instead of returning, and `APIError` is never built. The editor is left with an error carrying no cause, which is what it renders as "unknown error".

## Proposal

* [x] expose `DOCUMENT_IMAGE_MAX_SIZE` in the config endpoint, next to `CONVERSION_FILE_MAX_SIZE` which the import feature already uses that way
* [x] check the file size in `useUploadFile` before calling the API, and show an error toast naming the file and the limit — same wording as the document import, so both paths read alike
* [x] make `errorCauses` tolerate a body it cannot parse rather than throwing, which fixes every proxy-level error, not just this one
* [x] map a 413 with no usable cause to an explicit message, so instances whose proxy limit is lower than `DOCUMENT_IMAGE_MAX_SIZE` still get something readable

The size-formatting logic was inlined in `useImport`; it moved to `formatFileSize` in `utils/string.ts` and is now shared by both call sites rather than copied.

This follows what `🛂(frontend) use max size and extension from config` (b8bdcbf7) did for the import feature: same shape, same set of files.

I did not touch `translations.json` — the English source strings are picked up by the Crowdin sync.

## Testing

Added:

* `formatFileSize` unit tests
* an `errorCauses` test for a non-JSON body
* `useUploadFile` tests: a file under the limit reaches the API, a file over it is rejected with the toast and no request at all

Full impress suite on Node 26: 284 passing. Four failures in `features/auth/__tests__/utils.test.tsx` and `docs-grid/components/__tests__/DocsGridItemDate.test.tsx` are pre-existing — they fail identically on a clean `main` here, `localStorage` not being available under my local Node. `tsc --noEmit` and `eslint` are clean.

Note for reviewers: `CONFIG` in `apps/e2e/__tests__/app-impress/utils-common.ts` had to gain the new key, since `config.spec.ts` asserts `toStrictEqual(CONFIG)` against the live payload.

### General requirements

* [x] I have read and followed the contributing guidelines
* [x] I have read and agreed to the Code of Conduct

### CI requirements

* [ ] I made sure that all existing tests are passing — front-end suite is green locally, waiting on the CI for the back-end and the e2e

### AI requirements


AntoLC:
Duplicate of https://github.com/suitenumerique/docs/pull/2522

AntoLC (approved):
Thank you for your contribution 🙏

PR #2584, opened 2026-08-13
✨(frontend) duplicate with subdocuments

### LINKED ISSUE #1327: Duplicate with sub-docs
## Bug Report

**Problematic behavior**
When you duplicate a doc, the sub-doc are not duplicated.

AntoLC:
## Purpose

We can now choose to duplicate documents along with their subdocuments. 
A confirmation modal will now appear to let the user choose whether to include subdocuments in the duplication, only if the document has subdocuments.
A toast notification will inform the user and give the possibility to undo the duplication if needed.

## Demo

https://github.com/user-attachments/assets/223336f9-3b5d-4e2b-8a03-beea3b695bff


AntoLC:
@coderabbit review

src/frontend/apps/impress/src/features/docs/doc-management/components/ConfirmationDuplicateModal.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+  return (
+    <Modal
+      isOpen
+      closeOnClickOutside
+      hideCloseButton
+      onClose={onClose}
+      aria-label={t('Confirmation to duplicate the document')}
+      aria-labelledby="modal-duplicate-doc-title"
+      aria-describedby="modal-duplicate-doc-desc"
```
Ovgodd: This aria-describedby still points at the paragraph when it's empty (not a root). Either we always put a sentence in there, or we drop the attribute when there's nothing to read. wdyt ?

maybe have something like 

```
<Text id="modal-duplicate-doc-desc" as="p" ...>
  {isTopRoot
    ? t('The copy will be private and added to My docs.')
    : t('The copy will be added next to this document.')}
</Text>
```

src/frontend/apps/impress/src/cunningham/cunningham-style.css
```diff
@@ -57,6 +57,18 @@
...
+*/
+.c__checkbox:focus-within {
+  border-color: transparent;
+  box-shadow: none;
+}
+
+.c__checkbox .c__checkbox__wrapper:focus-within {
+  outline: none;
+}
+
```
Ovgodd, line 71:
```suggestion
.c__checkbox input:focus-visible {
  outline: 2px solid var(--c--globals--colors--brand-400);
  outline-offset: var(--c--globals--spacings--4xs);
}
```

The full-row focus box is too much, fine to remove that. What's left looks like a border because *:focus-visible uses a negative outline-offset. What about a ring on the checkbox box itself, with a positive offset, so there's a gap?

Before : 

Image

After : 

Image

Ovgodd (changes requested):
nice ! little suggestions in term of accessibility

the toast part will be fixed in the new toast ui from ui kit

Ovgodd (approved):
👍

PR #2585, opened 2026-08-13
🐛(frontend) fix clipped formatting toolbar in new comment composer

maboukerfa:
## Purpose

When you highlighted text in a document and started writing a new comment, the little pop-up menu for styling your text doesn't show up. 

## Proposal

It lets that formatting menu show in full above the new-comment box, so you can bold, italicize, or link text while writing a comment, exactly the way you already could when editing one.

Screenshot 2026-08-13 at 15 51 54

### DIFF
--- a/src/frontend/apps/e2e/__tests__/app-impress/doc-comments.spec.ts
+++ b/src/frontend/apps/e2e/__tests__/app-impress/doc-comments.spec.ts
@@ -206,7 +206,20 @@ test.describe('Doc Comments', () => {
     await editor.getByText('Hello').selectText();
     await page.getByRole('button', { name: 'Add comment' }).click();
 
-    await thread.getByRole('paragraph').first().fill('This is a new comment');
+    // The composer of a new thread must not clip its formatting toolbar
+    await expect(thread).toHaveCSS('overflow', 'visible');
+
+    // Write the new comment and select it to reveal the formatting toolbar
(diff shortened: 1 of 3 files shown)


src/frontend/apps/e2e/__tests__/app-impress/doc-comments.spec.ts
```diff
@@ -228,6 +228,43 @@ test.describe('Doc Comments', () => {
...
+    );
+    await expect(boldButton).toBeVisible();
+    await boldButton.click();
+
+    // Save the new comment
+    await thread.locator('[data-test="save"]').click();
+
+    // The comment is saved and keeps its bold formatting
+    await expect(
+      thread.locator('strong').getByText('Bold comment'),
```
AntoLC: Could you perform this check in the "it checks the comments interactions" test instead? I'm trying to limit the number of E2E tests, as they take quite a bit of time.

AntoLC (commented):
Nice ! I didn't even notice this bug before.

src/frontend/apps/e2e/__tests__/app-impress/doc-comments.spec.ts
maboukerfa: Alright

AntoLC (approved):
Nice, thank you

PR #2619, opened 2026-08-31
🐛(backend) manage async support for Docs custom middleware

lunika:
## Purpose

Docs have 2 custom middleware, both are only managing sync requests. With Python 3.13 we didn't have any errors, but since we upgraded to Python 3.14, we have a CancelledError exception. We decided to use the MiddlewareMixin from Django that is sync and async capable and will be responsible for executing both middleware in the good mode.

## Proposal

* [x] 🐛(backend) manage async support for Docs custom middleware


jmaupetit (approved):

AntoLC (approved):
Let's give a try.
I didn't get any problem with big documents on Chrome so let's not add a popup.

And good news, I don't succeed to loose data with Firefox when `keepalive` is set, so until we cross the max body size, I think it is safe to remove the alert modal.

Image

PR #2620, opened 2026-08-31
♻️(frontend) unified menu

### LINKED ISSUE #2417: Unify dropdown menus and streamline actions across views

AntoLC:
## Purpose

Depend the part of the app, we had different implementations of the toolbox menu. 
This commit unifies the implementation and uses the same component for both the docs grid, the doc tree and the doc header. It will make it easier to maintain and add new features to the toolbox menu.

Adding the toolbox in the tree highlighted some problems, we refactorized the accessibility of the tree to work correctly with the new component.

## Proposal

image

### DIFF
--- a/src/frontend/apps/impress/src/features/docs/doc-management/components/DocToolBox.tsx
+++ b/src/frontend/apps/impress/src/features/docs/doc-management/components/DocToolBox.tsx
@@ -1,14 +1,14 @@
 import {
   Button,
+  ButtonProps,
   DropdownMenu,
   DropdownMenuItem,
-  useTreeContext,
 } from '@gouvfr-lasuite/ui-components';
 import { Present } from '@gouvfr-lasuite/ui-components/icons';
 import { announce } from '@react-aria/live-announcer';
 import dynamic from 'next/dynamic';
 import { useRouter } from 'next/router';
-import { useEffect, useMemo, useState } from 'react';
+import { memo, useEffect, useMemo, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 
 import { Text } from '@/components/Text';
@@ -17,15 +17,16 @@ import { getWordCount } from '@/docs/doc-editor/utils';
 import { printDocumentWithStyles } from '@/docs/doc-export/utils_print';
 import { usePresenterStore } from '@/docs/doc-presenter/stores';
 import { useDetachDoc } from '@/docs/doc-tree/api/useDetach';
+import { useTreeContextOrNull } from '@/docs/doc-tree/utils';
 import { useAuth } from '@/features/auth';
 import ContentCopyIcon from '@/icons/copy.svg';
 import DocMoveInIcon from '@/icons/doc-move-in.svg';
 import DocMoveOutIcon from '@/icons/doc-move-out.svg';
-import DownloadSVG from '@/icons/download.svg';
-import HistorySVG from '@/icons/history.svg';
-import LeaveSVG from '@/icons/leave.svg';
+import DownloadIcon from '@/icons/download.svg';
+import HistoryIcon from '@/icons/history.svg';
+import LeaveIcon from '@/icons/leave.svg';
 import LinkIcon from '@/icons/link.svg';
-import MoreSVG from '@/icons/more_horiz.svg';
+import MoreIcon from '@/icons/more_horiz.svg';
 import PrintIcon from '@/icons/print.svg';
 import SharedIcon from '@/icons/shared.svg';
 import StarSlashIcon from '@/icons/star-slash.svg';
@@ -41,7 +42,7 @@ import {
   useDeleteFavoriteDoc,
   useDuplicateDoc,
 } from '../api';
-import { useCopyDocLink } from '../hooks';
+import { useCopyDocLink, useTrans } from '../hooks';
 import { Doc, Role } from '../types';
 
 const DocMoveModal = dynamic(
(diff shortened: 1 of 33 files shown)


AntoLC:
@coderabbit review

src/frontend/apps/impress/src/features/docs/doc-tree/components/DocTreeSubpages.tsx
```diff
@@ -0,0 +1,161 @@
...
+        return;
+      }
+
+      e.currentTarget
+        .querySelector<HTMLDivElement>('.c__tree-view--node')
+        ?.click();
+    },
+    [rootItemRef, treeContext],
+  );
+
```
Ovgodd, line 142:
```suggestion
if (e.key === 'Tab') {
  e.stopPropagation();
  return;
}
```
Setting tabindex="-1" here traps the keyboard.  ( tab does nt work anymore on a subdoc )

nothing gets focused, and focus stays on the row. This is a WCAG 2.1.2 failure.

src/frontend/apps/impress/src/features/docs/doc-tree/components/DocTree.tsx
Ovgodd:
```suggestion
          'Use the up and down arrow keys to move between documents, and Enter to open one. Press F2 to reach the actions of a document, the left and right arrow keys to move between them, and Escape to leave them.',

```

src/frontend/apps/impress/src/features/docs/doc-management/components/DocToolBox.tsx
```diff
@@ -314,23 +325,30 @@ export const DocToolBox = ({ doc }: DocToolBoxProps) => {
...
+        onOpenChange={(isOpen) => {
+          setOpenDropdown(isOpen);
+          onOpenChange?.(isOpen);
+        }}
         bottomMessage={
           wordCountLabel && <Text $variation="tertiary">{wordCountLabel}</Text>
         }
       >
         <Button
           aria-label={t('Open the document options')}
```
Ovgodd: Before, the grid button had `Open the menu of actions for the document: {{title}}`. Now every row uses `DocToolBox `with the same `Open the document options`, so all the buttons of the grid share one accessible name and a screen reader user can't tell them apart (WCAG 2.4.6). Same in the tree, where only the parent `role="toolbar`" gives the context.
```suggestion
    aria-label={t('Open the document options: {{title}}', {
            title: doc.title || untitledDocument,
          })}
```

Ovgodd (changes requested):
I tested the actions menu on the tree by hand, with the mouse and with the keyboard.

**What works:** Download opens correctly, and Escape brings the focus back to the "…" button as expected.

### 1. The Share modal does not receive the focus

When I open **Share** from a sub doc menu, the focus stays behind on the sub doc row instead of moving into the modal. It does not happen every time, which makes it worse: sometimes the focus goes into the modal, sometimes it stays on the row.

- **Expected:** when a modal opens, the focus moves into it, every time.
- **Impact:** a keyboard or screen reader user does not know the modal is open, and keeps navigating the tree behind it.
- **Files:** `DocToolBox.tsx`, `useTreeItemActions.ts`

### 2. After Star, Duplicate or Copy link, the focus goes back to the row

When I pick one of these actions, the menu closes and the focus lands on the sub doc row, not on the "…" button I opened the menu from. The toolbar is gone, so I have to press F2 again to do a second action on the same document.

- **Expected:** the focus returns to the "…" button when the menu closes. Escape from that button then brings me back to the row, which is already the case today.
- **Impact:** I lose my position after each action, and the behaviour is not consistent with Escape.
- **Files:** `useTreeItemActions.ts`

### 3. On a selected sub doc, the actions sometimes do not open

On the currently selected sub doc, I click on "…" or on "add a sub page" and nothing happens: the menu does not open and the buttons disappear under the cursor. It is not systematic, but it happens often on that row.

- **Expected:** the actions stay available and clickable as long as I interact with the item.
- **Impact:** the action is simply impossible to trigger, I have to move the mouse away and come back.
- **Files:** `DocSubPageItem.tsx`, `useTreeItemActions.ts`

src/frontend/apps/impress/src/features/docs/doc-management/components/DocToolBox.tsx
```diff
@@ -96,13 +97,21 @@ const ModalExport = dynamic(
...
+  doc,
+  isCurrentDoc,
+  onOpenChange,
+}: DocToolBoxProps) => {
   const { t } = useTranslation();
-  const treeContext = useTreeContext<Doc | null>();
+  const treeContext = useTreeContextOrNull<Doc | null>();
   const router = useRouter();
-  const isTopParent = doc.id === treeContext?.root?.id; // it can be a child but not for the current user
+  const isTopParent = !treeContext || doc.id === treeContext?.root?.id; // it can be a child but not for the current user
```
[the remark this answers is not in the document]
AntoLC: Yes, the cases are managed inside the modal.

src/frontend/apps/impress/src/features/docs/doc-tree/components/DocTreeSubpages.tsx
AntoLC: You can now just navigate with arrow, no need of the multiple tabs that was confusing.

Ovgodd (approved):
👍 as said together I will re-check the accessibility of this in another PR,

PR #2639, opened 2026-09-04
🐛(backend) allow to configure settings DJANGO_DATA_UPLOAD_MAX_MEMORY_SIZE

lunika:
## Purpose

Release 3.17.2 of DRF now takes care of DATA_UPLOAD_MAX_MEMORY_SIZE and is checked when the body request is parsed. Before that, DRF wasn't using it at all and we were only looking for custom settings linked to the media and conversion file upload. We must now also configure this setting.

## Proposal

* [x] 🐛(backend) allow to configure settings DJANGO_DATA_UPLOAD_MAX_MEMORY_SIZE


documentation/env.md
```diff
@@ -61,6 +61,7 @@ These are the environment variables you can set for the `impress-backend` contai
...
 | DJANGO_CORS_ALLOW_ALL_ORIGINS                   | Allow all CORS origins                                                                                                                                                     | false                                                                   |
 | DJANGO_CSRF_TRUSTED_ORIGINS                     | CSRF trusted origins                                                                                                                                                       | []                                                                      |
+| DJANGO_DATA_UPLOAD_MAX_MEMORY_SIZE                     | The maximum size in bytes that a request body may be should be equal to the maximum between `DOCUMENT_IMAGE_MAX_SIZE` and `CONVERSION_FILE_MAX_SIZE`. See: https://docs.djangoproject.com/en/6.1/ref/settings/#data-upload-max-memory-size | 20 MB |
```
jmaupetit: Looks like formatting is broken:

```suggestion
| DJANGO_DATA_UPLOAD_MAX_MEMORY_SIZE               | The maximum size in bytes that a request body may be should be equal to the maximum between `DOCUMENT_IMAGE_MAX_SIZE` and `CONVERSION_FILE_MAX_SIZE`. See: https://docs.djangoproject.com/en/6.1/ref/settings/#data-upload-max-memory-size | 20 MB |
```

src/backend/impress/settings.py
```diff
@@ -206,6 +206,8 @@ class Base(Configuration):
         environ_prefix=None,
     )
 
+    DATA_UPLOAD_MAX_MEMORY_SIZE = values.IntegerValue(20 * MB)  # 20 MB
```
jmaupetit, line 209: Would it be more consistent to also set `CONVERSION_FILE_MAX_SIZE = DATA_UPLOAD_MAX_MEMORY_SIZE` as default?

jmaupetit, line 209: And shouldn't it be configurable using an environment variable?

jmaupetit (commented):
It's just me neat picking 😉

src/backend/impress/settings.py
jmaupetit: It will as any core Django setting.

lunika: Yes it is.

lunika: Yes

jmaupetit (approved):

PR #2643, opened 2026-09-04
🐛(backend) serve the static files of the admin without whitenoise

lunika:
## Purpose

We previously removed whitenoise because its middleware is not compatible is not compatible with the ASGI application. But without it, static file for the admin are not served anymore. This commit allow to serve them using the static view from django. This view should not be used in production but we are only using it for the admin, which is not used a lot and a settings is also added to disable it if you want to serve static from a dedicated nginx for example.

## Proposal

* [x] 🐛(backend) serve the static files of the admin without whitenoise


lunika:
Other implementation in #2644

PR #2644, opened 2026-09-04
➕(backend) add servestatic dependency

lunika:
## Purpose

We removed previously whitenoise because it was not working with asgi application. By removing it we also removed the way to serve the static files in the application. There is an existing fork of whitenoise, servestatic, that manage async application and we can use it to serve static files.

## Proposal

* [x] ➕(backend) add servestatic dependency


src/helm/env.d/dev/values.impress.yaml.gotmpl
```diff
@@ -20,7 +20,7 @@ backend:
     COLLABORATION_WS_NOT_CONNECTED_READ_ONLY: False
     CONVERSION_UPLOAD_ENABLED: True
     DJANGO_CSRF_TRUSTED_ORIGINS: https://docs.127.0.0.1.nip.io
-    DJANGO_CONFIGURATION: Feature
+    DJANGO_CONFIGURATION: Production
```
jmaupetit: Sure?

jmaupetit (approved):

src/helm/env.d/dev/values.impress.yaml.gotmpl
lunika: Feature inherits from Production `class Feature(Production):`. I can revert it but it is the same.

lunika: I will revert it.

PR #2646, opened 2026-09-06
🤖mcp server oauth

AntoLC:
## Purpose

This PR lays the groundwork for an MCP server for Docs (docs-mcp): a remote [MCP](https://modelcontextprotocol.io/) server exposing three tools backed by the Django API:
- search_documents
- read_document
- create_document

It is an OAuth-protected resource server: MCP clients authenticate the user against Keycloak (Authorization Code + PKCE, public docs-mcp-client), and the server forwards that same access token, unchanged, to Django.

Django stays the sole authority on document permissions — ResourceServerAuthentication (django-lasuite) introspects the token on every call and the existing DocumentViewSet permissions/querysets decide access.

There is deliberately no RFC 8693 token exchange: the TypeScript server (Express + @modelcontextprotocol/sdk, stateless Streamable HTTP) holds no credentials of its own, only reading Keycloak's public JWKS to verify signatures.

We're eager to hear from people using it and to see it augmented — more tools, richer document operations, DCR, token exchange — so please try it and tell us what you need.

## How to do

To start the mcp server, as usual:
- `make build`
- `make run`

When your MCP is running, you can check that everything is fine by checking it with [mcp-inspector](https://modelcontextprotocol.io/docs/2026-07-28/tools/inspector), by running:
- `make mcp-inspector`
https://github.com/suitenumerique/docs/blob/b7f3becd30bf33c170788636c0a05a64ec3ffeb6/Makefile#L528-L530

You can use it with your favorite LLM (Claude,  Codex, Cursor...), here some helpers to plug it easily:
- `make mcp-claude`
- `make mcp-codex`

More here:
https://github.com/suitenumerique/docs/blob/b7f3becd30bf33c170788636c0a05a64ec3ffeb6/Makefile#L462-L527

## Extensive documentation here:

https://github.com/suitenumerique/docs/blob/feat/mcp-server-oauth/documentation/mcp_server.md

### DIFF
--- a/.github/workflows/docker-hub.yml
+++ b/.github/workflows/docker-hub.yml
@@ -60,11 +60,25 @@ jobs:
       should_push: ${{ github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'preview') }}
       docker_user: 1001:127
 
+  build-and-push-mcp:
+    uses: ./.github/workflows/docker-publish.yml
+    permissions:
+      contents: read
+    secrets: inherit
+    with:
+      image_name: lasuite/impress-mcp
+      context: .
+      file: src/frontend/servers/mcp/Dockerfile
+      target: mcp
+      should_push: ${{ github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'preview') }}
+      docker_user: 1001:127
+
   notify-argocd:
     needs:
       - build-and-push-backend
       - build-and-push-frontend
       - build-and-push-y-provider
+      - build-and-push-mcp
     runs-on: ubuntu-latest
     if: github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'preview')
     steps:
--- a/.github/workflows/ghcr.yml
+++ b/.github/workflows/ghcr.yml
@@ -158,3 +158,50 @@ jobs:
         run: |
           docker system prune -af
           docker volume prune -f
+
+  build-and-push-mcp:
+    runs-on: ubuntu-latest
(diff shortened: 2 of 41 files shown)


olivier-lacroix:
Thanks for this — evaluating this on a self-hosted deployment:

1. **The auth chain could be (mostly) IdP-agnostic — the backend already is**. The docs present the setup as Keycloak-specific (realm.json, audience mappers), but the `mcp_api` endpoints only rely on `ResourceServerAuthentication` + `MCPResourceServerPermission`, both settings-driven: we run them against a different OIDC provider end-to-end, without Keycloak or the bundled Node server. Two small changes would make the bundled `docs-mcp` Node server generic too: (a) rename the `KEYCLOAK_*` env vars to provider-neutral `OIDC_*` ones, and (b) make the expected audience claim configurable (`aud` vs `client_id`, like `OIDC_RS_AUDIENCE_CLAIM` on the backend) — the current `aud=docs-mcp` requirement needs a Keycloak audience mapper, which generic providers don't support. `realm.json` could stay as a documented dev convenience.

2. **Missing: an edit primitive.** Since bodies are BlockNote block trees with stable IDs, block-ID-addressed operations would be the natural fit for agentic editing — `POST .../content/edit/ `with `replace / insert_after / delete / append` by block ID (404 on unknown IDs, no partial application), plus `GET .../content/` returning the tree so reads and edits round-trip. Unlike full-document rewrites, ID-addressed edits merge safely with concurrent collaborators; a bulk PUT .../content/ with optimistic concurrency (409 on stale hash) could serve as the escape hatch. Happy to help contribute this.

PR #2653, opened 2026-09-07
♿️(frontend) improve keyboard tab order in document list

### LINKED ISSUE #2308: Improve keyboard navigation in the “All documents” list

Ovgodd:
## Purpose

A document row had two tab stops and two names for the same entry.
Keyboard and screen-reader users now get one announced link, then
the actions menu.

## Proposal

- [x] Put title, date and participant count on the document link
- [x] Take the share count out of the tab order
- [x] Hide the visible date from assistive tech so it is not read twice

### DIFF
--- a/src/frontend/apps/e2e/__tests__/app-impress/doc-grid-move.spec.ts
+++ b/src/frontend/apps/e2e/__tests__/app-impress/doc-grid-move.spec.ts
@@ -242,9 +242,7 @@ test.describe('Doc grid move', () => {
       .click();
 
     await expect(docsGrid.getByText(titleDoc1)).toBeHidden();
-    await docsGrid
-      .getByRole('link', { name: `Open document ${titleDoc2}` })
-      .click();
+    await docsGrid.getByRole('link', { name: new RegExp(titleDoc2) }).click();
 
     await verifyDocName(page, titleDoc2);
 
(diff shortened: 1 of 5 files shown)


Ovgodd:
Hey @cyberbaloo About the documents list 🙂

Do you think we could make the keyboard navigation a bit closer to the Tree pattern ( the left sidebar with the F2 options ) ?

The idea would be to keep a single tab stop for the documents list, navigate between documents with tab( only one tabstop ) and use F2 to enter the actions/options of the currently focused document.

Then Escape could bring the focus back to the document.

This would avoid having to Tab through every document and all of its actions, while keeping the keyboard navigation much more efficient.

What do you think?

PR #2654, opened 2026-09-07
🐛(backend) skip session creation for the liveness probe

lunika:
## Purpose

The ForceSessionMiddleware force the session creation, we wanto to
ignore it when the request is the liveness probe. The liveness probe
must not check if redis is available, this is the readiness probe job

## Proposal

* [x] 🐛(backend) skip session creation for the liveness probe
* [x] 🙈(docker) ignore **/.venv files

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,6 +5,8 @@ __pycache__
 **/*.pyc
 venv
 .venv
+**/.venv
+**/venv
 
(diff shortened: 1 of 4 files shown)


kernicPanel (approved):
Just some typos in each commits: 
wanto -> want

filder -> folder
leave -> live

PR #2659, opened 2026-09-08
🔧(backend) configure request.summary logger

lunika:
## Purpose

The request.summary logger is used by dockerflow. The INFO level is always empty and is used everytime the liveness or readiness endpoint are fetch.

## Proposal

* [x] 🔧(backend) configure request.summary logger


src/backend/impress/settings.py
```diff
@@ -1034,6 +1034,13 @@ class Base(Configuration):
                 ),
                 "propagate": True,
             },
+            "request.summary": {
+                "level": values.Value(
+                    "WARNING",
+                    environ_name="LOGGING_LEVEL_REQUEST_SUMMARY",
+                    environ_prefix="",
```
AntoLC: The other ` environ_prefix` are set to `None`.
https://github.com/suitenumerique/docs/blob/64b23e864b002b4843e5dbebfc9c97f41e9e9bcd/src/backend/impress/settings.py#L1024

AntoLC (approved):

src/backend/impress/settings.py
lunika: Nice catch

PR #2663, opened 2026-09-09
✨(frontend) Scroll to linked block when read-only

### LINKED ISSUE #2649: 🐛(Bug)  link to block flakiness
### What happened?

On big document with media, "link to block" does not work correctly.

Ex: https://docs.numerique.gouv.fr/docs/1bf73687-324a-4e26-a3ae-8c8363a4e0f3/#0aafcfa1-e350-4fc3-9b8c-3ba2298aaa95

maboukerfa:
## Purpose

This can be useful when using Docs as a **Wiki**, to point out un-authenticated users to certain sections of the doc.

## Proposal

Make the "Copy link to block" scroll for users with Read-only permissions.


ahofbeck:
additionally, besides the scrolling (which is necessary to use the link to block feature) it would be super help-ful to have this copy "link to block" not just as an editor but also as a reader.

A Wiki is curated centrally, not everyone has editing access. Some users might want to copy the link to a segment and present it to another colleague via E-Mail/chat. Both users might only have read access.

Sharepoint solves this with temporary visible icon (by clicking on it, it copies to clipboard)
image

maboukerfa:
> additionally, besides the scrolling (which is necessary to use the link to block feature) it would be super help-ful to have this copy "link to block" not just as an editor but also as a reader.
> 
> A Wiki is curated centrally, not everyone has editing access. Some users might want to copy the link to a segment and present it to another colleague via E-Mail/chat. Both users might only have read access.
> 
> Sharepoint solves this with temporary visible icon (by clicking on it, it copies to clipboard) image

Hi @ahofbeck ,
Thanks for your comment, I think that's a really good point. Let's get @virgile-dev 's opinion on this.

virgile-dev:
@maboukerfa Thank you for your contribution and spotting the bug ! 
@ahofbeck I agree it could be interesting. Could you create another issue ? 
So we discuss this separately ?  I think it's agood idea. The question is what UX do we choose ?
Do we reuse the door handle ⣿, do we display something else (a :link: for ex) ? Only on title or everything ?

AntoLC (approved):
Thank you for your contribution 🙏

PR #2668, opened 2026-09-09
⚡️(backend) speed up some slow queries

lunika:
## Purpose

We have identified some slow queries to improve. 
The first one is how the nb_accesses are calculated, the sql query was making a full scan of the document table.
The second is on the DocumentViewset.get_queryset used to list or retrieve documents. Filtering on the accesses and link_traces was leading to a full scan.

## Proposal

* [x] ⚡️(backend) stop using LEFT(value, LENGTH(path)) in sql queries
* [x] ⚡️(backend) improve DocumentViewset.get_queryset
* [x]  ⚡️(backend) increase nb_accesses cache TTL


qbey (approved):
Nice :)

I would have written a unit test for `get_self_and_ancestors_paths` even if it is already functionally tested because tests still works.

kernicPanel (approved):

PR #2679, opened 2026-09-14
🐛(frontend) fix css title issue

AntoLC:
## Purpose

Recent changes broke the way the titles were displayed when a emoji was present. It was displayed on 2
lines instead of a single line.
We fixed the CSS to ensure that titles with emojis are displayed on a single line.
We improved the way the underline is displayed with interlinking blocks.

## Demo

Before:
image
image
image

After:
image
image
image


src/frontend/apps/impress/src/features/left-panel/components/LeftPanelCollapseButton.tsx
```diff
@@ -34,9 +34,11 @@ export const LeftPanelCollapseButton = ({
         {!isSmallMobile && !!buttonTitle && (
           <FadeComponent isVisible={!!buttonTitle}>
             <Text
+              as="h3"
               $size="sm"
               $weight={700}
               $color="var(--c--globals--colors--gray-1000)"
+              $display="unset"
```
Ovgodd, line 40:
```suggestion
<Text
$size="sm"
$weight={700}
$color="var(--c--globals--colors--gray-1000)"
$display="unset"
```

`as="h3"` might not be needed here.

A `<button>` can only contain phrasing content, so a heading inside it is invalid HTML. It also shows up in the heading outline (HeadingsMap, screen readers) even though it's only a visual label. With the panel closed it can even become the only heading on the page.

The button already has an `aria-label`, so we could keep `$display="unset"` for the emoji wrapping and leave this as a `span`.

Image

Ovgodd (changes requested):
juste a little change in term of accessibility

Ovgodd (approved):

PR #2704, opened 2026-09-18
fix: enable Ctrl+Click on search results to open in new tab

### LINKED ISSUE #2603: Unable to open new tab with ctrl+click on search results
## Bug Report

**Problematic behavior**
When performing a `Ctrl + Click` (or `Cmd + Click` on macOS) on a search result link, the link does not open in a new tab as expected. Instead, it opens in the current tab, which is not handy when I want to open several docs from the results

**Expected behavior/code**
The link should open in a new tab when `Ctrl + Click` (or `Cmd + Click` on macOS) is performed, allowing users to keep the page open with the result of their search query while exploring the linked content.

**Steps to Reproduce**
1. Perform a search on the platform.
2. Hover over a search result link.
3. Press and hold the `Ctrl` key (or `Cmd` on macOS).
4. Click on the link.
5. Observe that the link does not open in a new tab.

**Environment**
- Docs version: 5.5
- Instance URL: prod

**Possible Solution**

Tested on chrome on MacOS and Firefox on Ubuntu

syamsundaryadla:
## Summary

Fixes #2603

When performing a `Ctrl+Click` (or `Cmd+Click` on macOS) on a search result link, the link now opens in a new tab as expected.

## Problem

Search result items used `cmdk`'s `Command.Item` with an `onSelect` callback that called `router.push()` — there was no native `<a>` tag, so the browser couldn't handle modifier-key clicks to open in a new tab.

## Solution

Wrapped each `DocSearchItem` in a Next.js `<Link>` component (`<a>` tag) so the browser natively handles:
- **Ctrl+Click / Cmd+Click** → opens in new tab
- **Middle-click** → opens in new tab
- **Normal click** → same-tab navigation (existing behavior preserved)
- **Enter key** → same-tab navigation (existing behavior preserved)

### Files changed

- **`DocSearchItem.tsx`**: Wrapped content in `<Link href="/docs/{id}">`. Added `onClick` handler that detects modifier keys — if modifier present, stops propagation to prevent cmdk's `onSelect` from also navigating; if normal click, prevents `<a>` default to let cmdk handle it.
- **`DocSearchContent.tsx`**: Added `onClose` prop, passed through to `DocSearchItem`.
- **`DocSearchModal.tsx`**: Passes modal's `onClose` to `DocSearchContent`.
- **`doc-search.spec.ts`**: Added e2e test verifying Ctrl+Click opens a new tab.

## Testing

- Added Playwright e2e test that Ctrl+clicks a search result and verifies a new page context opens with the correct URL.


src/frontend/apps/impress/src/features/docs/doc-search/components/DocSearchItem.tsx
```diff
@@ -9,39 +12,61 @@ import { useDocSearchFilterStore } from '../stores/useDocSearchFilterStore';
...
+      e.preventDefault();
+    }
+  };
+
   return (
-    <Box
-      data-testid={`doc-search-item-${doc.id}`}
-      $width="100%"
-      className="--docs--doc-search-item"
+    <Link
```
Ovgodd: Hello ! Thank you for the contribution ! 

This link is inside a cmdk option, and the modal already tells people to move with the arrows and select with Enter. Without tabIndex={-1}, Tab stops on every result, so you have to walk the whole list before reaching the filter or the close button. You can add tabIndex={-1}. Ctrl+click, middle-click and "Open in new tab" still work. I checked those.

src/frontend/apps/impress/src/features/docs/doc-search/components/DocSearchItem.tsx
```diff
@@ -9,39 +12,61 @@ import { useDocSearchFilterStore } from '../stores/useDocSearchFilterStore';
...
+      href={`/docs/${doc.id}`}
+      onClick={handleClick}
+      onAuxClick={handleClick}
+      style={{ textDecoration: 'none', color: 'inherit', width: '100%' }}
```
Ovgodd: Nit: we already have StyledLink in @/components, which removes the text decoration and accepts $css. Using it here would avoid the inline style.

PR #2706, opened 2026-09-18
🐛(frontend) fix redirect after deleting a document

### LINKED ISSUE #2465: Redirect to home when deleting from "My Docs"

AntoLC:
## Purpose

We fixed the redirect behavior after deleting a document to ensure the user stays on the correct page after deleting a document.
We added a test to avoid regression in the redirect behavior.


Ovgodd (approved):
💯

PR #2713, opened 2026-09-21
✨(frontend) turn pasted doc links into interlinks

### LINKED ISSUE #2611: Improve document mentioning system:  auto-conversion into mentions for internal links, open to docs outside of the tree
## Enhancement Request

Currently, users manually create HTML links to reference documents outside of the tree structure. This process is:
- Time-consuming and prone to errors.
- Inconsistent in appearance, leading to a poor user experience.

Additionally, internal links pasted into documents remain as raw URLs instead of converting into visually appealing and functional document mentions.

One benefit of this approach is that with good interlinking users will less likely build massive document trees and tend to link trees between them. (User feedback from Paul and Jay).

1. **Enhanced Document Mentioning**:
   - `@[query]` should return document in the tree as first results, then return document outside of the tree the user has access to

2. **Auto-Conversion of Internal Links**:
   - Automatically detect and convert pasted internal links (e.g., raw URLs) into document mentions.

**Drawbacks Considered**:
- **Query results dilution**: returning document outside the tree can dilute pertinent results.
- **Results look alike**: Many subdocs have the same name (QA, Meeting, Retex), for document outside of the tree we should probably add the parent document bellow (like we do in the search modal for subdocs)

Image

- **Performance Impact**: Real-time access checks and auto-conversion might introduce latency for large documents or systems with many users.

AntoLC:
## Purpose

Pasting a bare link to another doc on the same domain now inserts an interlink instead of a plain hyperlink, so it benefits from the same title-sync and navigation behavior as links added through the `/` or `@` menu. Pasting over a selection, inside a code block, or a link the current user can't resolve still falls back to a regular link.

Closes #2611

## Demo

https://github.com/user-attachments/assets/09548daf-bef6-4006-a562-f01f35daeb6e


src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-inline-content/Interlinking/pasteInterlink.ts
```diff
@@ -0,0 +1,67 @@
...
+  if (url.origin !== window.location.origin) {
+    return null;
+  }
+
+  const match = DOC_PATH_REGEX.exec(url.pathname);
+  const docId = match?.[1];
+
+  if (!docId || !uuidValidate(docId)) {
+    return null;
+  }
```
Ovgodd, line 62: The commit message says that pasting a link to a doc the user cannot open should stay a normal link. Pasting http://localhost:3000/docs/a1b2c3d4-e5f6-4789-a123-1234567890ab/ still becomes an interlink chip showing `/docs/<uuid>/` (see screenshot).  

Image

src/frontend/apps/impress/src/features/docs/doc-editor/components/custom-inline-content/Interlinking/LinkSelected.tsx
```diff
@@ -1,47 +1,35 @@
...
+  const { data: doc, isLoading } = useDoc(
+    { id: docId },
+    {
+      queryKey: [KEY_DOC, { id: docId }],
+      refetchOnWindowFocus: (query) => query.state.error?.status !== 403,
+    },
+  );
+  const href = `/docs/${docId}/${blockId ? `#${blockId}` : ''}`;
+  const isHrefFallback = !isLoading && !doc?.title;
+  const { emoji, titleWithoutEmoji } = getEmojiAndTitle(doc?.title || href);
```
Ovgodd, line 29:
```suggestion
const { untitledDocument } = useTrans();
const href = `/docs/${docId}/${blockId ? `#${blockId}` : ''}`;
const label = doc ? doc.title || untitledDocument : href;
const isHrefFallback = !isLoading && !doc;
const { emoji, titleWithoutEmoji } = getEmojiAndTitle(label);
```

Pasting a link to a doc with no title shows `/docs/<uuid>/` on the chip. As the doc tree shows "Untitled document" for the same doc (see screenshot). should the chip use that label too ? 

I took a shot : 

BEFORE : 

Image

AFTER : 

Image

src/frontend/apps/impress/src/features/docs/doc-export/inline-content-mapping/interlinkingLinkPDF.tsx
Ovgodd, line 23: Noticed this while testing. A link copied from a block opens the right place in the editor, but from an exported PDF (same for docx and odt) it opens the doc at the top. I tried keeping the block id in the exported link:

```suggestion
src={
  window.location.origin +
  `/docs/${inline.props.docId}/${inline.props.blockId ? `#${inline.props.blockId}` : ''}`
}
```

Same change on the `link` in `interlinkingLinkDocx.tsx` and the `url` in `interlinkingLinkODT.tsx`.

Ovgodd (commented):
Good feature I like it ! just some suggetions ;

